Issue #5, opened 2021-01-06
A DgsEntityFetcher with an argument of DgsDataFetchingEnvironment fails


paulbakker:
Upon invocation it fails with an invalid number of arguments exception. The code invoking an entityFetcher checks for an argument of type DataFetchingEnvironment, but not DgsDataFetchingEnvironment.

PR #6, opened 2021-01-06
Allow DgsDataFetchingEnvironment as an argument to a DgsEntityFetcher

paulbakker:
Fixes: #5 
Also includes improved error handling in finding entity fetchers.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Netflix, Inc.
+ * Copyright 2021 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
```diff
@@ -49,29 +46,39 @@ open class DefaultDgsFederationResolver() : DgsFederationResolver {
 
     private fun dgsEntityFetchers(env: DataFetchingEnvironment): CompletableFuture<List<Any?>> {
         val resultList = env.getArgument<List<Map<String, Any>>>(_Entity.argumentName)
-                .stream()
-                .map { values ->
-                    val fetcher = dgsSchemaProvider.entityFetchers[values["__typename"]]
+            .stream()
+            .map { values ->
+                val typename = values["__typename"] ?: throw RuntimeException("__typename missing from arguments in federated query")
+                val fetcher = dgsSchemaProvider.entityFetchers[typename]
```
srinivasankavitha, line 52: This is a good fix. We've had some support questions around this because entity fetcher was missing.

srinivasankavitha (approved):

PR #15, opened 2021-02-03
Provide the means to customize when the DefaultDgsQueryExecutor reloads the schema.

berngp:
This commit provides the means to customize when the `DefaultDgsQueryExecutor` will reload its schema.
This is done via the `DefaultDgsQueryExecutor.ReloadSchemaIndicator` functional interface.
We are also providing an implementation of such interface that respects the current reload behavior.


berngp:
Some additional observations... I think that moving the _"initial schema"_  to the `DgsSchemaProvider` in combination with the logic of when it should reload/cached would be cleaner. Shied away from a larger change but I think we should discuss it before we open a new interface, talking about `DefaultDgsQueryExecutor.ReloadSchemaIndicator`.

PR #52, opened 2021-02-09
Provide Micrometer Metrics that capture aspects of GraphQL query latency.

berngp:
## Provide metrics that capture the `graphql` endpoint latency.
This PR introduces the `graphql-dgs-spring-boot-micrometer` module. It has an _auto-configuration_ that enables the capture of GraphQL query metrics, via [micrometer]. The `GraphQLMetricsInstrumentation` will provide the following metrics:
| Name            | Kind  | Description |
| ------------  | -----|  ----------- |
| gql.query                    | Timer       | Captures the time it takes to execute the query.  |
| gql.error                     |  Counter  | Captures number of GraphQL errors, note that one GraphQL request can have multiple.|
| gql.resolver      |   Timer      | Tagged with the resolver's GraphQL field via `gql.field`, it captures the resolver's execution time.|
| gql.dataLoader |  Timer |  Captures the elapse time for a data loader invocation for batch of queries. This is useful if you want to find data loaders that might be responsible for poor query performance. |

All metrics are tagged with the tags provided by the `DefaultGraphQLTagsProvider`. Developers can override the `DefaultGraphQLTagsProvider` by defining their own implementation of the `GraphQLTagsProvider` interface.

The `DefaultGraphQLTagsProvider` defines the following tags.
| Tag                    | Description |
| ------------     | ----------- |
| outcome           | `"SUCCESS"` if the [ExecutionResult] has data, else an `"ERROR"`. |
| gql.path            | (optional) [ValidationError]'s `queryPath` or the [GraphQLError] `path`.[^1]|
| gql.errorCode  | (optional) The [ErrorType]'s name if the error is a [ValidationError], else it will attempt to extract its value from the Error's extensions via the `errorType` key.[^1]|
| gql.errorDetail  | (optional)  it will attempt to extract its value from the GgraphQLError's extensions via the `errorDetail` key. For details on the cardinality of the value please see refer to the [typederror interface](https://netflix.github.io/dgs/error-handling/#the-typederror-interface). |
| gql.loaderName | (specific to gql.Loader) Name of the data loader. |
| gql.loaderBatchSize | (specific to gql.Loader) The number of queries executed in the batch.  |

[^1]: Remember that one request can have multiple errors.

## Examples

* Included as part of this PR we have the [MicrometerServletSmokeTest.kt] test that depicts the metrics emitted on particular scenarios.

----

[micrometer]: https://micrometer.io/
[ErrorType]: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/ErrorType.java
[ExecutionResult]: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/ExecutionResult.java
[GraphQLError]:https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/GraphQLError.java
[ValidationError]: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/validation/ValidationError.java
[MicrometerServletSmokeTest.kt]: https://github.com/Netflix/dgs-framework/blob/2894c3fc96388f0e21e4954740a558dda6626f3a/graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt


jkschneider:
+1 for this support. Look forward to it :)

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/GraphQLMetricsInstrumentation.kt
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                recordDataFetcherTime(sampler, tags, registry)
+            }
+            result
+        }
+    }
+
+    private fun recordDataFetcherTime(timerSampler: Timer.Sample, tags: Tags, registry: MeterRegistry) {
+        timerSampler.stop(registry,
+                Timer.builder("gql.resolver.time").tags(tags).publishPercentileHistogram())
```
jkschneider, line 93: Suggest striking `publishPercentileHistogram` as it it can be turned on later by an application developer if they want histograms for this timer in application.properties with `management.metrics.distribution.percentiles-histogram.gql=true`

berngp: had no idea, thanks for the pointer.

srinivasankavitha (approved):

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/GraphQLMetricsInstrumentation.kt
```diff
@@ -0,0 +1,108 @@
...
+                state.stopTimer(autoTimer.builder("gql.query")
+                        .tags(tagsProvider.getExecutionTags(parameters, result, exc))
+                        .tags(tagsProvider.getEnvironmentTags()))
+            }
+        }
+    }
+
+    override fun instrumentExecutionResult(executionResult: ExecutionResult, parameters: InstrumentationExecutionParameters): CompletableFuture<ExecutionResult> {
+        sanitizeErrorPathsForTags(executionResult).forEach { _ ->
+            registry.counter("gql.error",
```
jkschneider, line 61: Do we need this counter? Timers always ship a count as well as other distribution statistics like max, sum, percentiles, etc. So if I understand this correctly, this data already is retrievable by filtering `gql.query{outcome='ERROR'}` and looking at the count statistic.

berngp: Could you please review the latest code. In the latest implementation a qgl.error is emitted per error in the _Grql Request_. Per the [spec](http://spec.graphql.org/June2018/#sec-Errors) a request can have multiple errors.

berngp:
There is plenty of documentation to write for this feature. In the meantime the [MicrometerServletSmokeTest.kt] offers a hint on the metrics, and the tags, according to different scenarios.

[MicrometerServletSmokeTest.kt]: https://github.com/Netflix/dgs-framework/blob/2894c3fc96388f0e21e4954740a558dda6626f3a/graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt

berngp:
Mining for decent. Most likely will merge tomorrow and have a soft release of this feature since I'm planning on adapting our internal implementation to it.

Issue #49, opened 2021-02-09
Custom scalars not coerced if used as mutation input


Fruko:
Hey guys,

First of all, I really like your framework! One thing that bugs me though, it seems that custom scalars are not coerced in the mutation. As per documentation the `@InputArgument` can be used in mutations see [docs](https://netflix.github.io/dgs/mutations/)

Give the schema with mutation such as
```
type Mutation {
    scalarAsInput(scalar: [UUID]): [UUID]
}

scalar UUID
```

and registering the UUIDScalar using the `@DgsScalar(name="UUID")` and finally writing a DataFetcher 

```
@DgsComponent
public class ScalarDataFetcher {
    @DgsData(parentType = DgsConstants.MUTATION.TYPE_NAME, field = DgsConstants.MUTATION.ScalarAsInput)
    public List<UUID> customScalarMutation(@InputArgument("scalar") List<UUID> scalar) {
        return scalar;
    }
}
```

I am getting the following error
```
    "errors": [
        {
            "message": "Can't serialize value (/scalarAsInput[0]) : Not a valid UUID",
            "path": [
                "scalarAsInput",
                0
            ],
            "extensions": {
                "classification": "DataFetchingException"
            }
        }
    ],
```

When debugging I can see that the `DataFetchingEnvironment` has the argument properly coerced from `String` to `UUID`

![image](https://user-images.githubusercontent.com/10186522/107389085-bff5fb00-6af6-11eb-846f-9df7ef64b335.png)

I forked the examples repository to reproduce the issue [here](https://github.com/Fruko/dgs-examples-java)

srinivasankavitha:
Thanks for the detailed description! We will look into fixing it.

srinivasankavitha:
After looking into this some more, it appears that the handling of the custom scalar,i .e., just returning UUID actually works. However, List<UUID> does not because graphql-java unboxes the List type and treats the individual element as a String. In the implementation of the UUID scalar, the serialize method currently only checks for instanceof UUID (while it is a string in this case)
```
@Override
    public String serialize(Object dataFetcherResult) throws CoercingSerializeException {
        if (dataFetcherResult instanceof UUID) {
            return dataFetcherResult.toString();
        } else {
            throw new CoercingSerializeException("Not a valid UUID");
        }
    }
```
If we modify serialize() to also handle string, it works correctly:
```
@Override
    public String serialize(Object dataFetcherResult) throws CoercingSerializeException {
[...]

PR #53, opened 2021-02-10
Fix handling of list of scalars and non-object types for @InputArgument.

srinivasankavitha:
https://github.com/Netflix/dgs-framework/issues/49

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -237,15 +237,21 @@ class DgsSchemaProvider(private val applicationContext: ApplicationContext,
 
                         val convertValue: Any? = if (parameterValue is List<*> && collectionType != Object::class.java) {
                             try {
+                                // Return a list of elements that are converted to their collection type, e.e.g. List<Person>, List<String> etc.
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -241,7 +241,7 @@ class DgsSchemaProvider(private val applicationContext: ApplicationContext,
                             } catch (ex: Exception) {
                                 throw DgsInvalidInputArgumentException("Specified type '${collectionType}' is invalid for $parameterName.", ex)
                             }
-                        } else if (parameterValue is MultipartFile) {
+                        } else if (parameterValue is List<*> || parameterValue is MultipartFile) {
```
paulbakker, line 244: Shouldn't this be part of the next if statement, and check the type of the the type argument?
The Multipartfile check is a very specific case to check for, so I think combining that with other use cases will be hard to understand later.

paulbakker (changes requested):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
srinivasankavitha: Added a separate case for it because it does handle other List types as well..list of custom scalars, list of Integers etc.

paulbakker (approved):

Issue #70, opened 2021-02-19
Custom scalars not usable as mutation input


faust2199:
Hey guys,

First of all, I really like your framework! I was trying to add a custom scalar (LocalDateTime) to my project, using the latest dgs release 3.5.1. I followed instructions on https://netflix.github.io/dgs/scalars/ and made changes accordingly. However when I tested it it threw exceptions. Then I tried the other method mentioned in https://github.com/Netflix/dgs-examples-java/blob/main/src/main/java/com/example/demo/scalars/DateTimeScalar.java and got the same error.

I've been able to reproduce the issue using my fork of the dgs-examples repo: https://github.com/faust2199/dgs-examples-java/tree/dgs-issue . After adding submittedDate to the input type, the following mutation causes an internal error:

```
mutation {
  addReview (review: {
    showId: 1
    username: "dgs-user"
    starScore: 0
    submittedDate: "1999-02-01T05:43:53.28Z"
  }) {
    username
    submittedDate
  }
}
```

```
{
  "errors": [
    {
      "message": "java.lang.IllegalArgumentException: Cannot construct instance of `java.time.OffsetDateTime` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)\n at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.example.demo.generated.types.SubmittedReview[\"submittedDate\"])",
      "locations": [],
      "path": [
        "addReview"
      ],
      "extensions": {
        "errorType": "INTERNAL"
      }
    }
  ],
  "data": {
    "addReview": null
  }
}
```
By the way, I ran across the following piece of code when looking at the stack trace.
https://github.com/Netflix/dgs-framework/blob/3d38065e9525f349563881ff069b6dc09a865373/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L268

Is it creating a new ObjectMapper on every request? I think it is more efficient to reuse.

Thanks in advance!

srinivasankavitha:
Looking into it. I did your PR, but not sure those changes are needed.

faust2199:
> Looking into it. I did your PR, but not sure those changes are needed.

[...]

PR #72, opened 2021-02-19
Reuse ObjectMapper and enable JavaTimeModule

faust2199:
Fixes my issue #70


srinivasankavitha:
Reviewing this.

srinivasankavitha (commented):
Changes look good to me. Thanks for the PR. Would you be able to add a test case to DgsSchemaProvider to cover this use case? Specifically, DateTime scalar inside an input type.

srinivasankavitha (approved):

paulbakker (approved):
Merging as a temporary workaround. Details why this isn't the proper fix and what should happen instead described here: https://github.com/Netflix/dgs-framework/issues/70#issuecomment-785461229

Issue #73, opened 2021-02-20
Runtime class not found exception with 3.5.1 version


maxixcom:
I've created the new application from start.spring.io with the following dependencies plus DGS `3.5.1`:

```
dependencies {
    implementation("org.springframework.boot:spring-boot-starter-data-mongodb")
    implementation("org.springframework.boot:spring-boot-starter-security")
    implementation("org.springframework.boot:spring-boot-starter-web")
    implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
    implementation("org.jetbrains.kotlin:kotlin-reflect")
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
    implementation("org.springframework.cloud:spring-cloud-starter-config")
    implementation("org.springframework.cloud:spring-cloud-starter-netflix-eureka-client")
    annotationProcessor("org.springframework.boot:spring-boot-configuration-processor")
    testImplementation("org.springframework.boot:spring-boot-starter-test")
    testImplementation("org.springframework.security:spring-security-test")

    // DGS
    implementation("com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter:3.5.1")
}
```

<details><summary>Full build.gradle.kts</summary>
<p>

```
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

plugins {
    id("org.springframework.boot") version "2.4.3"
    id("io.spring.dependency-management") version "1.0.11.RELEASE"
    kotlin("jvm") version "1.4.30"
    kotlin("plugin.spring") version "1.4.30"
}

group = "studio.buket.apps.bazar"
version = "0.0.1-SNAPSHOT"
java.sourceCompatibility = JavaVersion.VERSION_11

configurations {
    compileOnly {
        extendsFrom(configurations.annotationProcessor.get())
    }
}

repositories {
    mavenCentral()
    jcenter()
}

extra["springCloudVersion"] = "2020.0.1"

dependencies {
    implementation("org.springframework.boot:spring-boot-starter-data-mongodb")
    implementation("org.springframework.boot:spring-boot-starter-security")
    implementation("org.springframework.boot:spring-boot-starter-web")
    implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
    implementation("org.jetbrains.kotlin:kotlin-reflect")
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
    implementation("org.springframework.cloud:spring-cloud-starter-config")
[...]

PR #87, opened 2021-02-24
Removes the exclusion of json-smart from graphql-dgs-client

bono007:
@paulbakker 

Fixes gh-73

### DIFF
--- a/graphql-dgs-client/build.gradle.kts
+++ b/graphql-dgs-client/build.gradle.kts
@@ -18,9 +18,7 @@ plugins {
 }
 
 dependencies {
-    api("com.jayway.jsonpath:json-path:2.4.+") {
-        exclude(group = "net.minidev", module = "json-smart")
-    }
+    api("com.jayway.jsonpath:json-path:2.4.+")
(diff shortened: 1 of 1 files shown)


paulbakker:
Thanks @bono007 !

bono007:
> Thanks @bono007 !

No worries @paulbakker - had I known about `spring-data-common` use (partial swap of mapping provider but not json provider) and the fact that json-path does not really let you "swap" providers easily (and remove them from the classpath) I would not have submitted the proposal until at least one of the above were resolved. Sorry for introducing the toil. 

Once again, another learning of the "Dependencies are hard" variety. :)

paulbakker:
Yeah, same here, I didn't expect this problem either. 
We still made the code changes so that the framework doesn't depend on json-smart anymore though, so at least the result is that you now _can_ exclude the dependency.

bono007:
> We still made the code changes so that the framework doesn't depend on json-smart anymore though, so at least the result is that you now can exclude the dependency.

Absolutely. It was not time wasted IMO.

PR #89, opened 2021-02-24
Added support for ExecutionIdProvider

marceloverdijk:
This provides support for providing a custom `graphql-java` `ExecutionIdProvider` bean to be used in the `GraphqL` instance.

We generate id's for every HTTP request and re-use this id in the GraphQL execution for tracing purposes.

On a related note, I noticed a new `GraphQL` instance seems to be created for every GraphQL request [here](https://github.com/Netflix/dgs-framework/blob/1f2d40fe6d51fced9ed4e6d8043808b0f8cf16bb/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt#L85-L91).
Probably it is done this way to support [reloading the schema](https://github.com/Netflix/dgs-framework/blob/1f2d40fe6d51fced9ed4e6d8043808b0f8cf16bb/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt#L79-L83) but I think that could be optimized by caching the `GraphQL` instance. Interested in another PR for that?

### DIFF
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -32,6 +32,7 @@ import graphql.execution.AsyncExecutionStrategy
 import graphql.execution.AsyncSerialExecutionStrategy
 import graphql.execution.DataFetcherExceptionHandler
 import graphql.execution.ExecutionStrategy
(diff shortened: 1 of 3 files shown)


paulbakker:
Thanks! 

For recreating the GraphQL instance, the GraphQL-java [docs state](https://javadoc.io/doc/com.graphql-java/graphql-java/latest/index.html) “Building this object is very cheap and can be done on each execution if necessary”. So there isn’t a lot to win there I think.

marceloverdijk:
Good point about the mention of building the `GraphQL` doesn't matter, and thx for merging the PR.

Issue #88, opened 2021-02-24
Introduce config properties


marceloverdijk:
The DGS framework takes a hard stance on how to configure itself:

- Hardcoded `/graphql` path
- Hardcoded `/schema.json` path and whether it is enabled or not
- Hardcoded `/graphiql` path and whether it is enabled or not
- Hardcoded `classpath*:schema/**/*.graphql*` to load the schema files from

It would be nice if DGS would offer some configuration options like:

```
dgs.graphql.path = /graphql
dgs.graphql.schema-location = classpath*:schema/**/*.graphql*
dgs.graphql.schema-json.enabled = true
dgs.graphql.schema-json.path = /schema.json
dgs.graphql.graphiql.enabled = true
dgs.graphql.graphiql.path = /graphiql
```

Rationale is to be able to change paths, but also to disable certain functionality. 
E.g. GraphiQL is very useful during development but it might be desired or required to disable this in production because of security requirements or customized GraphiQL pages served differently (same for `/schema.json`). 
For the schema-location, some might want to use a different location like `classpath:graphql/schema.graphqls` to reduce classpath scanning.

Probably there are more hardcoded things that could be moved to [@ConfigrationProperties](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/context/properties/ConfigurationProperties.html) (websockets?)

paulbakker:
Using `@ConfigurationProperties` in general is something we want to do. 
Adding extra properties for paths etc. makes sense too, there isn’t really a reason these things aren’t configurable other than we didn’t have a need for it (but that doesn’t mean others won’t).

bono007:
### Preface Thoughts
- The "introduce config props" can be a nebulous bag of features as each configuration "knob" has to have something done w/ it to make it work. I do agree with and understand the motivation of the ticket though. 
- We all know this, but it is re-stating out loud: Premature configuration comes w/ a cost as the configuration variants have to be maintained and bug fixed. The bugs will come in. :). I do believe that the "configuration knobs" in this ticket make sense and somebody **may** want to use them, but unless someone is actually asking for them, we should probably punt on the non-trivial ones (graphiql path is in this camp). 

### Summary
So interestingly enough, only 1 of the example configuration "knobs" in the description would introduce a `@ConfigurationProperties` entry. Here is why:

[...]

PR #94, opened 2021-02-28
Add ability to disable schema-json and graphiql endpoints

bono007:
@paulbakker @berngp @marceloverdijk

This is "Part I" as described in the issue. 

This proposal is still missing (I will do this later today):
*  properties in additional meta info file
* docs

Fixes gh-88

### DIFF
--- a/graphql-dgs-graphiql-autoconfigure/build.gradle.kts
+++ b/graphql-dgs-graphiql-autoconfigure/build.gradle.kts
@@ -17,4 +17,6 @@
 dependencies {
     implementation("org.springframework.boot:spring-boot-autoconfigure")
     implementation("org.springframework:spring-webmvc")
+
+    testImplementation("org.springframework.boot:spring-boot-starter-web")
 }
--- a/graphql-dgs-graphiql-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/graphql-dgs-graphiql-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1,0 +1,10 @@
+{
+  "properties": [
+    {
+      "name": "dgs.graphql.graphiql.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enables GraphiQL functionality.",
+      "defaultValue": "true"
+    }
+  ]
+}
--- a/graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLAutoConfigurationTest.kt
+++ b/graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLAutoConfigurationTest.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.graphiql.autoconfiguration
+
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import org.springframework.boot.autoconfigure.AutoConfigurations
+import org.springframework.boot.test.context.runner.WebApplicationContextRunner
+
+class GraphiQLAutoConfigurationTest {
+
+    private val context = WebApplicationContextRunner().withConfiguration(AutoConfigurations.of(GraphiQLAutoConfiguration::class.java))!!
+
+    @Test
+    fun graphiqlAvailableWhenEnabledPropertyNotSpecified() {
+        context.run { ctx ->
+            Assertions.assertThat(ctx).hasSingleBean(GraphiQLConfigurer::class.java)
+        }
+    }
+
+    @Test
+    fun graphiqlAvailableWhenEnabledPropertySetToTrue() {
+        context.withPropertyValues("dgs.graphql.graphiql.enabled: true").run { ctx ->
+            Assertions.assertThat(ctx).hasSingleBean(GraphiQLConfigurer::class.java)
+        }
+    }
+
+    @Test
+    fun graphiqlNotAvailableWhenEnabledPropertySetToFalse() {
+        context.withPropertyValues("dgs.graphql.graphiql.enabled: false").run { ctx ->
+            Assertions.assertThat(ctx).doesNotHaveBean(GraphiQLConfigurer::class.java)
+        }
+    }
+}
--- a/graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1,0 +1,16 @@
+{
+  "properties": [
+    {
+      "name": "dgs.graphql.schema-json.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enables schema-json endpoint functionality.",
+      "defaultValue": "true"
+    },
+    {
+      "name": "dgs.graphql.schema-json.path",
+      "type": "java.lang.String",
+      "description": "Path to the schema-json endpoint.",
+      "defaultValue": "/schema.json"
+    }
+  ]
+}
--- a/graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfiguration/DgsWebMvcAutoConfigurationTest.kt
+++ b/graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfiguration/DgsWebMvcAutoConfigurationTest.kt
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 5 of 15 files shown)


graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLAutoConfigurationTest.kt
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package com.netflix.graphql.dgs.graphiql.autoconfiguration
+
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import org.springframework.boot.autoconfigure.AutoConfigurations
+import org.springframework.boot.test.context.runner.WebApplicationContextRunner
+
+class GraphiQLAutoConfigurationTest {
```
bono007, line 24: I know there was no tests in the auto-config modules for graphiql or mvc but I think its worth putting the tests in here rather than rely on them being in the "outer" autoconfig module as its easier for future contributors to see where the auto-config tests are and as this grows bigger so will these tests.

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -64,10 +61,10 @@ import org.springframework.web.multipart.MultipartFile
  */
 
 @RestController
-class DgsRestController(private val schemaProvider: DgsSchemaProvider, private val dgsQueryExecutor: DgsQueryExecutor) {
+class DgsRestController(private val dgsQueryExecutor: DgsQueryExecutor) {
     val logger: Logger = LoggerFactory.getLogger(DgsRestController::class.java)
 
-    @RequestMapping("/graphql", produces = ["application/json"])
+    @RequestMapping("\${dgs.graphql.path:/graphql}", produces = ["application/json"])
```
berngp, line 67: I like detaching the definition of default values from the place where they are used. This can be done with an EnvironmentPostProcessor that loads a _default properties file_. 

That said, I think this should be considered as a separate pr.

berngp (approved):
yhank you form the contribution and specially for the arguments you made on the issue that originated this change.

The only thing I would say is missing is the configuration metadata.

Red https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
bono007: Thats a good idea especially when the number of properties grows and you can see all of the defaults in a single place. 

The one area where this is, not in conflict, but would need to be agreed upon is the use of `@DefaultValue` in the `@ConfigurationProperties`. It would be confusing having defaults in the config props and the defaults file. We could just make a guideline to not use `@DefaultValue` in config props though.

bono007:
> yhank you form the contribution and specially for the arguments you made on the issue that originated this change.
> 
> The only thing I would say is missing is the configuration metadata.
> 
> Red https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html

You are more than welcome @berngp . I plan on doing docs and config props in this MR but later today as a follow on commit. I'll ping you when its in.

bono007:
@berngp @paulbakker @marceloverdijk 

I pushed a 2nd commit that adds the config metadata files along w/ the `kapt` plugin to process them. This sets us up nice when we start adding `@ConfigurationProperties`, they should just work. 

I verified in one of my teams app that uses the starters and I get this beauty in the IDE....

Screen Shot 2021-02-28 at 9 09 09 PM

I have added some inline comments to the review.

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -16,6 +16,7 @@
 
 object Versions {
     const val KOTLIN_VERSON = "1.4.10"
+	const val KAPT_VERSION = "1.4.10"
```
bono007, line 19: There is a more recent version but when I moved it and Kotlin version out of sync - it did not end well.

graphql-dgs-graphiql-autoconfigure/build.gradle.kts
```diff
@@ -16,9 +16,23 @@
...
     implementation("org.springframework:spring-webmvc")
-}
\ No newline at end of file
+
+	kapt("org.springframework.boot:spring-boot-configuration-processor:${Versions.SPRING_BOOT_VERSION}")
+
+	testImplementation("org.springframework.boot:spring-boot-starter-web")
+}
+
+kapt {
```
bono007, line 31: This should not be necessary and is probably a bug in kapt or the boot config processor - trust me, it does not include the metadata w/o this - it robbed me of 1-2 hrs of time. It works w/ this though. We will probably want to factor this out for easier re-use when we start adding config props.

graphql-dgs-graphiql-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+  "properties": [
+    {
+      "name": "dgs.graphql.graphiql.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enables GraphiQL functionality.",
```
bono007, line 6: **[HELP_SOLICITED]** not the best description.

graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "properties": [
+    {
+      "name": "dgs.graphql.schema-json.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enables schema-json endpoint functionality.",
```
bono007, line 6: **[HELP_SOLICITED]** not the best description.

bono007:
I think we should add a CONTRIBUTING.md or some developer facing light doc that tells how to run, test, and now about additional config props, regular config props, and kapt. Not a manual, just some important lightweight instruction. 

Thoughts?

berngp:
Will review further tomorrow, some remarks...

I would like to avoid kapt and instead use Spring Boot's annotation post-processor for gradle, ref [here](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html#configuration-metadata-annotation-processor-setup).


Agree on all the points you mentioned about the guidelines. On fact I'm interested at defining auto-formatting via spotless (foe example)

And again, thanks for all this work!

bono007:
> I would like to avoid kapt and instead use Spring Boot's annotation post-processor for gradle, ref here.

@berngp I believe you have to use `kapt` in order to process the `@ConfigurationProperties` - thats what I picked up from [this](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-kotlin-configuration-properties).

graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "properties": [
```
paulbakker, line 2: Shouldn't `dgs.graphql.path` be included as well?

paulbakker (commented):
> > I would like to avoid kapt and instead use Spring Boot's annotation post-processor for gradle, ref here.
> 
> @berngp I believe you have to use `kapt` in order to process the `@ConfigurationProperties` - thats what I picked up from [this](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-kotlin-configuration-properties).

The docs are a little confusing; is kapt only required to work with immutable config classes, or is it required to get it to work at all? I _think_ it's the latter, but I'm not sure.

paulbakker:
> > The docs are a little confusing; is kapt only required to work with immutable config classes, or is it required to get it to work at all? I _think_ it's the latter, but I'm not sure.
> 
> If that is the case, it will come down to a "What is worse, kapt or mutable config props?" decision. I don't have any ill feeling towards kapt yet and I have really grown to ❤️ my immutable config props. hehe. I am guessing that there are negatives and hence a desire to avoid kapt.

I have no experience with it, but generally prefer to not rely on too many plugins. However, this seems to be a very common one, so I have no issue with it.

paulbakker:
@bono007 Everything looks good to me - happy to merge whenever you're ready.

berngp:
@bono007 if you rebase against master you don't need to set `kapt` nor configure it to include any kind of resource.

bono007:
Awesome! I saw the email notifications but have not got a chance to rebase.
Will do that this evening as well as split out the path configuring portion
into follow on mr.

graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
bono007: I split out the path config based on our earlier discussion.

graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfiguration/DgsWebMvcAutoConfigurationTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    @Test
+    fun schemaJsonControllerNotAvailableWhenEnabledPropertySetToFalse() {
+        context.withPropertyValues("dgs.graphql.schema-json.enabled: false").run { ctx ->
+            Assertions.assertThat(ctx).doesNotHaveBean(DgsRestSchemaJsonController::class.java)
+        }
+    }
+
+    @Test
+    fun schemaJsonControllerMappedToDefaultPath() {
+        context.withConfiguration(AutoConfigurations.of(JacksonAutoConfiguration::class.java, WebMvcAutoConfiguration::class.java, DispatcherServletAutoConfiguration::class.java))
```
bono007, line 79: This and the following test are new since last commit. I did not have any coverage around switching the schema-json path and it actually service at that path. This adds a bit of Fu to the context runner to allow mockmvc in the runner. Its a trick I picked up from [Spring Boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/integrationtest/WebMvcEndpointCorsIntegrationTests.java#L51).

bono007:
Ok @paulbakker @berngp .. I rebased and ended up squashing as it was getting confusing to look at from a commit history perspective. 

Here is the summary of the current bits:
- add ability to disable graphiql and schema-json endpoints
- add ability to configure schema-json path 
- all properties added in metadata files
- all functionality unit/integration tests 

Here is what I removed:
- ability to configure graphql path 

Here is what I will do next (after this is merged):
- start PR that configures graphql path (the portion I removed from above)
- this PR will eventually hold the changes to configure graphiql path as well

berngp:
Thanks for all the work with this PR @bono007 !

PR #99, opened 2021-03-01
Enable Spring Boot's Configuration Processors.

berngp:
This commit enables Spring Boot's Configuration Processors via the [kapt] Gradle plugin.
This enables us to delivery, as part of our artifacts, both the `spring-autoconfigure-metadata.properties`
and the `additional-spring-configuration-metadata.json` files.

As define in the [Autoconfigure Module],
> Spring Boot uses an annotation processor to collect the conditions on auto-configurations in
> a metadata file (META-INF/spring-autoconfigure-metadata.properties). If that file is present,
> it is used to eagerly filter auto-configurations that do not match, which will improve startup time.

[kapt]: https://kotlinlang.org/docs/kapt.html
[Autoconfigure Module]: https://docs.spring.io/spring-boot/docs/2.3.x/reference/htmlsingle/#boot-features-custom-starter-module-autoconfigure

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,11 +26,11 @@ plugins {
     `java-library`
     id("nebula.netflixoss") version "8.10.0"
     id("nebula.dependency-recommender") version "9.1.1"
-    kotlin("jvm") version Versions.KOTLIN_VERSON apply false
+    kotlin("jvm") version Versions.KOTLIN_VERSION
+    kotlin("kapt") version Versions.KOTLIN_VERSION
     idea
     eclipse
 }
-
 
 allprojects {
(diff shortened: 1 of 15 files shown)


berngp:
CC. @bono007 this PR enables kapt post-processing for all submodules. We can apply this change before yours to facilitate your PR.

paulbakker (approved):

bono007:
Looks good @berngp - this will make it so that nobody has to think about `kapt` going forward everytime config props are added etc.. Thank you. Also cleans up the `build.gradle.kts` by pulling the `jvm` plugin up to root !

PR #107, opened 2021-03-03
Applied standard kotlin code style

paulbakker:
This PR from hell reformats each and every file in the project.

We've been struggling with keeping code formatting consistent in pull requests, and the "Kotlin Code Style" is now available. The idea is to just get it over with and apply this style, so it won't be in our way as much in the future.

This will obviously not be ideal for anyone with an open PR. I think the easiest way to work around it is rebase on this commit, force your changes to overwrite the ones form this commit, and reformat again.

@bono007 @berngp @marceloverdijk would you be ok with this?

Btw, don't bother actually reviewing this, the reformatting was completely automatic. This PR is only there to get eyes on the upcoming annoying change.

### DIFF
--- a/.github/workflows/releasedraft.yml
+++ b/.github/workflows/releasedraft.yml
@@ -8,7 +8,7 @@ on:
   # pull_request event is required only for autolabeler
   pull_request:
     # Only following types are handled by the action, but one can default to all as well
-    types: [opened, reopened, synchronize]
+    types: [ opened, reopened, synchronize ]
 
 jobs:
   update_release_draft:
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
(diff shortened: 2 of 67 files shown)


berngp (approved):
I
  do
     appreciate
  consistent
formatting.

Thanks!

bono007:
I tOo tHinK its a gOOD IDEa to cONsistently formAT.    . Lol

Thx for asking about the open PRs, much appreciated. I don't mind applying
the changes i have over the merged files.  There will never be a good time
and always some PRs in flight.

Thx for adding this.

paulbakker:
Ok, thanks for the feedback, and I went ahead and merged it. Probably the sooner we get it over with the less disruptive it is. 
New contributors that I didn't know about yet.... Please forgive me :)

bono007:
It's the "share the pain plan" :)

bono007:
Are there any specific commands we need to know about in order to use the formatter? Hook into IDEA? etc..

paulbakker:
I added the relevant IntelliJ config file, as suggested by their docs. So it _should_ just work.

paulbakker:
@bono007 thumbs down because it doesn't work, or...?

This is what you should see:
image

Additionally, I enable reformat on commit, but I don't think that setting should be shared.
image

bono007:
DOH! Stupid keyboard. Sorry about that @paulbakker - That was supposed to be a 👍🏻 UP not DOWN. I have not tried it yet. I will report here how it works here in a bit.

bono007:
I am seeing what you are seeing now! Also, thx for sharing those "Before Commit" - I added those as well. All in.

Issue #95, opened 2021-02-28
Provide access to WebRequest in DgsCustomContextBuilder


ntungare:
Is it possible to provide access to the original request in the [DgsCustomContextBuilder](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsCustomContextBuilder.kt)?

Background:
In the company where I work, we use Apollo Federation Gateway to generate certain headers related to the client and also forward certain user-level cookies over to the Federated Services.

Generally we are able to create a custom [GraphQLInvocation](https://github.com/graphql-java/graphql-java-spring/blob/master/graphql-java-spring-webmvc/src/main/java/graphql/spring/web/servlet/GraphQLInvocation.java) and get access to the `WebRequest` via which we can read all these details and then do things with the `WebRequest` such as authenticating the cookies and perform some side effects that are common for all DataFetchers.

I'm not entirely sure if this currently possible and whether such a functionality is possible.

ntungare:
I think I have this figured out by using Spring to extract current Request:
```java
public static Optional<HttpServletRequest> getCurrentHttpRequest() {
    return Optional.ofNullable(RequestContextHolder.getRequestAttributes())
            .filter(requestAttributes -> ServletRequestAttributes.class.isAssignableFrom(requestAttributes.getClass()))
            .map(requestAttributes -> ((ServletRequestAttributes) requestAttributes))
            .map(ServletRequestAttributes::getRequest);
}
```

But we do lose the visibility into the extensions that the [default context](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt#L74) has access to.

paulbakker:
Be aware that this can be problematic when working with async datafetchers that run on different threads. Spring uses ThreadLocal to store these things (for Spring WebMVC that is).
I do agree that at the very least we need documentation around this.

ntungare:
Yep, would love some docs for best practices on this from DGS perspective.

And re thread local + datafetchers, I do agree that this could be problematic for async datafetchers but accessing the request only in [DgsCustomContextBuilder](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsCustomContextBuilder.kt) should be relatively safe, but obviously depends on how that builder is written.

[...]

Issue #93, opened 2021-02-28
@InputArgument to default to parameter name


aramir:
It would be great for `@InputArgument`'s `value` to default parameter name -- similar to Spring Framework's `@PathVariable` and `@RequestParam` annotations. So we can compact
```java
public List<Show> shows(@InputArgument("titleFilter") String titleFilter)
```
to
```java
public List<Show> shows(@InputArgument String titleFilter)
```

paulbakker:
@aramir I thought parameter name reflection is only possible if the `-parameters` compiler flag is set. I'm now reading that with Spring Boot it _does_ work out of the box. Is that correct?

In that case I'm definitely in favor of support this.

aramir:
It actually works even with vanilla Spring, without Spring Boot. And yes - out of the box, no flag needed.

aramir:
After looking into it, Spring uses an implementation of [ParameterNameDiscoverer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/ParameterNameDiscoverer.html) interface called [DefaultParameterNameDiscoverer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/DefaultParameterNameDiscoverer.html) which, internally sequentially tries a few different methods to figure out parameter names. 

It *does* depend on either `-g:vars` or `-parameter`, however, both maven and gradle set those for java compile processes by default.

More info:

 - [Spring MVC - Automatic Parameter Names Discovery](https://www.logicbig.com/tutorials/spring-framework/spring-web-mvc/parameter-name-discoverer.html)
 - [How Spring gets parameter names without debug information](https://stackoverflow.com/questions/36466913/how-spring-gets-parameter-names-without-debug-information)

paulbakker:
Fixed in #112

Issue #67, opened 2021-02-18
Provide support for Spring's @RequestHeader annotation.


nephyst:
Currently to get a request header I have to do the following:

```
@DsgData(parentType = "Query", field = "shows")
public List<Show> shows(DataFetchingEnvironment env) {
    DgsContext context = env.getContext();
    DefaultRequestData requestData = (RequestData)context.getCustomContext();
    HttpHeaders headers = requestDatagetHeader();
    var someHeader = headers.get("some-header");
    
    //...
}    
```
Spring provides a @RequestHeader annotation that makes this a lot cleaner.

I would expect it to work similar to @InputArgument:

```
@DsgData(parentType = "Query", field = "shows")
public List<Show> shows(DataFetchingEnvironment env,
                        @RequestHeader(name = "some-header") String someHeader) {
    //...
}    
```

srinivasankavitha:
Thanks, we will look into it.

paulbakker:
Fixed in #112

PR #112, opened 2021-03-04
Revamped RequestData on DgsContext, support for @RequestHeader and @InputArgument without name

paulbakker:
Improves access to request data. This should always be available, also when custom context is provided. Also added `@RequestHeader` and improved support for annotated arguments without names such as `@RequestHeader` or `@InputArgument`.

Note that the change to how RequestData is stored is potentially breaking, although the old API wasn't advertised. To prevent this the old mechanism is still available but will show as deprecated.

This PR fixes: 
* #93
* #67 
* #95

These changes were combined in a single PR because they are related to the same code.

New features
====

-  The signature of `DgsContext` is now `DgsContext(val customContext: Any?, val requestData: DgsRequestData?)`
- Previously the RequestData was set as the custom context object. If a custom context was user-provided (which is common) the RequestData would not be set at all, and there was no good way to add it manually to the custom context. 
- `DgsContext.getRequestData(dataFetchtingEnvironment` as shortcut to get RequestData
- DgsRequestData also has a reference to WebRequest
- `@RequestHeader` as a shortcut to get a specific header value.
- Headers are technically Lists, so the annotation can be either used on a `List` or on a `String`. In the latter case multiple values will be concatinated.
- `@InputArgument` without specifying a name explicitly. If no name is provided the argument name is discovered using https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/DefaultParameterNameDiscoverer.html


srinivasankavitha (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsDataFetchingEnvironment.kt
```diff
@@ -34,6 +35,15 @@ import java.util.*
 
 class DgsDataFetchingEnvironment(private val dfe: DataFetchingEnvironment) : DataFetchingEnvironment {
 
+    fun getDgsContext(): DgsContext {
+        val context = dfe.getContext<Any>()
+        if (context is DgsContext) {
+            return context
+        } else {
+            throw RuntimeException("Context object is not a DgsContext. This method does not work if you have a custom implementation of DgsContextBuilder")
```
berngp, line 43: Show the type of the `context` in the exception.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -241,13 +244,13 @@ class DgsSchemaProvider(private val applicationContext: ApplicationContext,
     private fun invokeDataFetcher(method: Method, dgsComponent: Any, environment: DataFetchingEnvironment): Any? {
         return try {
             val args = mutableListOf<Any?>()
-
-            method.parameters.forEach { parameter ->
+            val parameterNames = DefaultParameterNameDiscoverer().getParameterNames(method) ?: emptyArray()
+            method.parameters.forEachIndexed { idx, parameter ->
```
berngp, line 248: the lambda is probably getting a bit large, should we reflect it as an internal method?

berngp (approved):
Great feature.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
paulbakker: Planning to address that separately

PR #119, opened 2021-03-05
Fix bug causing contextual tags to always be empty

kilink:
The method for collating the contextual tags was erroneously using a
collect on a stream instead of reduce like the pattern followed for the
other methods; the accumulator for a collect call is a BiConsumer vs. a
BiFunction for reduce and thus assumes a mutable state.

In the course of fixing the issue, also switch to more idiomatic
constructions using Kotlin flatMap / sequence / etc.


berngp (approved):
Thanks!!

PR #147, opened 2021-03-13
Support multiple beans of type DataLoaderInstrumentationExtensionProvider

berngp:
We are adding support for multiple `DataLoaderInstrumentationExtensionProvider` such that developers can define them to cover specific concerns. We are also fetching the beans of such type with an `orderedStream` such that we can control the order of application.


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsDataFetchingEnvironmentTest.kt
```diff
@@ -43,7 +43,7 @@ class DgsDataFetchingEnvironmentTest {
     @RelaxedMockK
     lateinit var dfeMock: DataFetchingEnvironment
 
-    val helloFetcher = object : Any() {
+    val helloFetcher = @DgsComponent object : Any() {
```
berngp, line 46: Will remove the `@DgsComponent`

berngp:
I'll be adding additional tests that assert multiple, ordered, implementations of `DataLoaderInstrumentationExtensionProvider`.

PR #150, opened 2021-03-13
Adding Gradle org.jlleitschuh.gradle.ktlint-idea plugin.

berngp:
Still having trouble having IntelliJ format the Kotlin code with the same style as the ktlint plugin.
This commit also adds the  `org.jlleitschuh.gradle.ktlint-idea`, in addition I ran the the `ktlintApplyToIdea` task.


paulbakker:
What are the issues/diferences you're running into?

berngp:
I'm not having a problem perse, just an unmet expectation. My expectation is that if I run the `ktlintFormat` Gradle task and see no files changed; therefore I should no see any formatting change if I open any Kotlin file and apply formatting changes via IntelliJ.  Right now this is not the case, for example the `DgsGraphQLCollatedMetricsTagsProvider.kt` file. If I run the task I see no change but If I apply the format via IntelliJ I do.

This PR doesn't fix things, or at least yet.

Issue #128, opened 2021-03-08
Improve path configuration with @ConfigurationProperties


bono007:
We added the ability to configure the GraphQL and GraphiQL paths in gh-100. 

However, we are using `@Value` for the properties in [GraphiQLConfigurer.kt](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-graphiql-autoconfigure/src/main/java/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLConfigurer.kt#L38). This can be improved with `@ConfigurationProperties`. 

Additionally, we require the paths to be configured w/o a trailing slash but we are not validating that requirement - config props would lend itself well to that task. 

### Acceptance criteria
- create one or more `@ConfigurationProperties` for:
  - dgs.graphql.path (default = `/graphql`)
  - dgs.graphql.graphiql.path (default = `/graphiql`)
- validate the above properties do not have trailing slash
- leverage the new config props in`GraphiQLConfigurer.kt`
- leverage the new config props in `DgsRestController.kt` if possible.
- start some form of config props docs at this point as well (the earlier we do this, the easier it will be going forward)

### Notes
- the GraphQL path is also specified in the `@RequestMapping` of [DgsRestController.kt](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt#L68). It would be nice to share the same config props mechanism for "all the things". 

- will there be a single `DgsProperties` that holds both GraphQL and GraphiQL paths? Both paths are needed by the GraphiQLConfigurer but both config props could be enabled there. GraphiQL will just need to be able to "see" the GraphQL config props. Anyways, some structure/design that needs to be noodled on a bit.

PR #152, opened 2021-03-14
[Fine Grained Approach] Replace path properties with @ConfigurationProperties

bono007:
Fixes gh-128

@paulbakker @berngp 

This is the "Fine Grained" approach as described in [this comment](https://github.com/Netflix/dgs-framework/pull/148#discussion_r593825252). 

I like this much better. One thing you will notice is that I had to make graphiql depend on webmvc module. It was either that or duplicate the same property for `dgs.graphql.path` in the graphiql config props. One of the reasons of moving to config props was to not have to specify the defaults in multiple places. 

While the graphiql module did not previously depend on the webmvc module via dependencies, it does depend on the GraphQL controller being available. For this reason it makes me feel that adding a dep to the wembvc module is not that bad. Really I think that the graphiql module should just be merged w/ the webmvc module. Were there other reasons for having them split up?

### DIFF
--- a/graphql-dgs-graphiql-autoconfigure/src/main/java/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLConfigurationProperties.kt
+++ b/graphql-dgs-graphiql-autoconfigure/src/main/java/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLConfigurationProperties.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.graphiql.autoconfiguration
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.ConstructorBinding
+import org.springframework.boot.context.properties.bind.DefaultValue
+
+/**
+ * Configuration properties for DGS GraphiQL.
+ */
+@ConstructorBinding
+@ConfigurationProperties(prefix = "dgs.graphql.graphiql")
+@Suppress("ConfigurationProperties")
+data class GraphiQLConfigurationProperties(
+    /** Path to the GraphiQL endpoint without trailing slash. */
+    @DefaultValue("/graphiql") val path: String
+)
--- a/graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLPathTestApplication.kt
+++ b/graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLPathTestApplication.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.graphiql.autoconfiguration
+
+import com.netflix.graphql.dgs.DgsQueryExecutor
+import com.netflix.graphql.dgs.internal.DgsSchemaProvider
+import io.mockk.mockk
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.context.annotation.Bean
+
+@SpringBootApplication
+open class GraphiQLPathTestApplication {
+    @Bean
+    open fun dgsSchemaProvider(): DgsSchemaProvider {
+        return mockk()
+    }
+
+    @Bean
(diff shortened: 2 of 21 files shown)


bono007:
Here goes commit #2 that shows the dependency impact of the MR.

berngp:
> While the graphiql module did not previously depend on the webmvc module via dependencies, it does depend on the GraphQL controller being available. For this reason it makes me feel that adding a dep to the wembvc module is not that bad. Really I think that the graphiql module should just be merged w/ the webmvc module. Were there other reasons for having them split up?


I agree that it makes sense to have them as one. Thoughts @paulbakker ?

graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLPathTestApplication.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.netflix.graphql.dgs.graphiql.autoconfiguration
+
+import com.netflix.graphql.dgs.DgsQueryExecutor
+import com.netflix.graphql.dgs.internal.DgsSchemaProvider
+import io.mockk.mockk
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.context.annotation.Bean
+
+@SpringBootApplication
+open class GraphiQLPathTestApplication {
```
berngp, line 26: Just curious where this is being used.

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestSchemaJsonController.kt
```diff
@@ -32,7 +32,8 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 class DgsRestSchemaJsonController(private val schemaProvider: DgsSchemaProvider) {
 
-    @RequestMapping("\${dgs.graphql.schema-json.path:/schema.json}", produces = ["application/json"])
+    // The @ConfigurationProperties bean name is <prefix>-<fqn>
+    @RequestMapping("#{@'dgs.graphql-com.netflix.graphql.dgs.webmvc.autoconfigure.DgsWebMvcConfigurationProperties'.schemaJson.path}", produces = ["application/json"])
```
berngp, line 36: Still impressed about seeing this syntax, TIL.

berngp (approved):
Thanks again @bono007 for this PR.
Thanks as well for taking the feedback from the previous implementation and landing with this great work.

bono007:
> Thanks again @bono007 for this PR.
> Thanks as well for taking the feedback from the previous implementation and landing with this great work.

You are more than welcome @berngp . I can imagine what a solution will look like in my head, but sometimes the caveats and wrinkles don't appear until I actually "see it". I am always in favor of reworking an initial solution to get to a pragmatically "good" solution. It is so much more  inexpensive to iterate now than it is once its released into the wild. :)

graphql-dgs-graphiql-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLPathTestApplication.kt
bono007: Ahh yes, I should have commented about this. Good catch. The **TL;DR** is  magic.

If you comment out the `@SpringBootApplication` on `GraphiQLPathTestApplication` you will see failures like this: 

```java
Unable to find a @SpringBootConfiguration, you need to use @ContextConfiguration or @SpringBootTest(classes=...) with your test
java.lang.IllegalStateException: Unable to find a @SpringBootConfiguration, you need to use @ContextConfiguration or @SpringBootTest(classes=...) with your test
	at org.springframework.util.Assert.state(Assert.java:76)
	at org.springframework.boot.test.context.SpringBootTestContextBootstrapper.getOrFindConfigurationClasses(SpringBootTestContextBootstrapper.java:235)
	at org.springframework.boot.test.context.SpringBootTestContextBootstrapper.processMergedContextConfiguration(SpringBootTestContextBootstrapper.java:151)
	at org.springframework.test.context.support.AbstractTestContextBootstrapper.buildMergedContextConfiguration(AbstractTestContextBootstrapper.java:395)
```

The reason is:

> When testing Spring Boot applications, this is often not required. Spring Boot’s @*Test annotations search for your primary configuration automatically whenever you do not explicitly define one.
> 
> The search algorithm works up from the package that contains the test until it finds a class annotated with @SpringBootApplication or @SpringBootConfiguration. As long as you structured your code in a sensible way, your main configuration is usually found.
> 

Taken from [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-testing-spring-boot-applications-detecting-config). 

While we could specify it explicitly w/ `@SpringBootApplication(classes = [ ....`, that mechanism disables the discovery of other auto-configs and results in having to list more auto-configs in that `classes` attribute. 

I added a comment.

graphql-dgs-graphiql-autoconfigure/src/main/java/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLConfigurationProperties.kt
```diff
@@ -0,0 +1,32 @@
+/*
```
paulbakker, line 1: Kotlin sources should be in `src/main/kotlin` (it's currently in the `java` dir)

paulbakker (changes requested):
Outstanding work! Thanks a lot.
I think this is a better approach. Nice to see both and compare.

* The Graphiql depending on the webmvc module - I think that's ok. It might be a minor problem for when we support WebFlux, but I'm not sure it would've worked previously either. We'll look at that when we get there.

Could you move the Kotlin sources to the kotlin (instead of Java) folder? That's really all the feedback I have on the code, it looks great. I'll merge this as soon as that's done.

graphql-dgs-graphiql-autoconfigure/src/main/java/com/netflix/graphql/dgs/graphiql/autoconfiguration/GraphiQLConfigurationProperties.kt
berngp: Good catch!

bono007: Yeh those were already in there. I moved them in the subsequent PR and also made the packaging consistent `autoconfigure` / `autoconfiguration`.

paulbakker:
Closing this in favor of #153

PR #153, opened 2021-03-14
[Fine Grained Approach w/ GraphiQL Collapse] Replace path properties with @ConfigurationProperties

bono007:
Fixes gh-128

@paulbakker  @berngp this is the version of #152 that ALSO collapses the graphiql module into the webmvc module. 

I verified that we can still disable graphiql via the `dgs.graphql.graphiql.enabled` property.


paulbakker:
I'm not sure what to prefer, both would be ok to me. @bono007 @berngp What do you think?

berngp:
I think it is fine to merge them. The only thing that i was thinking overnight was... _how would this look like with webflux_. But I think we can review it when that time comes.

bono007:
Sorry for the lack of overall context w/ the 3 of these sitting around here for review. :) 

Like. you suggested @paulbakker - lets close the coarse-grained one. I like this fine-grained w/ collapse of graphiql  one myself.  As for WebFlux, I kept the exact constraints in place w/ the collapse - they should behave identical. 

If we want to not enable graphiql when using webflux - we can add the appropriate guards to the module. Typically its cleaner to have separate modules for mvc/webflux but it can be done in the same module - we do this in several core/shared starters at my company and it has worked out well. 

Also, if it comes down to us having to split this back out in some form due to WebFlux - I will be more than happy to do that.

paulbakker:
I'm ok with either approach, we'll work out the WebFlux question when we have it. Do you prefer this approach (merging the modules) over keeping it separate?

bono007:
> I'm ok with either approach, we'll work out the WebFlux question when we have it. Do you prefer this approach (merging the modules) over keeping it separate?

It's hard to tell much difference between the two approaches here in a merge request, but when pulled down in editor I think the collapsed is the cleaner of the two.

paulbakker:
Ok, let's go with that! Are you ready to merge (I think it looks good)

bono007:
Cool.  Lets do it

jandrade1:
@bono007 @paulbakker unfortunately this doesn't take into account that `server.servlet.context-path` may be configured to something other than `/`, which breaks GraphiQL since it's trying to find the GraphQL endpoint under `/graphql`. The controller is still working fine. I think the context path should probably be prepended to the path before replacing `<DGS_GRAPHQL_PATH>` in `GraphiQLConfigurer`.

bono007:
@jandrade1 good catch. The servlet context path is something I don't use that often so it's easy for me to forget that use case. Would you mind creating an issue for this and I will get to it in the next 24hrs? 

I bet this was an issue prior to any configurable paths. Since the path in graphiql was `/graphql`, if the context path was set to something other than `/` then it would not have been able to find it. The evolution of this feature has been:

I will probably leverage one of `ServletUriComponentsBuilder`, `UriComponentsBuilder` or just grab the "server.servlet.context-path" configuration property.

jandrade1:
thanks @bono007, created a new issue: https://github.com/Netflix/dgs-framework/issues/171

PR #161, opened 2021-03-16
Feature/update graphql client

srinivasankavitha:


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
```diff
@@ -40,7 +40,12 @@ fun interface RequestExecutor {
     fun execute(url: String, headers: Map<String, List<String>>, body: String): HttpResponse
 }
 
-data class HttpResponse(val statusCode: Int, val body: String?)
+data class HttpResponse(val statusCode: Int, val body: String?) {
+    var headers = mapOf<String, List<String>>()
+    constructor(statusCode: Int, body: String?, headers: Map<String, List<String>>) : this(statusCode, body) {
```
paulbakker, line 45: So this constructor is to keep backwards compatibility with existing Java users? 
Would be good to test that explicitly.

paulbakker (changes requested):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
```diff
@@ -40,7 +40,12 @@ fun interface RequestExecutor {
     fun execute(url: String, headers: Map<String, List<String>>, body: String): HttpResponse
 }
 
-data class HttpResponse(val statusCode: Int, val body: String?)
+data class HttpResponse(val statusCode: Int, val body: String?) {
+    var headers = mapOf<String, List<String>>()
```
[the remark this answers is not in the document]
srinivasankavitha: I can't because it's a secondary constructor field

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
srinivasankavitha: yes, to avoid breaking existing api.

srinivasankavitha: will add a test case.

srinivasankavitha: Added a separate test case for the response headers.

paulbakker (approved):

Issue #141, opened 2021-03-10
Support for changing the GraphQL schema location


marceloverdijk:
This is a follow up for https://github.com/Netflix/dgs-framework/issues/88 (closed).
DGS now already supports setting some options based on properties which is great.

It would be nice to be able to set the location of the `*.graphqls` schema files as well, like:

```
dgs.graphql.schema-location = classpath*:schema/**/*.graphql*
```

Rationale is that it currently looks in `schema` folder which by name is very generic.
I have some other (non GraphQL) schema related resources, so it would be nice to use a different folder.

marceloverdijk:
After #128 is done I can provide a PR for this as well if you prefer. Will be an easy one.

paulbakker:
Yes, that would be great! This is actually something that has come up many times already, so it's clearly something folks want/need.

marceloverdijk:
OK, I looked into it a bit but it might be not that easy/straightforward...

What needs to be done is to be able pass a `baseDir`  to the `DgsSchemaProvider`  so that can be used in the `findSchemaFiles`.
`DgsSchemaProvider` is part of the `graphql-dgs` submodule so that class should not depend on any autoconfiguration stuff nor configuration properties classes. So we should just pass a `baseDir: String` to `DgsSchemaProvider` as mentioned.

The `DgsSchemaProvider` is instantiated from the `graphql-dgs-spring-boot-oss-autoconfigure`  submodule, but this module has no access to the recently added `DgsWebMvcConfigurationProperties`. Nor would it be smart to add it here as it is `WebMvc`  related.

The separation of `graphql-dgs-spring-boot-webmvc-autoconfigure` and `graphql-dgs-spring-boot-oss-autoconfigure` does not make it easier, but I guess this is done as the latter one is swapped out at Netflix and replaced with a Netflix specific one.

2 options to implement this:

1.  Introduce a `DgsConfigurationProperties` class in `graphql-dgs-spring-boot-oss-autoconfigure` similar as `DgsWebMvcConfigurationProperties`.
2. Use a `@Value("${dgs.graphql.schema-base-dir:schema}")` var in `DgsAutoConfiguration` to pass to the `DgsSchemaProvider` instead.

@bono007 @paulbakker which option do you prefer?

bono007:
@marceloverdijk I was thinking about this ticket and was going to add some context this morning - but you beat me to it. And you found the thing I was going to write about. :) 

[...]

PR #165, opened 2021-03-17
Introduced schema-base-dir property for loading graphql(s) schema files

marceloverdijk:
PR for #141 

~~Note I also changed `DgsRestSchemaJsonController.kt` to re-use the schema as it was now creating another `GraphQLSchema` instance.~~

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -69,13 +69,13 @@ class DgsSchemaProvider(
 
     private val objectMapper = jacksonObjectMapper().registerModule(JavaTimeModule())
 
-    fun schema(schema: String? = null): GraphQLSchema {
+    fun schema(schema: String? = null, basedir: String = "schema"): GraphQLSchema {
         val startTime = System.currentTimeMillis()
         val dgsComponents = applicationContext.getBeansWithAnnotation(DgsComponent::class.java)
         val hasDynamicTypeRegistry = dgsComponents.values.any { it.javaClass.methods.any { m -> m.isAnnotationPresent(DgsTypeDefinitionRegistry::class.java) } }
 
         var mergedRegistry = if (schema == null) {
-            findSchemaFiles(hasDynamicTypeRegistry = hasDynamicTypeRegistry).map {
+            findSchemaFiles(basedir = basedir, hasDynamicTypeRegistry = hasDynamicTypeRegistry).map {
                 InputStreamReader(it.inputStream, StandardCharsets.UTF_8).use { reader -> SchemaParser().parse(reader) }
             }.fold(TypeDefinitionRegistry()) { a, b -> a.merge(b) }
         } else {
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -59,7 +59,8 @@ class DgsSchemaProvider(
     private val applicationContext: ApplicationContext,
     private val federationResolver: Optional<DgsFederationResolver>,
     private val existingTypeDefinitionRegistry: Optional<TypeDefinitionRegistry>,
-    private val mockProviders: Optional<Set<MockProvider>>
+    private val mockProviders: Optional<Set<MockProvider>>,
+    private val basedir: String = "schema"
 ) {
 
     val dataFetcherInstrumentationEnabled = mutableMapOf<String, Boolean>()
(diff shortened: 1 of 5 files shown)


paulbakker:
Looks like a compile error is happening: 

```
e: /home/runner/work/dgs-framework/dgs-framework/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt: (52, 48): Type mismatch: inferred type is DgsSchemaProvider but GraphQLSchema was expected
```

marceloverdijk:
Yes, let me check what went wrong.

marceloverdijk:
@paulbakker fixed now

paulbakker:
Tested and seems to work well!

One thing I started thinking about is if it should be possible to load from an "external"/non-classpath location. For example by providing an absolute path: 
```
dgs.graphql.schema-base-dir=/Users/pbakker/someschemadir
```
Thoughts?

While playing around with it, I also found that if you specify a directory that doesn't exist, it falls back to the generic error: 
```
Caused by: com.netflix.graphql.dgs.exceptions.NoSchemaFoundException: No schema files found. Define schemas in src/main/resources/schema/*.graphqls
```
That can probably be more specific when a directory is provided?

marceloverdijk:
Instead of a `schema-base-dir` property we could have a `schema-location-pattern`; with a default value of `"classpath*:schema/**/*.graphql*"`

This way files can also be loaded from e.g. non-classpath locations.
I thought about it but I think it needs a bit more of refactoring in the schema loading related methods.

It is more flexible for sure, so if you prefer I can update the PR.

paulbakker:
If you're up for it, I think it's interesting to explore if it's do-able. If it turns out to be too big/difficult of a change we can leave it for later, but it's probably easier to do it before folks start using the new config.

marceloverdijk:
Sure, I will try an alternative PR using that approach tomorrow; now watching the Dutch elections ;-)

paulbakker:
> Sure, I will try an alternative PR using that approach tomorrow; now watching the Dutch elections ;-)

Fun fact: you might or might not know that I'm Dutch as well. Although I've been in California for the past 4 years.

marceloverdijk:
Yes I know , that's why I mentioned it 😉

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -59,7 +59,8 @@ class DgsSchemaProvider(
     private val applicationContext: ApplicationContext,
     private val federationResolver: Optional<DgsFederationResolver>,
     private val existingTypeDefinitionRegistry: Optional<TypeDefinitionRegistry>,
-    private val mockProviders: Optional<Set<MockProvider>>
+    private val mockProviders: Optional<Set<MockProvider>>,
+    private val basedir: String = "schema"
```
bono007, line 63: @marceloverdijk I am sorry I totally missed this MR and just saw it now. Better late than never, eh? 

Can we remove this default from the constructor so that the default is only in the config props?

marceloverdijk: No problem. I will be working on an alternative PR this week anyway.

bono007: Cool. I look forward to it.

marceloverdijk:
@paulbakker I didn't got time last week, but while looking at the code yesterday I noticed this `DgsSchemaProvider` `findSchemaFiles` method:

```
        val testSchemas = try {
            resolver.getResources("classpath:$basedir-test/**/*.graphql*")
        } catch (ex: Exception) {
            arrayOf<Resource>()
        }

        val metaInfSchemas = try {
            resolver.getResources("classpath*:META-INF/$basedir/**/*.graphql*")
        } catch (ex: Exception) {
            arrayOf<Resource>()
        }
```

Whether to go for a `schema-base-dir` or "resource" `schema-location`s approach this might be impacting.

I think the `$basedir-test` can be removed. With a `schema-location` property users can define multiple locations so such a "test" folder could added by users - even only for a specific Spring profile.

But `"classpath*:META-INF/$basedir/**/*.graphql*"` is needed for the `graphql-error-types`, right?

2 options:

1. After loading the schema's based on the `schema-location` property, also `"classpath*:META-INF/$basedir/**/*.graphql*"` is loaded. 
2. We set the default value for the `schema-location` property to `["classpath*:schema/**/*.graphql*", "classpath*:META-INF/schema/**/*.graphql*"]` 

With the second approach we cannot distinguish whether the user has provided schema's, hence a `NoSchemaFoundException` will not happen anymore. So from that perspective the first option would be favorable imo.

Wdyt?

marceloverdijk:
PR #210 created with option 1.

marceloverdijk:
Closing this one as superseded by #210

PR #166, opened 2021-03-17
Add support for application/graphql requests

paulbakker:
Implements support for #34, which is further discussed here: https://github.com/Netflix/dgs-framework/discussions/34


paulbakker:
For context: this allows for a request to be sent from for example Intellij like this:

```
POST http://localhost:8080/graphql
Content-Type: application/graphql

{
   hello
}
```

Note that the body is not JSON like it normally would.

srinivasankavitha (approved):

PR #176, opened 2021-03-21
Add validation to DgsWebMvcConfigurationProperties (manual approach)

bono007:
This adds validation to the `DgsWebMvcConfigurationProperties` config props. 

ℹ️ this uses a manual set of checks implemented inside a `@PostConstruct` method on the config props. 
**pros/cons**
* (+) simple, no new deps required
* (-) not extensible, will have to implement each new validation we add to the config props

ℹ️ an alternate approach using JSR-303 annotations is #179

#### Startup failure looks like this:

image

@paulbakker @berngp - which approach are you in favor of?

### DIFF
--- a/graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcConfigurationPropertiesTest.kt
+++ b/graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcConfigurationPropertiesTest.kt
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.webmvc.autoconfigure
 
-import org.assertj.core.api.Assertions
+import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.springframework.boot.context.properties.bind.Binder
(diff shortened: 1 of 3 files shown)


graphql-dgs-spring-webmvc-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcConfigurationPropertiesTest.kt
```diff
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.webmvc.autoconfigure
 
-import org.assertj.core.api.Assertions
+import org.assertj.core.api.Assertions.assertThat
```
bono007, line 19: **[UNRELATED]**: sorry for sneaking this in, but its cleaning up the `assertThat` so it does not use `Assertions.` on every `assertThat` call.

berngp: don't mind that, I'll do it in every chance i get.

bono007:
@paulbakker @berngp I removed the WIP since we closed #179 in favor of this approach.

berngp (approved):
Thanks @bono007 !

PR #179, opened 2021-03-21
WIP: Add validation to DgsWebMvcConfigurationProperties (JSR-303 approach) 

bono007:
This adds validation to the `DgsWebMvcConfigurationProperties` config props via JSR-303 annotations.

ℹ️ this uses [@Valdiated](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config-validation) and JSR-303 validation annotations 
**pros/cons**
* (+) declarative w/ many [built-in validators](https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#section-builtin-constraints)
* (+) failures give exact context of where in application.yml the problem value is etc..
* (-) requires `kotlin-allopen` plugin to make config props non-final as it now gets cglib'ed via `@Validated`
* (-) requires JSR-303 impl on classpath - default is `hibernate-validator` which is in Spring Boot BOM and has nothing to do w/ Hibernate JPA/ORM. 

ℹ️ an alternate manual approach is #176 

--- 
#### Startup failure looks like this:

--- 

@paulbakker @berngp - which approach are you in favor of?

### DIFF
--- a/graphql-dgs-spring-webmvc-autoconfigure/build.gradle.kts
+++ b/graphql-dgs-spring-webmvc-autoconfigure/build.gradle.kts
@@ -14,9 +14,14 @@
  * limitations under the License.
  */
 
+plugins {
+    id("org.jetbrains.kotlin.plugin.allopen") version Versions.KOTLIN_VERSION
+}
+
 dependencies {
     api(project(":graphql-dgs"))
     api(project(":graphql-dgs-spring-webmvc"))
+    api("org.hibernate.validator:hibernate-validator")
     implementation("org.springframework.boot:spring-boot-starter")
     implementation("org.springframework:spring-webmvc")
     implementation("jakarta.servlet:jakarta.servlet-api")
@@ -24,4 +29,8 @@ dependencies {
     testImplementation("io.mockk:mockk:1.10.3-jdk8")
     testImplementation("org.springframework.boot:spring-boot-starter-web")
     testImplementation(project(":graphql-dgs-spring-boot-oss-autoconfigure"))
+}
+
+configure<org.jetbrains.kotlin.allopen.gradle.AllOpenExtension> {
+    annotations("org.springframework.boot.context.properties.ConfigurationProperties")
 }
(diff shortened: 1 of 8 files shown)


graphql-dgs-spring-webmvc-autoconfigure/build.gradle.kts
```diff
@@ -14,9 +14,14 @@
...
  */
 
+plugins {
+    id("org.jetbrains.kotlin.plugin.allopen") version Versions.KOTLIN_VERSION
+}
+
 dependencies {
     api(project(":graphql-dgs"))
     api(project(":graphql-dgs-spring-webmvc"))
+    api("org.hibernate.validator:hibernate-validator")
```
berngp, line 24: can it be `implementation`?

graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcConfigurationProperties.kt
```diff
@@ -19,31 +19,51 @@ package com.netflix.graphql.dgs.webmvc.autoconfigure
...
+        @DefaultValue("/schema.json") @field:ValidPath val path: String
+    )
+
+    /**
+     * Convenience annotation to prevent having to repeat the @Pattern for every path field.
+     */
+    @kotlin.annotation.Target(AnnotationTarget.FIELD)
+    @kotlin.annotation.Retention(AnnotationRetention.RUNTIME)
+    @Constraint(validatedBy = [])
+    @Pattern(regexp = "^/.*[^/]\$", message = "path must start with '/' and not end with '/'")
```
berngp, line 63: _side note_, there is no [org.hibernate.validator.constraints](https://docs.jboss.org/hibernate/validator/7.0/api/) that already does this _path_ validation.

paulbakker:
I’m really on the fence about this one because it pulls in another dependency (I know it’s just the validator, but still). Any thoughts on doing validation manually instead? It’s just config properties, so it feels a bit heavy handed to pull in something for it.

graphql-dgs-spring-webmvc-autoconfigure/build.gradle.kts
bono007: It could, but if apps did not bring it in then it would bypass validation.

bono007:
> I’m really on the fence about this one because it pulls in another dependency (I know it’s just the validator, but still). 

Me too. :) 

>Any thoughts on doing validation manually instead? It’s just config properties, so it feels a bit heavy handed to pull in something for it.

Yep, if this was just an app (not an OSS framework) this would be a no-brainer. We use it internally at my company and its really nice (when more complex validation is needed). For now, its literally path validation. 

I think we should close this PR and use #176 instead. We know what this looks like now though if at later date we think its warranted.

paulbakker:
Thanks @bono007, this was a useful experiment and discussion at least 👍

Issue #130, opened 2021-03-08
Unsafe TypedGraphQLError constants (INTERNAL, NOT_FOUND...)


kiov:
`TypedGraphQLError` has some statics fields like `INTERNAL`, `NOT_FOUND`, an so on.

The problem is that they reuse the same builder instance:

```java
    public static Builder INTERNAL = newBuilder().errorType(ErrorType.INTERNAL);
```
`Builder` instances are not thread safe and could share state between threads.

Multiples threads could be using the same unsafe instance.

Test code:
```java
final Semaphore s1 = new Semaphore(0);
final Semaphore s2 = new Semaphore(0);

ForkJoinTask<TypedGraphQLError> t1 = ForkJoinPool.commonPool().submit(() -> {
  Builder t = TypedGraphQLError.INTERNAL.extensions(Map.of("t", "1"));
  // allows t2 to run
  s2.release();

  //wait for t2
  s1.acquire();
  return t.build();
});
ForkJoinTask<TypedGraphQLError> t2 = ForkJoinPool.commonPool().submit(() -> {
  // ensures that t1 runs first
  s2.acquire();
  Builder t = TypedGraphQLError.INTERNAL.extensions(Map.of("t", "2"));
  // allows to t1 to return
  s1.release();
  return t.build();
});
TypedGraphQLError thread1Result = t1.join();
TypedGraphQLError thread2Result = t2.join();

if (thread1Result.getExtensions().get("t").equals("1")) {
  System.out.println("Thread 1: OK");
} else {
  System.out.println("Thread 1: Race condition, should be 1 instead of " + thread1Result.getExtensions().get("t"));
}

if (thread2Result.getExtensions().get("t").equals("2")) {
  System.out.println("Thread 2: OK");
} else {
  System.out.println("Thread 2: Race condition, should be 2 instead of " + thread2Result.getExtensions().get("t"));
}
```
Prints:
```
Thread 1: Race condition, should be 1 instead of 2
Thread 2: OK
```

Also this effectively allows to remap error as other types:
```java
TypedGraphQLError internalError = TypedGraphQLError.INTERNAL.build();
if (internalError.getExtensions().get("errorType").equals("INTERNAL")) {
  System.out.println("OK: INTERNAL error is INTERNAL");
} else {
  System.out.println("ERROR: INTERNAL error is " + internalError.getExtensions().get("errorType"));
}
// remap internal error as BAD_REQUEST
[...]

PR #184, opened 2021-03-23
Fix for #130

paulbakker:
Deprecate all static builder fields in TypedGraphQLError since they are thread-unsafe. Added a few builder shortcuts for the most common error types.

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

The static builder fields in TypedGraphQLError are thread-unsafe and should be removed. Because users may depend on these fields directly, I've deprecated the fields, and replaced them with a few convenience builder shortcuts.

Issue #130

### DIFF
--- a/graphql-error-types/src/main/java/com/netflix/graphql/types/errors/TypedGraphQLError.java
+++ b/graphql-error-types/src/main/java/com/netflix/graphql/types/errors/TypedGraphQLError.java
@@ -70,46 +70,118 @@ public class TypedGraphQLError implements GraphQLError {
         return extensions;
     }
 
+    /**
+     * @deprecated Static builder fields are thread-unsafe. Use TypedGraphQLError.newBuilder() instead.
+     */
     @Deprecated
     public static Builder UNKNOWN = newBuilder();
+
+    /**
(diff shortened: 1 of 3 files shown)


berngp (approved):

graphql-error-types/src/main/java/com/netflix/graphql/types/errors/TypedGraphQLError.java
```diff
@@ -70,30 +70,84 @@ public ErrorClassification getErrorType() {
         return extensions;
     }
 
+    @Deprecated
```
berngp, line 73: Should we add a comment that the static builders are not _thread-safe_?

srinivasankavitha (approved):

graphql-error-types/src/main/java/com/netflix/graphql/types/errors/TypedGraphQLError.java
paulbakker: Java 8 doesn't have an argument to the annotation, so added javadoc.

Issue #92, opened 2021-02-28
Shorthand Annotations


aramir:
In an effort to make code less verbose and easier to read: in addition to `@DgsData` annotation please consider adding the following two:

 - `@DgsQuery`
 - `@DgsMutation`

which are to serve as shorthands for `parentType="Query"` and `parentType="Mutation"`, respectively.

Further, it would be great to have a `field` parameter for the above annotations (as well as for `@DgsData`) as optional and have it to default to the annotated method's name.

This way the following annotation:
```java
@DgsData(parentType = "Query", field = "shows")
public List<Show> shows()
```
Becomes:
```java
@DgsQuery
public List<Show> shows()
```
Which is much more compact and follows the same approach that Spring Framework takes for composed annotations `@GetMapping`/`@PostMapping` vs `@RequestMapping(...)`

paulbakker:
`@DgsData` is used both to define queries/mutations, but also for field datafetchers on other types. Would you still use `@DgsData` for those? And the new annotations would just be shortcuts for query/mutation fields? For example: 

```
@DgsQuery
List<Show> shows() {}

@DgsData(type = "Show", field = "director")
Director director() {}
```

I'll give it some more thought, but I think I like the idea.

aramir:
> `@DgsData` is used both to define queries/mutations, but also for field datafetchers on other types. Would you still use `@DgsData` for those?

Of course! Just additional shorthands - but with `parentType = Query/Mutation` being the most common ones out there - that would eliminate a lot of boilerplate.

golfingal72:
2 Other Ideas:

1). We have a union type class, a field from the abstract maps to different target field names.. this way, no need to
write 5 resolvers

@DgsDatas(datas = {
            @DgsData(field = "productId", parentType = FavoriteProductDto.class),
            @DgsData(field = "brandId", parentType = FavoriteBrandDto.class),
            @DgsData(field = "retailerId", parentType = FavoriteRetailerDto.class),
            @DgsData(field = "userId", parentType = FavoriteUserDto.class),
            @DgsData(field = "postId", parentType = FavoritePostDto.class)
    })
   public Object favObjectIdFetcher(DataFetchingEnvironment environment) {
        log.trace("favObjectIdFetcher()");
        return ((Favorite) environment.getSource()).getReferencedId();
    }

2.  Same target field into different schema objects.. again no need to write 5 different resolvers ( I suppose if you implemented option 1, then the below could be satisfied with that

[...]

PR #190, opened 2021-03-24
Added @DgsMutation, @DgsQuery and @DgsSubscription

paulbakker:
Added @DgsMutation, @DgsQuery and @DgsSubscription as shorthand annotations. Also, the "field" argument on @DgsData is now optional, it will use the method name if no field name is provided.

All the examples below have the same effect: A `hello` datafetcher on the `Query` type.
The same mechanism also works for `@DgsMutation` and `@DgsSubscription`, but for their respective parent types.

Annotation only.
```java
 @DgsQuery
public String hello() {
      return "Hello!";
}
```

Field name specified.
```java
 @DgsQuery(field = "hello")
public String someMethod() {
      return "Hello!";
}
```

On top of that, the `field` argument on  `@DgsData` is now optional. If not provided, the method name is used.
```java
@DgsData(parentType = "Query")
public String hello() { 
    return "Hello!";
}
```

This implements feature request #92

### DIFF
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
@@ -49,14 +49,14 @@ internal class DgsSchemaProviderTest {
     @MockK
     lateinit var applicationContextMock: ApplicationContext
 
-    val defaultHelloFetcher = object : Any() {
+    private val defaultHelloFetcher = object : Any() {
         @DgsData(parentType = "Query", field = "hello")
         fun someFetcher(): String {
             return "Hello"
         }
     }
 
-    val defaultVideoFetcher = object : Any() {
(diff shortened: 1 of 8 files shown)


berngp (approved):
Great feature.

srinivasankavitha (approved):

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
```diff
@@ -522,6 +522,143 @@ internal class DgsSchemaProviderTest {
         assertThat(provider.dataFetcherInstrumentationEnabled["Show.title"]).isFalse
     }
```
srinivasankavitha, line 524: Should we add a test for the subscription annotation as well?

paulbakker: good point, added

PR #195, opened 2021-03-24
Allow specifying the operation name and query names

lucastorri:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

Hi All,

this PR allows specifying a query name and also an `operationName` in the request body, as allowed by the spec. Here are some relevant references:

- Query name:
  - <https://spec.graphql.org/June2018/#sec-Language.Operations>
- `operationName` in request body
  - <https://spec.graphql.org/June2018/#sec-Executing-Requests>
  - <https://graphql.org/learn/serving-over-http/#post-request>

I believe that the PR introduces no backward compatibility issues and that the test cases are sufficient, but I would be glad if someone else could check and confirm that.

Thank you very much!

Alternatives considered
----

The name could be added to `GraphQLQueryRequest`.

I would be happy to hear any suggestions.

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
@@ -54,7 +54,26 @@ class DefaultGraphQLClient(private val url: String) : GraphQLClient, MonoGraphQL
             "Content-type" to listOf("application/json")
         )
 
-        private data class Request(val query: String, val variables: Map<String, Any>)
+        private data class Request(val query: String, val variables: Map<String, Any>, val operationName: String?)
+    }
+
+    /**
+     * Executes a query and returns a GraphQLResponse.
+     * The actual HTTP request is done by an implementation of RequestExecutor, which is user provided.
+     * The RequestExecutor is typically provided as a lambda expression.
+     * The Accept and Content-Type headers are set. Additional headers can be set in the RequestExecutor.
+     * @param query The Query as a String
+     * @param variables Query variables. May be empty
+     * @param operationName optional operation name
+     * @param requestExecutor The code that does the actual HTTP request. Typically provided as a lambda expression.
+     * @return GraphQLResponse
+     * @throws GraphQLClientException when the HTTP response code is not 2xx.
+     */
+    override fun executeQuery(query: String, variables: Map<String, Any>, operationName: String?, requestExecutor: RequestExecutor): GraphQLResponse {
+        val serializedRequest = objectMapper.writeValueAsString(Request(query, variables, operationName))
+
+        val response = requestExecutor.execute(url, defaultHeaders, serializedRequest)
+        return handleResponse(response, serializedRequest)
     }
 
     /**
@@ -69,10 +88,7 @@ class DefaultGraphQLClient(private val url: String) : GraphQLClient, MonoGraphQL
      * @throws GraphQLClientException when the HTTP response code is not 2xx.
      */
     override fun executeQuery(query: String, variables: Map<String, Any>, requestExecutor: RequestExecutor): GraphQLResponse {
-        val serializedRequest = objectMapper.writeValueAsString(Request(query, variables))
-
-        val response = requestExecutor.execute(url, defaultHeaders, serializedRequest)
-        return handleResponse(response, serializedRequest)
+        return executeQuery(query, variables, null, requestExecutor)
     }
 
     /**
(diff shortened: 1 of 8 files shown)


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
```diff
@@ -54,7 +54,7 @@ class DefaultGraphQLClient(private val url: String) : GraphQLClient, MonoGraphQL
             "Content-type" to listOf("application/json")
         )
```
srinivasankavitha, line 56: Better to add another constructor with the additional name parameter for compatibility with java.

srinivasankavitha (changes requested):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
lucastorri: Hi @srinivasankavitha, thanks for reviewing this!

Given that `Request` is private, wouldn't it only be accessible inside this class? Perhaps the behavior differs between Kotlin and Java?

Anyhow, I'll add another constructor to it. Thanks once again!

lucastorri: Hi, I just wanted to double-check this. After adding the constructor, the IDE instantly tells me this constructor is never used. I also tried to use the class from Java to check what is the behavior and saw that the following would fail compilation, telling me that the constructor is private:

```java
new DefaultGraphQLClient.Companion.Request(/*...*/)
```

Thanks once again!

srinivasankavitha: Yes, you're right. Actually my comment would be more relevant for the GraphQLRequest, not so much the GraphQLClient. For the GraphQLRequest, we need 2 constructors for compatibility.

srinivasankavitha: I do question the need to update the GraphQLClient and the executeQuery methods altogether. We could just update the GraphQLRequest to take in the operation name and that should be sufficient, no? Am I missing something?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQuery.kt
```diff
@@ -18,9 +18,10 @@ package com.netflix.graphql.dgs.client.codegen
 
 import java.util.*
 
-abstract class GraphQLQuery(val operation: String) {
```
srinivasankavitha, line 21: We need the additional constructor here that takes in the name.

srinivasankavitha:
This would also need corresponding changes in code generation to actually take advantage of the new parameter, since most of our users just generate the request classes. However, that requires some careful coordination to not break compatibility between the 2.

Basically, GraphQLQueryRequest needs an additional constructor to take in the optional operationName so it doesn't break Java compatibility and existing codegen.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQuery.kt
lucastorri: Do you mean something like the following?

```kotlin
constructor(name: String) : this("query", name)
```

That would have the same signature as the existing following one, and unfortunately, it would not compile :(

```kotlin
constructor(operation: String) : this(operation, null)
```

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
lucastorri: They are different things. One is the name you give to a specific operation, and the other is the operation that you want to execute when you [send multiple operations](https://graphql.org/learn/serving-over-http/#post-request).

Take the examples below (I use `>...<` to highlight it). The first is in the query, e.g.:

```graphql
query >UserInformation< {
  me {
    id
    username
  }
}
```

The second in the actual request body ([some reasoning](https://stackoverflow.com/questions/49714484/why-would-you-send-multiple-operations-in-a-graphql-query-document-when-one-can)):

```json
{
  "operationName": ">A<",
  "query": "query A { ... }\n\nquery B { ... }",
  "variables": {}
}
```

lucastorri:
Hi @srinivasankavitha, I left a comment [here](https://github.com/Netflix/dgs-framework/pull/195#discussion_r602578354) explaining the difference between the two and why they are required. Please let me know what your thoughts are.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQuery.kt
srinivasankavitha: Sorry, what I mean for that is 
```
constructor(operation: String, name: String) : this("query", name)
``` 
in addition to 
```constructor(operation: String) : this(operation, null)
```

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
srinivasankavitha: I see. So this means the query executor would execute only query A in the above example? Why send multiple operations at all in that case?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQuery.kt
lucastorri: Hi @srinivasankavitha, I'm a bit confused.

Isn't that the primary constructor already? https://github.com/Netflix/dgs-framework/pull/195/files/5dde55c5a57c884e45ed961060dc3edac3fb63a8#diff-c4fd37e186aa99535318e2b88b6058255c651239802ca9a1df7cd181b857d835R21

Thanks!

srinivasankavitha: I missed that, sorry for the confusion (apparently viewing code on the phone is not ideal :)). That should take care of backwards compatibility in that case.

srinivasankavitha:
The changes look good. We will test these changes and merge the PR this week if it looks good.

srinivasankavitha:
Tested and seems to work well with Java code. The one missing piece is that we can't quite yet use the GraphQLRequest builder to build the query strings because it doesn't support specifying the operation name. It requires the user to build the the query string themselves. Which is fine I think for this pass. Because if we update the GraphQLRequest, we'll also need corresponding updates in codegen, and that would expand the scope of this PR.

PR #202, opened 2021-03-26
add DgsDataLoaderOptionsCustomizer

lmy654:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Be able to cusomize more about a DataLoader's DataLoaderOptions, #201


srinivasankavitha:
Could you also provide an example of what the DgsDataLoaderCustomizer would look like?  Also, the tests don't actually verify that the custom options are set. Is there a way to test for that?

lmy654:
> Could you also provide an example of what the DgsDataLoaderCustomizer would look like? Also, the tests don't actually verify that the custom options are set. Is there a way to test for that?

Yeah, actually in our system, we hava an demand to customize the implementation of `org.dataloader.CacheMap` and `org.dataloader.CacheKey` and config `DataLoaderOptions` with them like below:
```java
public class HotKeyCacheMap<U, V> implements CacheMap<U, V> {

    private Map<U, V> cache;

    private HotKeyCache hotKeyCache;

    /**
     * Default constructor
     */
    public HotKeyCacheMap(HotKeyCache hotKeyCache) {
        cache = new HashMap<>();
        this.hotKeyCache = hotKeyCache;
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public boolean containsKey(U key) {
        return cache.containsKey(key) || hotKeyCacheContainsKey(key);
    }

    private boolean hotKeyCacheContainsKey(U key){
        ValueModel valueModel = hotKeyCache.getValueModel(key.toString());
        return  valueModel != null && !valueModel.isDefaultValue();
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public V get(U key) {
        if(cache.containsKey(key)) {
            return cache.get(key);
        }
        else {
            return (V)hotKeyCache.getValueModel(key.toString()).getValue();
        }
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public CacheMap<U, V> set(U key, V value) {
        cache.put(key, value);
        hotKeyCache.setHotValue(key.toString(), value);
        return this;
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public CacheMap<U, V> delete(U key) {
        cache.remove(key);
        return this;
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public CacheMap<U, V> clear() {
        cache.clear();
        return this;
    }
}

@Configuration
public class HotkeyConfiguration {

    @Bean
    public DgsDataLoaderOptionsCustomizer dgsDataLoaderOptionsCustomizer(HotKeyCacheMapService hotKeyCacheMapService){
        return (dgsDataLoader, dataLoaderOptions) -> {
            dataLoaderOptions.setCacheMap(hotKeyCacheMapService.getCacheMap());
            dataLoaderOptions.setCacheKeyFunction(input -> dgsDataLoader.name() + "-" + input);
        };
    }
}
```
The `HotKeyCache` is a second level data cache to store some really "hot" data to protect downstream service from being overwhelmed, and we also need to customize the `org.dataloader.CacheKey` to distinct the key in different DataLoaders.

About the tests, I will provide a specific implementation to verify that custom options.

srinivasankavitha:
The changes are looking good to me.

srinivasankavitha:
@paulbakker @berngp - could you also take a look?

srinivasankavitha (approved):
Should we make this an annotation? Like this:
```
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Component
public @interface DgsDataLoaderOptionsCustomizer {
    String name();
}
```

srinivasankavitha:
Tested these changes and it works nicel.

paulbakker:
@lmy654 thanks for working on this. 

It looks like the mechanism is designed to have all dataloaders be configured in the same way. Is that correct? I'm wondering if it's common to have different configurations, eg. different caching strategies, for different dataloaders. I guess you could still do that based on the dataloader reference passed in into the customizer.

As an alternative, would something like a `@DgsDataLoaderOptionsCustomizer(MyCustomizer.class)` make sense?

Just bouncing ideas to get an understanding of what the best approach would be.

lmy654:
> @lmy654 thanks for working on this.
> 
> It looks like the mechanism is designed to have all dataloaders be configured in the same way. Is that correct? I'm wondering if it's common to have different configurations, eg. different caching strategies, for different dataloaders. I guess you could still do that based on the dataloader reference passed in into the customizer.
> 
> As an alternative, would something like a `@DgsDataLoaderOptionsCustomizer(MyCustomizer.class)` make sense?
> 
> Just bouncing ideas to get an understanding of what the best approach would be.

@paulbakker Thanks for review. I understand your concern. Before I thought configurations like cacheMap, cacheKeyFuntion should be common for dataloaders in most scenarios, whereas other configurations like batchingEnabled, maxBatchSize are not common and have been customized by DgsDataLoader's properties. Now I think again, as a framework, we should always give user an option. Although now it could be done based on the dataloader reference passed in into the customizer, it is not good way. 

As `@DgsDataLoader` has already some properties for customizing DataLoaderOptions, I think we could add another property specifing the bean name of `DgsDataLoaderOptionsCustomizer` and don't need to add another annotation. What do you think about?
I have just tried to refactor this, like below:
```java
@Target({ElementType.TYPE, ElementType.FIELD})
@Retention(RetentionPolicy.RUNTIME)
@Component
public @interface DgsDataLoader {
    String name();
    boolean caching() default true;
    boolean batching() default true;
    int maxBatchSize() default 0;
    String optionsCustomizerName() default "";
}
```
```kotlin
@DgsDataLoader(name = "exampleLoader", optionsCustomizerName = "exampleDgsDataLoaderOptionsCustomizer")
class ExampleBatchLoader : BatchLoader<String, String> {
    override fun load(keys: MutableList<String>?): CompletionStage<MutableList<String>> {
        return CompletableFuture.supplyAsync { mutableListOf("a", "b", "c") }
    }
}
class DgsDataLoaderProvider(private val applicationContext: ApplicationContext) {
    
    // omit some code...

    private fun createDataLoader(
        batchLoader: BatchLoader<*, *>,
        dgsDataLoader: DgsDataLoader
    ): DataLoader<out Any, out Any>? {
        val options = DataLoaderOptions.newOptions()
            .setBatchingEnabled(dgsDataLoader.batching)
            .setCachingEnabled(dgsDataLoader.caching)
        if (dgsDataLoader.maxBatchSize > 0) {
            options.setMaxBatchSize(dgsDataLoader.maxBatchSize)
        }
        if (dgsDataLoader.optionsCustomizerName.isNotBlank()) {
            applicationContext.getBean(dgsDataLoader.optionsCustomizerName, DgsDataLoaderOptionsCustomizer::class.java)
                .customize(dgsDataLoader, options)
        }
        val extendedBatchLoader = wrappedDataLoader(batchLoader, dgsDataLoader.name)
        return DataLoader.newDataLoader(extendedBatchLoader, options)
    }

    // omit some code
}
```

paulbakker:
@lmy654 I think specifying it in the annotation is a good way. I'm wondering if we can use a `Class` instead of the bean name though. Something like:

```
public @interface DgsDataLoader {
    String name();
    boolean caching() default true;
    boolean batching() default true;
    int maxBatchSize() default 0;
    Class<?> optionsCustomizer() default Object.class;
}
```

WDYT?

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/MultiDataLoaderInstrumentationProvidersProviderTest.kt
```diff
@@ -16,16 +16,9 @@
 
 package com.netflix.graphql.dgs.internal
 
-import com.netflix.graphql.dgs.DataLoaderInstrumentationExtensionProvider
-import com.netflix.graphql.dgs.ExampleBatchLoader
-import com.netflix.graphql.dgs.ExampleBatchLoaderWithContext
-import com.netflix.graphql.dgs.ExampleMappedBatchLoader
-import com.netflix.graphql.dgs.ExampleMappedBatchLoaderWithContext
+import com.netflix.graphql.dgs.*
```
sbilello, line 19: why replace with *?

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/MultiDataLoaderInstrumentationProvidersProviderTest.kt
```diff
@@ -16,16 +16,9 @@
...
 import org.assertj.core.api.Assertions.assertThat
-import org.dataloader.BatchLoader
-import org.dataloader.BatchLoaderWithContext
-import org.dataloader.MappedBatchLoader
-import org.dataloader.MappedBatchLoaderWithContext
+import org.dataloader.*
```
sbilello, line 21: why replace with * ?

prokop7:
Any progress on this?

evanmalmud:
Anything stopping this from being merged than the conflicts? 
Would love to have this feature to be able to use a CacheMap/ValueCache.

prokop7:
https://github.com/Netflix/dgs-framework/pull/1485
This pull request partially solves the problems.

> Would love to have this feature to be able to use a CacheMap/ValueCache.
You definitely can use DataLoaderOptionsProvider to utilize custom cachemap/valuecache

reneleonhardt:
@lmy654 Thank you for your contribution! Can the conflicts be resolved or is this feature not needed anymore?

reneleonhardt:
Thank you! Can you push your work? `This branch has conflicts that must be resolved` still displayed.

paulbakker:
Closing as stale.

evanmalmud:
Ticket has been open for over a year with the work completed and it gets marked as stale after a month for conflicts because of the above is crazy. 
I'll take a look at this, this week.

paulbakker:
Apologies @evanmalmud, I'm doing some cleanup of no longer relevant issues and didn't realize there was actually activity a few weeks ago. I got a bit too trigger-happy on the close button.

paulbakker:
@evanmalmud I'm not sure why this PR fell off the radar, but I went over it again to get context.
I'm wondering if the following is not sufficient, which I think didn't exist at the time the PR was originally created.

```java
 @Bean
    DgsDataLoaderOptionsProvider dgsDataLoaderOptionsProvider() {
        var defaults = new DefaultDataLoaderOptionsProvider();

        return ((dataLoaderName, annotation) -> switch (dataLoaderName) {
            case "messages" -> new DataLoaderOptions()
                    .setBatchingEnabled(false).setCachingEnabled(false).setMaxBatchSize(1);
            default -> defaults.getOptions(dataLoaderName, annotation);
        });
    }
```

Just one way to implement this, but basically I just use a switch to do "special" configuration for the "messages" dataloader, while defaulting for all others. Is this missing anything?

evanmalmud:
Fair enough I haven't used dgs in about 6 months. If something exists today to solve the issue then that's different haha.

paulbakker:
Well we really dropped the ball on the PR, so that's no excuse, but I *think* the available solution now is quite nice. If you agree I'll re-close the PR. If anyone else has reasons why the current solution isn't sufficient we can reconsider.

evanmalmud:
Seems like a valid solution you have there that solves the problem. I think the only thing this PR would allow you to do it split up the config in code. Not the end of the world but potentially helps code organization. I can also totally see the argument that the config is all in one place now too, but that isn't really the case for anything else in DGS.

paulbakker:
I would think this kind of custimization is fairly small (e.g. you typically have a pretty small set of dataloaders), so I don't yet see a need to split that further.
The one thing I didn't like in the PR is how it would need to tie the customizers to the datafetchers (and we didn't have a better solution either), so I'm happy if we can avoid that.

Closing the PR for now. If anyone comes to across this and has specific needs, please start a new discussion.
Thanks for the feedback @evanmalmud!

PR #203, opened 2021-03-26
Introduce a Platform BOM

berngp:
This commit introduces two new modules, the `graphql-dgs-platform-dependencies` and `graphql-dgs-platform`.
Both modules provide [Maven BOM]s intended to help projects keep a coordinate set of versions across
their dependencies, as they relate to DGS. The `graphql-dgs-platform` will only express recommendations for
`graphl-dgs` artifacts and a few other artifacts used by such modules as _APIs_.
In fact, the `graphql-dgs-platform` is used internally to reflect a consistent versions of dependencies
such as `com.graphql-java:graphql-java`. The `graphql-dgs-platform-dependencies` module, on the other hand,
is not used internally but is a superset and aimed on aiding developers that not only want to leverage the
GraphQL DGS framework, but are also interested in using the same recommendation that the DGS project uses internally.

Why would I be interested in the `graphql-dgs-platform-dependencies` BOM? Well, for example,
let's say that you want to leverage `spring-boot-starter-data-mongodb` and you are not sure which version to use.
Since the DGS Framework internally uses Spring Boot, and specially its BOM, it will provide a recommendation for you.

How can I use the recommendations from both, the `graphql-dgs-platform-dependencies` and `graphql-dgs-platform` modules?

In Maven you will have to add it as part of the `dependencyManagement`, let's take  `graphql-dgs-platform-dependencies` as an example and assume that we want to use the DGS version 3.10.0. You will add it as follows.

```xml
<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>com.netflix.graphql.dgs</groupId>
            <artifactId>graphql-dgs-platform-dependencies</artifactId>
            <version>3.10.0</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>
```

In Gradle you will define it as part of the `dependencyRecommendations` as follows,
note that the example is in Kotlin but there is a Groovy equivalent.

```kotlin
dependencyRecommendations {
    mavenBom(mapOf("module" to "com.netflix.graphql.dgs:graphql-dgs-platform-dependencies:3.10.0"))
}
```

References:
* [The Java Platform Plugin](https://docs.gradle.org/current/userguide/java_platform_plugin.html)
* [Maven Publish Plugin](https://docs.gradle.org/current/userguide/publishing_maven.html)
* [Maven BOM](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms)

---
[The Java Platform Plugin]: https://docs.gradle.org/current/userguide/java_platform_plugin.html
[Maven Publish Plugin]: https://docs.gradle.org/current/userguide/publishing_maven.html
[Maven BOM]: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-


graphql-dgs-platform-dependencies/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+dependencies {
+    api(platform(project(":graphql-dgs-platform")))
+}
+
+/* ----------------------------------------------------------- */
+// Define Exclusions...
+// Placeholder for external modules we should exclude, add "group:name", e.g. "com.google.code.guice:guice"
+val ignoreExternalModules by extra(emptyList<String>())
```
paulbakker, line 50: Wondering if we should keep this there isn't anything concrete to put there?

berngp: I agree but I want us to be aware as well, in case is needed int he future.

PR #214, opened 2021-03-29
Fixing BOM Nebula Metadata

berngp:
This PR fixes the BOM Metadata that is published by Nebula.

### DIFF
--- a/graphql-dgs-platform/build.gradle.kts
+++ b/graphql-dgs-platform/build.gradle.kts
@@ -17,13 +17,14 @@
 plugins {
     `java-platform`
     `maven-publish`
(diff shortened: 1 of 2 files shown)


paulbakker (approved):

graphql-dgs-platform/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import org.gradle.kotlin.dsl.codegen.apiTypeProviderFor
```
berngp, line 1: This is not needed, removing.

srinivasankavitha (approved):

PR #232, opened 2021-04-03
Define a recommendation for graphql-java-extended-scalars

berngp:
Pull Request type
----
- [x] Feature

Changes in this PR
----

This commit adds a recommendation for graphql-java-extended-scalars that will accommodate the requirements of compatibly in relationship with graphql-java. The latter is already part of the recommendations.

Per graphql-java-extended-scalars

> use 1.0.1 or above for graphql-java 14.x and above
> use 15.0.0 or above for graphql-java 15.x and above
> use 16.0.0 or above for graphql-java 16.x and above

Reference https://github.com/graphql-java/graphql-java-extended-scalars

### Note
The graphql-java-extended-scalars 16.0.0 artifact is not published in Maven Central. 
For this reason we are recommending 15.0.0, with a _preferred_ scope, until  https://github.com/graphql-java/graphql-java-extended-scalars/issues/38 is fixed.

Alternatives considered
----

We could not have a recommendation, but this will mean that developers will then have to figure out which version of `graphql-java-extended-scalars` to use according to the version of `graphql-java` that we do recommend.

### DIFF
--- a/graphql-dgs-platform/build.gradle.kts
+++ b/graphql-dgs-platform/build.gradle.kts
@@ -48,6 +48,13 @@ dependencies {
         api("com.graphql-java:graphql-java") {
             version { require(Versions.GRAPHQL_JAVA) }
         }
+        api("com.graphql-java:graphql-java-extended-scalars") {
+            // Note that the version of graphql-java should dictate the version of the scalars,
+            // but until https://github.com/graphql-java/graphql-java-extended-scalars/issues/38 is addressed we
+            // are preferring 15.0.0.
+            // Ref. https://github.com/graphql-java/graphql-java-extended-scalars
+            version { prefer("15.0.0") }
+        }
         api("com.apollographql.federation:federation-graphql-java-support") {
             version { require(Versions.GRAPHQL_JAVA_FEDERATION) }
         }


berngp:
The `graphql-java-extended-scalars` `16.0.0` artifact is not published in Maven Central. We can't merge this as is before this gets addressed.
Ref https://github.com/graphql-java/graphql-java-extended-scalars/issues/38

graphql-dgs-platform/build.gradle.kts
```diff
@@ -46,10 +46,17 @@ dependencies {
     // For more information at https://docs.gradle.org/current/userguide/rich_versions.html
     constraints {
         api("com.graphql-java:graphql-java") {
-            version { require(Versions.GRAPHQL_JAVA) }
```
paulbakker, line 49: Why get rid of the constant?

berngp: The version is no explicitly set in the BOM. IMO it is easier to visually identify the versions, specially if there is correlation between two artifacts. In the past there were several places where the version was required making the constant useful, now there is only one.

berngp:
> I assume it is compatible to use graphql-java 16 with graphql-java-extended-scalars 15?

It is as far as our examples and internal usage, in fact internally folks are using graphql-java-extended-scalars 1.0 and 1.1 while using graphql-java 16. That said in the `graphql-java-extended-scalars` page there is no references on backward compatibility and the statement so far is to use the version that matches the version used for graphql-java.

> use 1.0.1 or above for graphql-java 14.x and above
> use 15.0.0 or above for graphql-java 15.x and above
> use 16.0.0 or above for graphql-java 16.x and above

graphql-dgs-platform/build.gradle.kts
berngp: @paulbakker reverted the change with the constants. I also opened #243 to remove the rxjava recommendation.

srinivasankavitha (approved):

paulbakker (approved):

PR #242, opened 2021-04-06
Auto-Configuration that registers graphql-java-extended-scalars' Scalars

berngp:
Pull Request type
----
- [x] Feature

Dependencies
----

* Ideally the https://github.com/Netflix/dgs-framework/pull/232 should be merged first.

Changes in this PR
----

Via the new graphql-dgs-extended-scalars module developers can automatically register extended scalars.
It will add the `com.graphql-java:graphql-java-extended-scalars` artifact as a `compile`/`api` dependency.
The feature can be disabled by setting the `dgs.graphql.extensions.scalars.enabled` flag to `false`.

Developers can disable automatic registration of specific scalar extensions as they relate to types:

* Time and Dates via can be disabled by setting `dgs.graphql.extensions.scalars.time-dates.enabled` to `false`,
* JSON, URL, Object, and Locale can be disabled by setting `dgs.graphql.extensions.scalars.objects.enabled` to `false`,
* All numeric extensions can be disabled by setting `dgs.graphql.extensions.scalars.numbers.enabled` to `false`,
* Character extensions, just ExtendedScalars.GraphQLChar for now, can be disabled by setting `dgs.graphql.extensions.scalars.chars.enabled` to `false`.

Other Notes
---

* We are grouping the Scalar Extensions in buckets, such as `numbers`, `objects`, `time-dates`, and `chars`. This is a bit arbitrary but better ideas evade me for the time being.


paulbakker:
We already have `@DgsTypeDefinitionRegistry` and `@DgsCodeRegistry`, which are mechanisms to hook into the runtime wiring. Why not use that? 

For reference: https://netflix.github.io/dgs/advanced/dynamic-schemas/

berngp:
@paulbakker so the implementation uses the `DgsRuntimeWiring` to add the scalars. The `RuntimeWiring.Builder` for this particular change is made available via auto-wiring to facilitate testing. I'm taking a look at `@DgsCodeRegistry ` and `@DgsTypeDefinitionRegistry` and I don't think they provide mechanism to assert which scalars have been added vi a `@DgsRuntimeWiring`.

paulbakker:
What do you mean by "spy"? Just for testing?

berngp:
@paulbakker per our offline discussion...
1. Will move this feature to its own module, and therefore enable the `dgs.graphql.extensions.scalars.enabled` flag by default.
2. Remove the `RuntimeWiring.Builder` from being auto-wire and move to a Smoke Test that asserts the scalars.
3. POM change that we discussed will be reflected in #232.

berngp:
@paulbakker  all of the items are done.

- [x] 1. Will move this feature to its own module, and therefore enable the dgs.graphql.extensions.scalars.enabled flag by default.
- [x] 2. Remove the RuntimeWiring.Builder from being auto-wire and move to a Smoke Test that asserts the scalars.
- [x] 3. POM change that we discussed will be reflected in #232.

paulbakker (approved):

Issue #229, opened 2021-04-01
bug: InvocationTargetException when a GraphQLError is thrown for a federated entity query


marcelvanderperk:
## Expected behavior
That my GraphQLError is returned to the client.

## Actual behavior
The GraphQLError gets wrapped in a InvocationTargetException and hides the real exception from the client. I have written a workaround to unwrap this exception in a DataFetcherExceptionHandler, but I think this should work for everyone without having to write a custom exception handler.

## Steps to reproduce

I'm using the latest version 3.10.2 and its easily reproducible with this basic example; 

```java
  @DgsEntityFetcher(name = "Image")
  public Image getImageEntity(Map<String, Object> values) {
    throw new ImageDoesNotExist();
  }

  public static class ImageDoesNotExist extends RuntimeException implements GraphQLError {

    @Override
    public String getMessage() {
      return "Blabla";
    }

    @Override
    public List<SourceLocation> getLocations() {
      return Collections.emptyList();
    }

    @Override
    public ErrorClassification getErrorType() {
      return ErrorType.NOT_FOUND;
    }
  }
```

The Query
```json
query ($representations: [_Any!]!) {
    _entities(representations: $representations) {
        ... on Image {
            id 
            url
        }
    }
}
```

GraphQL variable

```json
{
    "representations": [
        {
            "__typename": "Image",
            "id": "34b85b92-e219-4fbb-8bf2-682c4172682b"
        }
    ]
}
```

This results in.
```json
{
    "errors": [
        {
            "message": "java.lang.reflect.InvocationTargetException: null",
            "locations": [],
            "path": [
                "_entities"
            ],
            "extensions": {
                "errorType": "INTERNAL"
            }
        }
    ],
    "data": null
}
```

My current workaround by unwrapping the InvocationTargetException in a custom DataFetcherExceptionHandler.

```java
@Component
public class DgsExceptionHandler implements DataFetcherExceptionHandler {
  private final DefaultDataFetcherExceptionHandler defaultHandler = new DefaultDataFetcherExceptionHandler();

  @Override
  public DataFetcherExceptionHandlerResult onException(DataFetcherExceptionHandlerParameters handlerParameters) {
    //exceptions seem to get wrapped in a InvocationTargetException
    if (handlerParameters.getException() instanceof InvocationTargetException) {
      InvocationTargetException invocationTargetException = (InvocationTargetException) handlerParameters.getException();
      Throwable targetException = invocationTargetException.getTargetException();
[...]

PR #253, opened 2021-04-10
Exceptions thrown by DgsEntityFetcher should be handled

srinivasankavitha:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
DgsFederationResolver throws an InvocationTargetException if the invoked DgsEntityFetcher throws any exception. We need to unwrap the InvocationTargetException to get the relevant information in the GraphQL response (already handled by the DefaultExceptionHandler). Otherwise, the actual exception information is lost and the response just contains the InvocationTargetException.

Fixes Issue  https://github.com/Netflix/dgs-framework/issues/229

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
@@ -29,6 +29,8 @@ import graphql.schema.TypeResolver
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
+import java.lang.IllegalArgumentException
+import java.lang.reflect.InvocationTargetException
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
```diff
@@ -70,22 +71,25 @@ open class DefaultDgsFederationResolver() :
...
-                if (result == null) {
-                    throw MissingDgsEntityFetcherException(typename.toString())
-                }
-
-                if (result is CompletionStage<*>) {
-                    result.toCompletableFuture()
-                } else {
-                    CompletableFuture.completedFuture(result)
+                } catch (e: InvocationTargetException) {
+                    throw e.targetException
```
paulbakker, line 92: Should it check if targetException is not null? I'm thinking about the case where the arguments of the method are of the wrong type. In that case the actual exception *is* `InvocationTargetException`. Maybe handle that specifically?

srinivasankavitha: Good call, will fix.

PR #266, opened 2021-04-14
Throw an error for queries on subscription endpoint without explicit operation type.

srinivasankavitha:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Check for the operation type in a subscription query and return an error. If not specified, the operation is assumed to be a query and the error is misleading.

### DIFF
--- a/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
+++ b/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
@@ -20,6 +20,8 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.netflix.graphql.dgs.DgsQueryExecutor
 import graphql.ExecutionResult
+import graphql.language.OperationDefinition
+import graphql.parser.Parser
 import graphql.validation.ValidationError
 import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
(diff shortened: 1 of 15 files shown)


graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -59,6 +59,12 @@ class DgsSSESubscriptionHandler(private val dgsQueryExecutor: DgsQueryExecutor)
             return ResponseEntity.badRequest().body(emitter)
         }
 
+        if (! query.contains("subscription")) {
```
paulbakker, line 62: Can't the query be parsed to check for the operation name instead of just doing a contains?

srinivasankavitha: the operation name is different. It applies to ```subscription ABC { ...}``` where ABC is the operation name.

srinivasankavitha: Do you mean using the SchemaParser and parsing the operation definition?

paulbakker: Not `SchemaParser`, because it's about parsing the query, not the schema. But with a `contains`, a query like the following would be accepted: `query SomeQuery { membership { subscriptionId }}`, while it clearly shouldn't.

srinivasankavitha:
Dropping this PR after discussion, we can simply rely on the ClassCastException to handle this scenario.

PR #271, opened 2021-04-15
Bugfix: Allow JSON Documents with DateTime scalars to be deserialized to Objects

### LINKED ISSUE #269: bug: dgsQueryExecutor.executeAndExtractJsonPathAsObject cannot deserialize DateTime scalars
## Expected behavior

The framework should serialize the json response to an object, including DateTime scalar values.

## Actual behavior

Jackson's ObjectMapper does not know about the scalar and a DgsQueryExecutionDataExtractionException is thrown.

## Steps to reproduce

Please see the issue I opened in [dgs-examples-java](https://github.com/Netflix/dgs-examples-java/issues/7).

richardcresswell:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Allow the DgsQueryExecutor to deserialize JSON responses to Objects for documents containing scalar DateTime types.
Resolves #269

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt
@@ -18,6 +18,7 @@ package com.netflix.graphql.dgs.internal
 
 import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.ObjectMapper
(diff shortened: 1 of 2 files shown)


paulbakker:
We have the same in `GraphQLResponse` in the `graphql-dgs-client` module, so it makes sense to support the same here.

srinivasankavitha:
Agree with @paulbakker. We actually also do the same thing in DgsSchemaProvider.kt as well. I wonder if we can use a common object mapper, at least within the same core module.

PR #275, opened 2021-04-16
Add open keyword to class definition of @RestController classes

lucatk:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

The classes `DgsRestController`, `DgsRestSchemaJsonController`, and `DgsSSESubscriptionHandler` have been changed to be declared as `open class`.

We (company) use Spring AOP along with [CGLIB proxying](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-scopes-other-injection-proxies), which causes issues if beans on final classes are introduced. I am aware this is candidate for improvement on our side, however it's an easy fix if the remaining 3 beans in dgs-framework were changed to be open classes.
This might improve usability for other users with side dependencies like Spring AOP aswell.
As far as I am aware, this change doesn't introduce any side effects or change in functionality of the library.

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -62,7 +62,7 @@ import org.springframework.web.multipart.MultipartFile
  */
 
 @RestController
-class DgsRestController(private val dgsQueryExecutor: DgsQueryExecutor) {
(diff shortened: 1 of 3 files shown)


paulbakker:
For context, we had some discussion about this.

I'm really not in favor of opening classes for extension that aren't explicitly designed for extension. Making a class extensible makes it part of the API, which implies we need to be careful about backwards compatibility. 
At the same time, I do understand the proxy/aop use case, and also understand that's a different type of use than manually extending a class.

We'll merge this, and I'll follow up by adding some JavaDocs to the classes specifying that they are *not* considered API, and can be changed at any time.

Issue #213, opened 2021-03-29
Auto configuration modules have inconsistent package naming


bono007:
| ℹ️ This is not quite a feature request, but not a bug either. |
| --- |

* WebMvc has the expected "autoconfigure" package name
https://github.com/Netflix/dgs-framework/tree/master/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure
* OSS is named "autoconfig"
https://github.com/Netflix/dgs-framework/tree/master/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig
* SSE does not have an autoconfigure pacakge
https://github.com/Netflix/dgs-framework/tree/master/graphql-dgs-subscriptions-sse-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse
* Websocket does not have an autoconfigure package
https://github.com/Netflix/dgs-framework/tree/master/graphql-dgs-subscriptions-websockets-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/websockets

Lets make these consistent and use the "standard" naming scheme for an auto configure module  `PROJECT_PACKAGE_NAME.autoconfigure`.

PR #279, opened 2021-04-17
Rename packages from autoconfig -> autoconfigure

bono007:
@paulbakker @berngp long time no see 👋🏻. I have been meaning to circle back and make this refactor - the day job been keeping me too busy 😼 

Fixes gh-213

### DIFF
--- a/graphql-dgs-example-java/src/test/java/com/netflix/graphql/dgs/example/ExampleSpringBootTest.java
+++ b/graphql-dgs-example-java/src/test/java/com/netflix/graphql/dgs/example/ExampleSpringBootTest.java
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.example;
 
-import com.netflix.graphql.dgs.autoconfig.DgsAutoConfiguration;
+import com.netflix.graphql.dgs.autoconfigure.DgsAutoConfiguration;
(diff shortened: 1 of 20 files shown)


berngp:
Thanks @bono007, and sorry for taking some time to look into this.

bono007:
> Thanks @bono007, and sorry for taking some time to look into this.

NP @berngp . I am going to try to make some additional contributions this weekend as well.

berngp:
Rolling back this change. Will reconsider when we have our 4.x release.

bono007:
@berngp I am guessing people got hit by the package name changes? I had some original concern about this but I think we decided to make the move as it was in the auto-config classes. But yeh, I think this is better suited for 4.x as well.

paulbakker:
@bono007  What we forgot about is that folks use the DgsAutoconfig class in `@SpringBootTest(classes = ...)`. Other than that you woulder depend on it directly, but it didn't seem right to break test compilation in a minor release.

berngp:
@bono007 we discussed adding a _test slice_ to prevent folks from directly referencing the auto-configuration. Regardless I think we can include this change for the 4.x release.

bono007:
Agreed on both of your points @paulbakker and @berngp .

Issue #290, opened 2021-04-23
bug: Java client query serialization issue with Extended Scalar DateTime


amitgupta1202:
## Expected behavior

We are using extended Scalar DateTime as an input parameter. 

GraphQL Schema

        scalar DateTime

        type Query {
            exampleQuery(someDateTime: DateTime!): [String]!
        }

Implementation:

```
@DgsComponent
public class ScalarConfig {
    @DgsRuntimeWiring
    public RuntimeWiring.Builder addScalar(RuntimeWiring.Builder builder) {
        return builder.scalar(ExtendedScalars.DateTime);
    }
}

@DgsComponent
public class SupplierQueryResolver {

    @DgsQuery
    public String exampleQuery(@InputArgument(value = "someDateTime") OffsetDateTime someDateTime) {
        return "hello";
    }
}
```

We  make a call using generated Java client and expect response back as "hello".

## Actual behavior
When we execute the query using the generated Java Client we get error 
`com.netflix.graphql.dgs.exceptions.QueryException: Invalid Syntax : Invalid syntax: token recognition error at: '-01' at line 1 column 39`

## Steps to reproduce
```
@SpringBootTest(webEnvironment = RANDOM_PORT)
public class ExampleExtendedScalarDateTime {

    @Autowired
    private DgsQueryExecutor dgsQueryExecutor;

    @Test
    void extended_scalar_DateTime_serialization_issue() {
        var epochDateTime = OffsetDateTime.ofInstant(Instant.EPOCH, ZoneId.from(ZoneOffset.UTC));
        var query = ExampleQueryGraphQLQuery.newRequest().someDateTime(epochDateTime).build();
        var request = new GraphQLQueryRequest(query);

        var jsonResult = dgsQueryExecutor.executeAndGetDocumentContext(request.serialize()).jsonString();
        assertThat(jsonResult).isEqualTo("{\"data\":{\"exampleQuery\":\"hello\"}}");
    }
}
```

Above test fails with exception described above.

amitgupta1202:
The issue seems to be in the Code GraphQLQueryRequest:55, only String is serialised by wrapping value double quote, where are rest of value are not, which in case of Extended Scalar Types is definitely not true. Types Such as OffsetDateTime, OffSetTime, LocalDate, Locale will produce a String value.

amitgupta1202:
I have got a temporary fix for it as patch file https://gist.github.com/amitgupta1202/c0649cc41395fc4cda12912a0832b871

[...]

PR #292, opened 2021-04-24
Serialize extended scalar types OffsetDateTime, OffsetTime, LocalDate…

amitgupta1202:
…, Locale as String value wrapped in double quote.

Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Fixes java client serialisation issue to extended scalar types OffsetDateTime, OffsetTime, LocalDate and Locale.

Issue #
https://github.com/Netflix/dgs-framework/issues/290

Alternatives considered
----
New feature that allows for custom serializers this gives full control on how the objects will be serialized.

I have an example on the branch explaining the approach
https://github.com/amitgupta1202/dgs-framework/tree/feature/custom_serializer

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
@@ -16,6 +16,11 @@
 
 package com.netflix.graphql.dgs.client.codegen
 
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.time.OffsetTime
+import java.util.*
+
 class GraphQLQueryRequest(private val query: GraphQLQuery, private val projection: BaseProjectionNode?) {
 
     constructor(query: GraphQLQuery) : this(query, null)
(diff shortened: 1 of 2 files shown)


amitgupta1202:
Couldn't reproduce the test failure, I have rebased the branch from latest master.

berngp:
Thanks for the PR @amitgupta1202. I think adding OffsetDateTime, OffsetTime, LocalDate and Locale is sensible, that said we suggest to leverage the _code generation_ plugins that will properly serialize more complex data. More information [here](https://netflix.github.io/dgs/generating-code-from-schema/).

paulbakker:
I have some concerns about this being a workaround for a bigger problem that needs attention because I think the same issue applies to _all_ custom scalar types.

I need to look into the problem a little more to form an opinion, which I'll do later this week. We might just merge this, but I need to experiment a little to know what's the right thing to do.

amitgupta1202:
I think I agree with @paulbakker, It seems to be wider issue. I also had another idea which I posted on original pull request https://github.com/amitgupta1202/dgs-framework/tree/feature/custom_serializer

amitgupta1202:
> I have some concerns about this being a workaround for a bigger problem that needs attention because I think the same issue applies to _all_ custom scalar types.
> 
> I need to look into the problem a little more to form an opinion, which I'll do later this week. We might just merge this, but I need to experiment a little to know what's the right thing to do.

@paulbakker did you manage to look into it? i do agree its a general bigger  issue, i did experiment with the custom serializer, issuehttps://github.com/amitgupta1202/dgs-framework/tree/feature/custom_serializer not sure its good enough. Plus i think we can support all extended scalar type, without need to custom serializer?

paulbakker:
@amitgupta1202, Sorry for the slow response - the Webflux stuff kept me busy for a while.

I've experimented a bit and came up with this PR proposal: https://github.com/Netflix/dgs-framework/pull/333. WDYT?

amitgupta1202:
Idea seems great. further comment on https://github.com/Netflix/dgs-framework/pull/333

amitgupta1202:
Different implementation provided via issue  https://github.com/Netflix/dgs-framework/pull/333

PR #300, opened 2021-04-28
Implement calculation of query complexity metric.

srinivasankavitha:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Query Complexity is the total number of nodes in the query. `graphql-java` provides a MaxQueryComplexityInstrumentation class that users can add and set limits on the query complexity (https://www.javadoc.io/doc/com.graphql-java/graphql-java/2019-07-13T23-07-14-8c71e18/graphql/analysis/MaxQueryComplexityInstrumentation.html)

This PR implements the query complexity calculation and sets the tag `gql.queryComplexity` with the value on the `gql.query` metric. To do so, we override the beginValidation() in the instrumentation class and implement the complexity by traversing the nodes using the same implementation as linked below:
Reference Implementation : https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java

In addition, we add the result of the query complexity calculation to our execution state for tagging as a metric.

Note that this does not provide any ability to limit the complexity of the query (as intended by MaxQueryComplexityInstrumentation). This implementation is simply for the framework to track the complexity for metrics.

Alternatives considered
----
I looked into using the MaxQueryComplexityFunction (provided as lambda) to extract the result of the query complexity computation, but this does not allow us to set the result in any execution specific state. This would have ideally avoided re-implementing beginValidation(). Since we don't have access to any execution specific state, I resorted to the approach above.

### DIFF
--- a/graphql-dgs-example-java/build.gradle.kts
+++ b/graphql-dgs-example-java/build.gradle.kts
@@ -19,7 +19,6 @@ dependencies {
     implementation(project(":graphql-dgs-subscriptions-websockets-autoconfigure"))
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("io.projectreactor:reactor-core")
-
     // Enabling Spring Boot Actuators. We are not expressing any opinion on which Metric System should be used
     // please review the Spring Boot Docs in the link below for additional information.
     // https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-metrics-export-simple
@@ -27,5 +26,4 @@ dependencies {
     // Adding the DGS Micrometer integration that provides timers for gql.*
     // For additional information go to the link below:
     // https://netflix.github.io/dgs/advanced/instrumentation/
-    implementation(project(":graphql-dgs-spring-boot-micrometer"))
 }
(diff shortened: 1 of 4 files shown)


graphql-dgs-example-java/build.gradle.kts
```diff
@@ -19,4 +19,6 @@ dependencies {
     implementation(project(":graphql-dgs-subscriptions-websockets-autoconfigure"))
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("io.projectreactor:reactor-core")
+    implementation(project(":graphql-dgs-spring-boot-micrometer"))
```
berngp, line 22: you will not this anymore as part of the pr. I just merged https://github.com/Netflix/dgs-framework/pull/297.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -102,6 +112,52 @@ class DgsGraphQLMetricsInstrumentation(
         }
     }
 
+    /**
+     * Port the implementation from MaxQueryComplexityInstrumentation in graphql-java and store the computed complexity
+     * in the MetricsInstrumentationState for access to add tags to metrics.
+     */
+    override fun beginValidation(parameters: InstrumentationValidationParameters): InstrumentationContext<List<ValidationError>> {
```
berngp, line 119: Should we send a PR to the graphql-java folks such that they expose this as part of a utility such that it allows some sort of composability that we can leverage?

berngp (approved):

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
srinivasankavitha: yeah, that makes sense.

graphql-dgs-example-java/build.gradle.kts
antholeole: This PR removed micrometer from the example project.

srinivasankavitha: Sound good! Thanks, that would be helpful!

PR #301, opened 2021-04-29
WebFlux support

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Added support for WebFlux. Four new modules are added:

* **graphql-dgs-reactive**: Contains reactive variants of core interfaces such as `DgsReactiveQueryExecutor` and `DgsReactiveCustomContextBuilderWithRequest`. 
* **graphql-dgs-spring-webflux-autoconfigure**: WebFlux based HTTP endpoints.
* **graphql-dgs-webflux-starter**: Starter using Webflux modules
* **graphql-dgs-example-java-webflux**: Copy of the example application, but using the Webflux starter. Where necessary it also uses new reactive interfaces, such as for custom context.

There are now two interfaces to execute queries: the original `DgsQueryExecutor` and the new `DgsReactiveQueryExecutor`. All the methods are the same, except that `DgsReactiveQueryExecutor` wraps results in a `Mono`.
When using the WebFlux modules, both interfaces are available for injection. The WebFlux HTTP endpoints internally use the `DgsReactiveQueryExecutor`, but a user could still write tests using the blocking `DgsQueryExecutor`. This should make migration to WebFlux from an existing code base easier.

There are also two interfaces for creating a custom context. The original `DgsCustomContextBuilderWithRequest` and `DgsReactiveCustomContextBuilderWithRequest`. 
The new `DgsReactiveCustomContextBuilderWithRequest` has the same functionality, but accepts a `ServerRequest` instead of a `WebRequest`, and returns its result as a `Mono`.
**A `DgsCustomContextBuilderWithRequest` will not be picked up by the WebFlux stack, because the types are incompatible.** 

**Backward incompatible changes**

* The already deprecated `DgsContextBuilder` interface has been removed. Its use was already discouraged since introducing custom context.
* `DgsRequestData` is now an interface. The interface has only `extensions` and `headers`, but not `WebRequest` like it previously had. When using WebMVC the framework will use `DgsWebMvcRequestData` which adds `WebRequest` as a value, and WebFlux uses `DgsWebfluxRequestData` which uses `ServerRequest`. This does mean the user needs to type check to get to webRequest/serverRequest. See below for details.

```kotlin


interface DgsRequestData {
    val extensions: Map<String, Any>?
    val headers: HttpHeaders?
}

data class DgsWebMvcRequestData(
    override val extensions: Map<String, Any>? = null,
    override val headers: HttpHeaders? = null,
    val webRequest: WebRequest? = null,
) : DgsRequestData
```

Example use: 

```kotlin
@DgsData(parentType = "Query", field = "usingWebRequest")
fun usingWebRequest(dfe: DgsDataFetchingEnvironment): String {
     return ((DgsContext.getRequestData(dfe) as DgsWebMvcRequestData)?.webRequest as ServletWebRequest).request.serverName
}
```

Missing features
----

* Websockets/SSE is currently not supported (subscriptions). This should be added before releasing this.
* The GraphiQL endpoint can't be changed. 

Alternatives considered
----

The chosen implementation is a trade-off between keeping backward compatibility, while fully embracing WebFlux if the user opt-in. Because of type incompatibilities between WebFlux and WebMVC, introducing new interfaces seemed like the best approach. The downside is that this may be confusing, but it's a similar approach as what the Spring Framework uses.


ooga:
Really nice work 👍 

Could it be possible to also support Mono as return value in dataloader and datafetcher ?
Adding an example of security (https://netflix.github.io/dgs/advanced/security/) would be awesome

graphql-dgs-example-java-webflux/src/main/java/com/netflix/graphql/dgs/example/ExtraTypeDefinitionRegistry.java
```diff
@@ -0,0 +1,40 @@
...
+import graphql.language.TypeName;
+import graphql.schema.idl.TypeDefinitionRegistry;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ExtraTypeDefinitionRegistry {
+    @Bean
+    public TypeDefinitionRegistry registry() {
+        ObjectTypeExtensionDefinition objectTypeExtensionDefinition = ObjectTypeExtensionDefinition.newObjectTypeExtensionDefinition().name("Query").fieldDefinition(FieldDefinition.newFieldDefinition().name("myField").type(new TypeName("String")).build())
```
Nokinori, line 30: really long line) Mb put on separate lines?

graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/DefaultDgsReactiveQueryExecutor.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    override fun execute(
+        query: String,
+        variables: MutableMap<String, Any>?,
+        extensions: MutableMap<String, Any>?,
+        headers: HttpHeaders?,
+        operationName: String?,
+        serverHttpRequest: ServerRequest?
+    ): Mono<ExecutionResult> {
+        val graphQLSchema: GraphQLSchema =
```
Nokinori, line 64: Why not do this in one chain? Then all actions will be done not in Assembly time
Like this:
```
Mono.fromCallable {
        if (reloadIndicator.reloadSchema()). schema.updateAndGet { schemaProvider.schema() }
        else schema.get()
}
.zipWith( contextBuilder.build(DgsReactiveRequestData(extensions, headers, serverHttpRequest)) )
.flatMap { 
    val graphQlSchema = it.t1
    val dsgContext = it.t2
    Mono.fromCompletionStage(...)
}
```

graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/DefaultDgsReactiveQueryExecutor.kt
```diff
@@ -0,0 +1,153 @@
...
+            }
+    }
+
+    private fun getJsonResult(query: String, variables: Map<String, Any>): Mono<String> {
+        return execute(query, variables).map { executionResult ->
+            if (executionResult.errors.size > 0) {
+                throw QueryException(executionResult.errors)
+            }
+
+            val objectMapper = ObjectMapper()
```
Nokinori, line 149: Do we really need to create `ObjectMapper ` here every time?
Also, mb we can provide an ability to inject a custom realization?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
```diff
@@ -0,0 +1,194 @@
...
+                        .build()
+                val execute = graphQL.executeAsync(executionInput)
+
+                return@GET Mono.fromCompletionStage(execute)
+                    .map { it.toSpecification() }
+                    .flatMap { ok().json().bodyValue(it) }
+            }.build()
+    }
+
+    class GraphQlHandler(private val dgsQueryExecutor: DgsReactiveQueryExecutor) {
```
Nokinori, line 149: Don't you think about move it to another class an make it open? Then inject in here.
I suppose, other people will be glad to add additional logic in it)

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
```diff
@@ -0,0 +1,194 @@
...
+
+                        dgsQueryExecutor.execute(
+                            queryInput.query,
+                            queryInput.queryVariables,
+                            queryInput.extensions,
+                            request.headers().asHttpHeaders(),
+                            "",
+                            request
+                        )
+                    }.subscribeOn(Schedulers.parallel())
```
Nokinori, line 180: Can you explain, please. Why do we need here an parallel scheduler? What the benefit of using it?
I can't find any explanation in source code(

graphql-dgs-spring-webflux-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/WebRequestTestWithCustomContext.kt
```diff
@@ -0,0 +1,134 @@
...
+                            .type(TypeName("String"))
+                            .build()
+                    ).fieldDefinition(
+                        FieldDefinition
+                            .newFieldDefinition()
+                            .name("withHeader")
+                            .type(TypeName("String"))
+                            .build()
+                    )
+
```
Nokinori, line 101: empty line)

graphql-dgs-webflux-starter/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+dependencies {
+    api(project(":graphql-dgs-spring-boot-oss-autoconfigure"))
+    api(project(":graphql-dgs-spring-webflux-autoconfigure"))
+    api(project(":graphql-dgs-client"))
+    api(project(":graphql-error-types"))
+
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
```
Nokinori, line 23: So I need to add dependency in my project to `graphql-dgs-webflux-starter` and  `org.springframework.boot:spring-boot-starter-webflux`, am I right?
wouldn't it be better to mark is as `api` dependency? So after that  `graphql-dgs-webflux-starter` will be enough.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt
```diff
@@ -217,3 +162,65 @@ class DefaultDgsQueryExecutor(
         fun reloadSchema(): Boolean
     }
 }
+
+class BaseDgsQueryExecutor {
+    companion object {
```
Nokinori, line 167: It looks like an candidat for object.
```suggestion
object BaseDgsQueryExecutor {
```

Nokinori (commented):
Thank you for such magnificent work!) Really appreciate it 👍

srinivasankavitha:
Can we share code between the example apps? Seems like a lot of it is the same?

paulbakker:
Added websocket support for subscriptions as well now. Tests are still missing, but it works with the example app.

paulbakker:
Websocket tests added as well now. Will process review comments tomorrow, and that should wrap things up for now.

graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/DefaultDgsReactiveGraphQLContextBuilder.kt
```diff
@@ -0,0 +1,75 @@
...
+                    dgsRequestData
+                )
+            )
+        }.switchIfEmpty(
+            Mono.just(
+                DgsContext(
+                    requestData = dgsRequestData
+                )
+            )
+        )
```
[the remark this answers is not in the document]
paulbakker: fixed

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
```diff
@@ -0,0 +1,214 @@
...
+import reactor.core.publisher.Mono
+import reactor.core.scheduler.Schedulers
+import java.net.URI
+import java.util.*
+
+@Suppress("SpringJavaInjectionPointsAutowiringInspection")
+@Configuration
+@EnableWebFlux
+@EnableConfigurationProperties(DgsWebfluxConfigurationProperties::class)
+open class DgsWebFluxAutoConfiguration(private val configProps: DgsWebfluxConfigurationProperties) {
```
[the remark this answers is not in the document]
berngp: Another thing that we can do, specially if we remove the `@EnableWebFlux` annotation, is provide a _Test Slice_ that mainly does

```
@EnableWebFlux
@SpringBootTest(
    classes = [
        DgsWebFluxAutoConfiguration::class,
        DgsAutoConfiguration::class,
    ]
)
```

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
berngp: I'm also having trouble understanding the value of `.subscribeOn(Schedulers.parallel())`. Also note that this _Scheduler_ is not a _TaskScheduler_ managed by Spring/Spring Boot, it will not be able to leverage the _delegating task executors_ that different Spring libraries might need. For example, in Spring Security there is the `DelegatingSecurityContextAsyncTaskExecutor` that makes sure a _SecurityContext_ is available.

graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/DefaultDgsReactiveQueryExecutor.kt
paulbakker: fixed

graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/DefaultDgsReactiveQueryExecutor.kt
paulbakker: It's now using the objectmapper created in the `BaseDgsQueryExecutor` companion object; that way it's also using the same configuration as when using JsonPath. 

We can consider making it injectable later.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
paulbakker: Moved to another class for clarity. Leaving injection (as a hook to override) for later, as there are some decisions to make related to it.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
paulbakker: That shouldn't have been there, it's a leftover of something I was trying.

graphql-dgs-webflux-starter/build.gradle
paulbakker: done

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt
paulbakker: Done

paulbakker:
> Can we share code between the example apps? Seems like a lot of it is the same, except the ContextBuilder and the build.gradle?

Yes, I've moved the code to `graphql-dgs-example-shared` now, which both examples depend on.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -50,6 +46,7 @@ import java.util.*
  * Framework auto configuration based on open source Spring only, without Netflix integrations.
  * This does NOT have logging, tracing, metrics and security integration.
  */
+@Suppress("SpringJavaInjectionPointsAutowiringInspection")
```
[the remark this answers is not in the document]
paulbakker: You can technically use the framework core without being in a web environment. The web stuff is all in separate modules.

graphql-dgs-example-java-webflux/src/main/resources/schema/schema.graphqls
```diff
@@ -0,0 +1,82 @@
...
+type Rating {
+    avgStars: Float
+}
+
+input FileUploadInput {
+    description: String
+    files: [Upload]
+}
+
+#metatron curl -a exampledgs https://localhost:8443/subscriptions\?query\=ewoicXVlcnkiOiAic3Vic2NyaXB0aW9uIHsgIHN0b2NrcyB7bmFtZX0gfSIKfQ==
```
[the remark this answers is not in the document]
paulbakker: done

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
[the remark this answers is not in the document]
paulbakker: Removed, and added to the tests.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsReactiveWebsocketHandler.kt
```diff
@@ -0,0 +1,165 @@
...
+                session.bufferFactory(),
+                resolvableType,
+                MimeTypeUtils.APPLICATION_JSON,
+                null
+            )
+        )
+    }
+}
+
+const val GQL_CONNECTION_INIT = "connection_init"
```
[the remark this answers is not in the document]
paulbakker: done

paulbakker:
@Nokinori @berngp @srinivasankavitha Thanks for all the comments. I've addressed them all, except comments about the example app code. We should fix the example app code as well, but that's separate from the webflux work.

paulbakker:
> Really nice work 👍
> 
> Could it be possible to also support Mono as return value in dataloader and datafetcher ?
> Adding an example of security (https://netflix.github.io/dgs/advanced/security/) would be awesome

@ooga Good idea, but I'll keep that as a separate todo to keep things scoped a little bit.

berngp (approved):

graphql-dgs-reactive/build.gradle
```diff
@@ -0,0 +1,24 @@
+/*
```
berngp, line 1: @paulbakker could we make it build.gradle.kts to keep it consistent?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
```diff
@@ -0,0 +1,157 @@
...
+
+package com.netflix.graphql.dgs.webflux.autoconfiguration
+
+import com.netflix.graphql.dgs.internal.DefaultDgsQueryExecutor
+import com.netflix.graphql.dgs.internal.DgsDataLoaderProvider
+import com.netflix.graphql.dgs.internal.DgsSchemaProvider
+import com.netflix.graphql.dgs.reactive.DgsReactiveCustomContextBuilderWithRequest
+import com.netflix.graphql.dgs.reactive.DgsReactiveQueryExecutor
+import com.netflix.graphql.dgs.reactive.internal.DefaultDgsReactiveGraphQLContextBuilder
+import com.netflix.graphql.dgs.reactive.internal.DefaultDgsReactiveQueryExecutor
```
srinivasankavitha, line 25: Is "netfliux" intentional? Did you mean netflix?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflux/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+                        GQL_STOP -> {
+                            subscriptions[webSocketSession.id]?.get(operationMessage.id)?.cancel()
+                            subscriptions[webSocketSession.id]?.remove(operationMessage.id)
+                            logger.debug("Client stopped subscription ${operationMessage.id} for connection ${webSocketSession.id}")
+                            Flux.empty()
+                        }
+
+                        GQL_CONNECTION_TERMINATE -> {
+                            subscriptions[webSocketSession.id]?.values?.forEach { it.cancel() }
+                            webSocketSession.close()
```
srinivasankavitha, line 129: Remove the session from subscriptions as well? Like in line 122.

srinivasankavitha (changes requested):

graphql-dgs-reactive/build.gradle
paulbakker: Huh I hadn't even noticed. Should be consistent for sure.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsReactiveWebsocketHandler.kt
```diff
@@ -0,0 +1,143 @@
...
+
+                        GQL_STOP -> {
+                            subscriptions[webSocketSession.id]?.get(operationMessage.id)?.cancel()
+                            subscriptions[webSocketSession.id]?.remove(operationMessage.id)
+                            Flux.empty()
+                        }
+
+                        GQL_CONNECTION_TERMINATE -> {
+                            subscriptions[webSocketSession.id]?.values?.forEach { it.cancel() }
+                            webSocketSession.close()
```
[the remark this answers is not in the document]
paulbakker: The session itself should be removed. Added that now.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsReactiveWebsocketHandler.kt
```diff
@@ -0,0 +1,143 @@
...
+                                    val publisher: Publisher<Any> = executionResult.getData()
+                                    Flux.from(publisher).map {
+                                        toWebsocketMessage(
+                                            OperationMessage(GQL_DATA, it, operationMessage.id),
+                                            webSocketSession
+                                        )
+                                    }.doOnSubscribe {
+                                        if (operationMessage.id != null) {
+                                            subscriptions[webSocketSession.id] = mutableMapOf(operationMessage.id to it)
+                                        }
```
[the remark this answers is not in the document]
paulbakker: In both cases is does this: 

```
  subscriptions[webSocketSession.id]?.remove(operationMessage.id)
```

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflux/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
paulbakker: done

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
paulbakker: fixed

srinivasankavitha (approved):
Looks really good to me!

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflux/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+        const val GQL_START = "start"
+        const val GQL_STOP = "stop"
+        const val GQL_DATA = "data"
+        const val GQL_ERROR = "error"
+        const val GQL_COMPLETE = "complete"
+        const val GQL_CONNECTION_TERMINATE = "connection_terminate"
+    }
+
+    override fun handle(webSocketSession: WebSocketSession): Mono<Void> {
+        return webSocketSession.send(
```
Nokinori: I think, it looks really complicated now. It hard to track flow
And also, as I remember from Reactor specification, it not good solution to call `.subscribe()` inside `.doOnXXX`

Mb it's better to separate input and output? And also change output to resulting command and separated publisher.
Like this:
```kotlin
val publisher: FluxProcessor  = EmitterProcessor() // or other new Sinks builder
val input: Mono<OperationMessage> = webSocketSession.receive().flatMap { .... decoder.decode(... } }
val output: Flux<WebSocketMessage> = 
  input.map { 
    when(om.type) {
        GQL_CONNECTION_INIT -> initConnection(om)
        ....
    } 
val sender = webSocketSession.send(Flux.from(publisher))
return Mono.zip(input, output, sender).then()

fun initConnection(om: OperationMessage): Flux<WebSocketMessage> = 
    Flux.just(publisher.offer(om.exctractInitPayload()))
```

But yes... it also going to be too complicated :)
Not in this PR, it require more investigation and discussions

Nokinori (approved):
@paulbakker Thanks for implementation!

PR #311, opened 2021-05-04
Test Example Application as part of the CI Process

berngp:
Summary
=======

This commit, via the ci-compatibility.yml workflow action, will use the build's snapshot version and apply it to a set of example applications. This will help us see if the current changes are compatible with an application using the previous released version.

Implementation
==============

Via the `scripts/test-examples.py` we are adding the ability to download defined git repositories that we want to use to assert the build changes. Before running a `clean build`, this script will update the downloaded projects' build file to use the specific dgs-framework snapshot generated from a local build.

Requirements of Example Projects.

1. Should build with the same JDK as this project. In this case JDK 1.8.
2. Should use the BOM, since its through the BOM how we control de versions.

Running The Example Tests Locally
=================================

1. You will need Python3 and have `pip3` and `python3` available in your
   path.
2. You will need to install the Python3 required modules defined in `scripts/requirements.txt`.
   You can just run `make install-py-libs`.
3. Run the `make test-examples` tasks.

You can also run `./scripts/test-examples.py` manually. Explore the options with the `-h` flag.

### DIFF
--- a/.github/workflows/ci-compatibility.yml
+++ b/.github/workflows/ci-compatibility.yml
@@ -1,0 +1,76 @@
+name: CI Compatibility Check
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk-version: [ 1.8 ]
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


srinivasankavitha (approved):
Looks good!

PR #315, opened 2021-05-06
Refactor the UI-Example to remove generated files from being stored in git

berngp:
POC to refactor the UI-Example to remove the generated files, via NPM, from being stored in GIT.


paulbakker:
Does this run on every build? That's what I tried to avoid because it adds to build time and might be problematic for running in different environments (e.g. our contributors machines).

berngp:
It does run on every build. Gradle will download nodjes if not available, so that will add time too.

graphql-dgs-example-java-webflux/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("org.springframework.boot") version "2.4.5"
```
berngp, line 18: It also turns out that using the _Spring Boot_ Plugin "2.4.5" will bump the Spring Boot version to 2.4.5 as well.

PR #327, opened 2021-05-11
Use AopUtils when looking for @DgsTypeResolver

paulbakker:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Account for (AOP) proxy classes when searching for `@DgsTypeResolver`.

Previously, when a `@DgsTypeResolver` was in the same class as a `@Secured` method, the type resolver would not be found. `@Secured` causes the class to be proxied, so we need to account for that when scanning for annotations.
We do this for other methods, but not for type resolvers.


berngp (approved):

srinivasankavitha (approved):
Do we have other places where this might also need to be fixed?

PR #333, opened 2021-05-11
Scalar in client

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Issue #290 

For query parameters of a complex scalar such as DateTime, code generation will create a query method using the Java type that represents the scalar. This is currently now correctly serialized.
Inputs are serialized using the objects `toString()` method. In some cases this may work, but in more complex cases (DateTime), it doesn't.

In a DGS, the serialization of scalars is implemented in a `Coercing` implementation. Because this code is already available, the same code should be used for serialization in the client.

In this PR a map of types mapped to a scalar implementation can optionally be passed into the `GraphQLQueryRequest`.
If an input has the type of one of the mapped types, its scalar will be used during serialization.

The following is an example.

```java
GraphQLQueryRequest request = new GraphQLQueryRequest(
                ReviewsGraphQLQuery.newRequest().dateRange(new DateRange(LocalDate.of(2020, 01, 01), LocalDate.now())).build(),
                new ReviewsProjectionRoot().submittedDate().starScore(),
                Map.of(DateRange.class, new DateRangeScalar())
        );
```

The following is the `DateRange` type and custom `Coercing` implementation.

```java
public class DateRange {
    private final LocalDate from;
    private final LocalDate to;

    public DateRange(LocalDate from, LocalDate to) {
        this.from = from;
        this.to = to;
    }

    public LocalDate getFrom() {
        return from;
    }

    public LocalDate getTo() {
        return to;
    }
}

public class DateRangeScalar implements Coercing<DateRange, String> {
    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MM/dd/yyyy");

    @Override
    public String serialize(Object dataFetcherResult) throws CoercingSerializeException {
        DateRange range = (DateRange) dataFetcherResult;

        return range.getFrom().format(formatter) + "-" + range.getTo().format(formatter);
    }

    @Override
    public DateRange parseValue(Object input) throws CoercingParseValueException {
        String[] split = ((String) input).split("-");
        LocalDate from = LocalDate.parse(split[0], formatter);
        LocalDate to = LocalDate.parse(split[0], formatter);
        return new DateRange(from, to);
    }

    @Override
    public DateRange parseLiteral(Object input) throws CoercingParseLiteralException {
        String[] split = ((StringValue)input).getValue().split("-");
        LocalDate from = LocalDate.parse(split[0], formatter);
        LocalDate to = LocalDate.parse(split[0], formatter);
        return new DateRange(from, to);
    }
}

```

When invoking the `request.serialize()` method, the `DateRangeScalar.serialize` method will be used to serialize the `DateRange` input.

**Optional extras**

We can consider providing a number of well-known scalars (date types, etc.) out of the box. This is not included in this PR yet.

Alternatives considered
----

This idea came out of this [PR](https://github.com/Netflix/dgs-framework/pull/292) which started some discussion how we could implement this in a more extendable way.

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
@@ -73,6 +73,7 @@ class InputValueSerializer(private val scalars: Map<Class<*>, Coercing<*, *>> = 
         }
 
         val type = input::class.java
+
         return if (scalars.contains(type)) {
             """"${scalars[type]!!.serialize(input)}""""
         } else if (type.isPrimitive || type == Integer::class.java || type == Long::class.java || type == Double::class.java || type == Float::class.java || type == Boolean::class.java || type == Short::class.java || type == Byte::class.java || type.isEnum) {
@@ -82,10 +83,10 @@ class InputValueSerializer(private val scalars: Map<Class<*>, Coercing<*, *>> = 
         } else if (input is List<*>) {
             """[${input.filterNotNull().joinToString(", ") { listItem -> serialize(listItem) ?: "" }}]"""
         } else {
-            input.javaClass.declaredFields.map {
+            input.javaClass.declaredFields.filter { !it.type::class.isCompanion }.map {
                 it.isAccessible = true
                 val nestedValue = it.get(input)
-                if (nestedValue != null) {
+                if (nestedValue != null && !nestedValue::class.isCompanion) {
                     """${it.name}:${serialize(nestedValue)}"""
                 } else {
                     null
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
@@ -18,10 +18,16 @@ package com.netflix.graphql.dgs.client.codegen
 
 import graphql.schema.Coercing
 
-class GraphQLQueryRequest(private val query: GraphQLQuery, private val projection: BaseProjectionNode?, private val scalars: Map<Class<*>, Coercing<*, *>>?) {
+class GraphQLQueryRequest(
+    private val query: GraphQLQuery,
+    private val projection: BaseProjectionNode?,
+    private val scalars: Map<Class<*>, Coercing<*, *>>?
+) {
 
     constructor(query: GraphQLQuery) : this(query, null, null)
     constructor(query: GraphQLQuery, projection: BaseProjectionNode?) : this(query, projection, null)
+
+    private val inputValueSerializer = InputValueSerializer(scalars ?: emptyMap())
 
     fun serialize(): String {
         val builder = StringBuilder()
@@ -39,29 +45,7 @@ class GraphQLQueryRequest(private val query: GraphQLQuery, private val projectio
                 if (value != null) {
                     builder.append(key)
                     builder.append(": ")
-                    if (value is String) {
-                        builder.append("\"")
-                        builder.append(value.toString())
-                        builder.append("\"")
-                    } else if (value is List<*>) {
-                        if (value.isNotEmpty() && value[0] is String) {
-                            builder.append("[")
-                            val result = value.joinToString(separator = "\", \"", prefix = "\"", postfix = "\"")
-                            builder.append(result)
-                            builder.append("]")
-                        } else {
-                            builder.append(value.toString())
-                        }
-                    } else {
-                        if (scalars?.contains(value::class.java) == true) {
-                            val serializedValue = scalars[value::class.java]!!.serialize(value)
-                            builder.append("\"")
-                            builder.append(serializedValue)
-                            builder.append("\"")
-                        } else {
-                            builder.append(value.toString())
-                        }
-                    }
+                    builder.append(inputValueSerializer.serialize(value))
                 }
                 if (inputEntryIterator.hasNext()) {
                     builder.append(", ")
@@ -79,5 +63,34 @@ class GraphQLQueryRequest(private val query: GraphQLQuery, private val projectio
 
         builder.append(" }")
         return builder.toString()
+    }
+}
+
+class InputValueSerializer(private val scalars: Map<Class<*>, Coercing<*, *>> = emptyMap()) {
+    fun serialize(input: Any?): String? {
+        if (input == null) {
+            return null
+        }
+
+        val type = input::class.java
+        return if (scalars.contains(type)) {
+            """"${scalars[type]!!.serialize(input)}""""
+        } else if (type.isPrimitive || type == Integer::class.java || type == Long::class.java || type == Double::class.java || type == Float::class.java || type == Boolean::class.java || type == Short::class.java || type == Byte::class.java || type.isEnum) {
+            input.toString()
+        } else if (type == String::class.java) {
+            """"$input""""
+        } else if (input is List<*>) {
+            """[${input.filterNotNull().joinToString(", ") { listItem -> serialize(listItem) ?: "" }}]"""
+        } else {
+            input.javaClass.declaredFields.map {
+                it.isAccessible = true
+                val nestedValue = it.get(input)
+                if (nestedValue != null) {
+                    """${it.name}:${serialize(nestedValue)}"""
+                } else {
+                    null
+                }
+            }.filterNotNull().joinToString(", ", "{", " }")
+        }
     }
 }
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
@@ -19,17 +19,17 @@ package com.netflix.graphql.dgs.client.codegen
 import graphql.schema.Coercing
 import org.springframework.util.ReflectionUtils
 import java.lang.reflect.Field
+import java.time.*
 import java.util.*
 
 class GraphQLQueryRequest(
     private val query: GraphQLQuery,
     private val projection: BaseProjectionNode?,
-    private val scalars: Map<Class<*>, Coercing<*, *>>?
+    scalars: Map<Class<*>, Coercing<*, *>>?
 ) {
 
     constructor(query: GraphQLQuery) : this(query, null, null)
     constructor(query: GraphQLQuery, projection: BaseProjectionNode?) : this(query, projection, null)
-
     private val inputValueSerializer = InputValueSerializer(scalars ?: emptyMap())
 
     fun serialize(): String {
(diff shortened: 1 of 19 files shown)


paulbakker:
As @srinivasankavitha pointed out in an offline conversation, this doesn't account for scalars nested in complex types.
Example:

```
type Query {
    movies(filter: MovieFilter)
}
input MovieFilter {
    dateRange: DateRange
}
scalar DateRange
```

Will follow up with adding support for that as well.

amitgupta1202:
@paulbakker I like the overall Idea, I believe its in same direction as https://github.com/amitgupta1202/dgs-framework/commit/492bf7e9c4cd6a94caa91cf757f58dea4d9d4b75 . 

Potential area that needs to be looked when considering above proposal. 

1. How it ties with ExtendedScalar implementation as it already provides implementation for ScalarTypes. 
2. Coercing implements both serialiazation and deserialization, and as a client use case deserialization won't be used at all. May be interface Coercing can be compose of 2 different interfaces?

richardcresswell:
I'm excited about this PR, we've implemented some complex workarounds that this would resolve.  +1 to needing it to work for nested types as well.

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/GraphQLQueryRequestTest.kt
```diff
@@ -131,3 +151,31 @@ class MovieProjection : BaseProjectionNode() {
...
+
+    @Throws(CoercingSerializeException::class)
+    override fun serialize(dataFetcherResult: Any): String {
+        val range: DateRange = dataFetcherResult as DateRange
+        return range.from.format(formatter) + "-" + range.to.format(formatter)
+    }
+
+    @Throws(CoercingParseValueException::class)
+    override fun parseValue(input: Any): DateRange {
+        val split = (input as String).split("-").toTypedArray()
```
amitgupta1202, line 166: In current context it seems only serialize() method is in use. It seems additional overheads to define deserialization methods as its not in use.

amitgupta1202: Is it worth to compose Coercing out of 2 different interfaces (serialization and deserialization), and force client to implement only the serialization part?

paulbakker: The `Coercing` type is what all scalars are implemented with, so the goal is to just reuse that code "as is". It's unlikely a user would implement such a scalar _just_ for the client use case I think.

paulbakker:
I've added support for nested types now, and replaced most of the existing serialization logic.
All the existing tests pass.

srinivasankavitha:
These changes are looking really good! 🎉

graphql-dgs-client/src/test/java/com/netflix/graphql/dgs/client/MovieInput.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+
+public class MovieInput {
+    private int movieId;
+    private String title;
+    private Genre genre;
+    private Director director;
+    private List<Actor> actor;
+    private DateRange releaseWindow;
+
```
srinivasankavitha, line 30: I'd recommend also adding a List<Integer> or some list of non-string but primitive type to the test case.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
```diff
@@ -63,10 +57,40 @@ class GraphQLQueryRequest(private val query: GraphQLQuery, private val projectio
...
+        } else if (type.isPrimitive || type == Integer::class.java || type == Long::class.java || type == Double::class.java || type == Float::class.java || type == Boolean::class.java || type == Short::class.java || type == Byte::class.java || type.isEnum) {
+            input.toString()
+        } else if (type == String::class.java) {
+            """"$input""""
+        } else if (input is List<*>) {
+            """[${input.filterNotNull().joinToString(", ") { listItem -> serialize(listItem) ?: "" }}]"""
+        } else {
+            input.javaClass.declaredFields.map {
+                it.isAccessible = true
+                val nestedValue = it.get(input)
```
srinivasankavitha, line 87: Does this null check work if the field is an unboxed type, e.g. int?

paulbakker: primitives can't be null, they always have a default value instead.

srinivasankavitha: What I mean is do you need to do the null check only if the type is not primitive? because it is an incompatible check, no?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
```diff
@@ -63,10 +57,41 @@ class GraphQLQueryRequest(private val query: GraphQLQuery, private val projectio
...
+        return if (scalars.contains(type)) {
+            """"${scalars[type]!!.serialize(input)}""""
+        } else if (type.isPrimitive || type.isAssignableFrom(java.lang.Integer::class.java) || type.isAssignableFrom(java.lang.Long::class.java) || type.isAssignableFrom(java.lang.Double::class.java) || type.isAssignableFrom(java.lang.Float::class.java) || type.isAssignableFrom(java.lang.Boolean::class.java) || type.isAssignableFrom(java.lang.Short::class.java) || type.isAssignableFrom(java.lang.Byte::class.java) || type.isEnum) {
+            input.toString()
+        } else if (type == String::class.java) {
+            """"$input""""
+        } else if (input is List<*>) {
+            """[${input.filterNotNull().joinToString(", ") { listItem -> serialize(listItem) ?: "" }}]"""
+        } else {
+            input.javaClass.declaredFields.filter { !it.type::class.isCompanion }.map {
```
richardcresswell, line 86: When I tried this I was getting null values and narrowed it down to here -- unfortunately declaredFields does not include inherited fields, so this breaks extended objects.  A quick trip to stackoverflow shows some solutions:

https://stackoverflow.com/questions/1042798/retrieving-the-inherited-attribute-names-values-using-java-reflection

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
```diff
@@ -63,10 +57,41 @@ class GraphQLQueryRequest(private val query: GraphQLQuery, private val projectio
...
+        }
+
+        val type = input::class.java
+
+        return if (scalars.contains(type)) {
+            """"${scalars[type]!!.serialize(input)}""""
+        } else if (type.isPrimitive || type.isAssignableFrom(java.lang.Integer::class.java) || type.isAssignableFrom(java.lang.Long::class.java) || type.isAssignableFrom(java.lang.Double::class.java) || type.isAssignableFrom(java.lang.Float::class.java) || type.isAssignableFrom(java.lang.Boolean::class.java) || type.isAssignableFrom(java.lang.Short::class.java) || type.isAssignableFrom(java.lang.Byte::class.java) || type.isEnum) {
+            input.toString()
+        } else if (type == String::class.java) {
+            """"$input""""
```
richardcresswell, line 82: I believe backslashes and double quotes need to be escaped here.

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/GraphQLQueryRequestTest.kt
amitgupta1202: :) We have use case that we only consume graphQL API to enrich the data set. But its no deal breaker as i can leave it unimplemented.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
```diff
@@ -16,9 +16,12 @@
 
 package com.netflix.graphql.dgs.client.codegen
 
-class GraphQLQueryRequest(private val query: GraphQLQuery, private val projection: BaseProjectionNode?) {
+import graphql.schema.Coercing
 
-    constructor(query: GraphQLQuery) : this(query, null)
+class GraphQLQueryRequest(private val query: GraphQLQuery, private val projection: BaseProjectionNode?, private val scalars: Map<Class<*>, Coercing<*, *>>?) {
+
+    constructor(query: GraphQLQuery) : this(query, null, null)
```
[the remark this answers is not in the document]
paulbakker: Default parameters aren't made available for Java users, you only get the constructor with all parameters. Hence the overloads.

graphql-dgs-client/src/test/java/com/netflix/graphql/dgs/client/MovieInput.java
paulbakker: done

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
paulbakker: I moved the null check after updating the implementation, so this isn't relevant anymore

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
paulbakker: Good point, I've added support for extended objects now as well.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
paulbakker: It's a multiline string, so I don't think any escaping is necessary. But maybe I'm not understanding what you're pointing out?

richardcresswell: The escapes need to make it out to the serialized request, otherwise once it becomes GraphQL it will look like this:
```
inputParam: "my"unescaped"string\"
```

paulbakker: Added a test: https://github.com/Netflix/dgs-framework/pull/333/commits/a762cb7656376cc4be9250cdae5e6e281fe40e8d

I think that's what you're describing?

richardcresswell: Yeah, that's right - if that works then we're all good!  Thanks Paul.

richardcresswell: So actually, this test case does not show the solution, it shows the problem.  Those need to be double escaped so that the escapes are added for the query.  Try running your test case example through the DgsQueryExecutor and it will get confused because of stray quotes and backslashes in the middle of the String.

PR #344, opened 2021-05-15
Handle non null scalar types as input

briandilley:
If you have a custom scalar type and you take it as an input argument in non-null form you get an exception, this should fix it.

Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Handle non null scalar types as input


paulbakker:
I've added a test that verifies the behavior with a null value for a date/time scalar. It now works as expected, so we can close this PR.

PR #389, opened 2021-06-08
Support returning Mono/Flux from data fetchers

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

`graphql-java` doesn't support `Mono` or `Flux` return types, only `CompletableFuture`.
In a WebFlux environment, it may be desirable to return Mono/Flux from data fetchers, however.

This PR adds a convenience that transforms a Mono/Flux to a `CompletableFuture` automatically so that `graphql-java` can process its execution. 
`graphql-java` is already executing async, so this doesn't add any extra blocking.

Do note that a Mono/Flux *must* be completed, otherwise the request will never complete. This means it doesn't support a streaming style of response (but you should use subscriptions in that case!).

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -225,7 +225,7 @@ class DgsSchemaProvider(
                         if (type is InterfaceTypeDefinition) {
                             val implementationsOf = typeDefinitionRegistry.getImplementationsOf(type)
                             implementationsOf.forEach { implType ->
-                                val dataFetcher = createBasicDataFetcher(method, dgsComponent)
+                                val dataFetcher = createBasicDataFetcher(method, dgsComponent, parentType == "Subscription")
                                 codeRegistryBuilder.dataFetcher(
                                     FieldCoordinates.coordinates(implType.name, field),
                                     dataFetcher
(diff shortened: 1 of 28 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -276,8 +278,11 @@ class DgsSchemaProvider(
 
     private fun createBasicDataFetcher(method: Method, dgsComponent: Any): DataFetcher<Any?> {
         return DataFetcher<Any?> { environment ->
-            val result = invokeDataFetcher(method, dgsComponent, DgsDataFetchingEnvironment(environment))
-            result
+            when (val result = invokeDataFetcher(method, dgsComponent, DgsDataFetchingEnvironment(environment))) {
```
driverpt, line 281: Won't this create a hard dependency on `reactor` ?

srinivasankavitha (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
paulbakker: It would, which I also wasn't happy with.

I've refactored it behind a `DataFetcherResultProcessor` abstraction now, which is implemented in the reactive module: https://github.com/Netflix/dgs-framework/pull/389/commits/395c0a46c19354afa557ea9345b36fff3f4caec3

paulbakker:
Also added Mono/Flux datafetchers to the webflux example (in this repo). Will update the external examples once this is released.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -532,3 +550,8 @@ class DgsSchemaProvider(
         return schemas + metaInfSchemas
     }
 }
+
+interface DataFetcherResultProcessor {
```
berngp, line 554: Java/Kotlin Doc? probably its own file?

berngp (approved):

PR #395, opened 2021-06-09
Explicitly nullable input arguments

paulbakker:
If an input field in a Query builder is explicitly set to null, it should be included in the "inputs"
This PR comes together with https://github.com/Netflix/dgs-codegen/pull/162 on codegen.


srinivasankavitha (approved):

cohem1:
@paulbakker  I wonder if I could distinguish a key of input that value actually is null and the key itself is unset in this case

PR #409, opened 2021-06-16
Support disabling introspection

sbilello:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

The goal of this PR is to disable introspection via an `application.yml` config parameter.

Discussion:
https://github.com/Netflix/dgs-framework/discussions/361

Notes
----

I removed a not necessary instantiation over here https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L103

Thanks
@berngp @paulbakker @srinivasankavitha

### DIFF
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -146,7 +146,7 @@ open class DgsAutoConfiguration(
     @Bean
     @ConditionalOnMissingBean
     open fun schema(dgsSchemaProvider: DgsSchemaProvider): GraphQLSchema {
-        return dgsSchemaProvider.schema()
+        return dgsSchemaProvider.schema(null, configProps.introspection)
     }
 
     @Bean
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -31,9 +31,13 @@ import graphql.execution.instrumentation.Instrumentation
 import graphql.schema.GraphQLCodeRegistry
 import graphql.schema.GraphQLSchema
 import graphql.schema.idl.TypeDefinitionRegistry
+import graphql.schema.visibility.DefaultGraphqlFieldVisibility.DEFAULT_FIELD_VISIBILITY
+import graphql.schema.visibility.GraphqlFieldVisibility
+import graphql.schema.visibility.NoIntrospectionGraphqlFieldVisibility.NO_INTROSPECTION_FIELD_VISIBILITY
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.boot.autoconfigure.ImportAutoConfiguration
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.autoconfigure.jackson.JacksonAutoConfiguration
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.ApplicationContext
@@ -53,6 +57,10 @@ import java.util.*
 open class DgsAutoConfiguration(
     private val configProps: DgsConfigurationProperties
 ) {
+
+    companion object {
+        const val AUTO_CONF_PREFIX = "dgs.graphql"
+    }
 
     @Bean
     open fun dgsQueryExecutor(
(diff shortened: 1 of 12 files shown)


graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsConfigurationProperties.kt
```diff
@@ -29,5 +29,7 @@ import org.springframework.boot.context.properties.bind.DefaultValue
 @Suppress("ConfigurationProperties")
 data class DgsConfigurationProperties(
     /** Location of the GraphQL schema files. */
-    @DefaultValue(DEFAULT_SCHEMA_LOCATION) val schemaLocations: List<String>
+    @DefaultValue(DEFAULT_SCHEMA_LOCATION) val schemaLocations: List<String>,
+    /** Enable Introspection */
+    @DefaultValue("true") val introspection: Boolean
```
berngp, line 34: I personally prefer a more explicit name, such as `introspection.enabled` as well as docs that express the caveats. e.g.
```
/**
 * Setting this value to `false` will prevent Introspection queries from being performed.
 * Note that this is against GraphQL Specification, that said,  some production systems want this lock down in place.
 */
```
---

If we treat properties as APIs, which imo we should, this looks like a _good enough_ first approach. At the same time it got me thinking. Are there any other options that we would like to configure for _field visibility_? What if someone wants to use `BlockedFields` instead of `NoIntrospectionGraphqlFieldVisibility`?

Currently the `GraphQLCodeRegistry` accepts only one kind of _field visibility_, developers can implement their own and out of the box we have:

- DefaultGraphqlFieldVisibility
- NoIntrospectionGraphqlFieldVisibility
- BlockedFields

Then _visibility_ could be treated as a _bean_ and be a `@ConditionalOnMissingBean`. By _default_ the `DefaultGraphqlFieldVisibility` bean will be provided by it can be turned off by enabling a different _kind_ such as the `NoIntrospectionGraphqlFieldVisibility`, or provide your own implementation of a `GraphqlFieldVisibility` and express it as a bean. The latter will help if developers are interested in using `BlockedFields` or express their own logic.

@paulbakker @srinivasankavitha wdyt?

berngp (commented):
As a first approach there are some minor comments. That said I question a bit the implementation if there is any appetite to further configure _field visibility_.  Are there any other options that we would like to configure for field visibility? What if someone wants to use `BlockedFields` instead of `NoIntrospectionGraphqlFieldVisibility`?

_I expressed my thoughts on this further down the review, specifically while looking at the changes to the `DgsConfigurationProperties`_.

sbilello:
@berngp  
Please take a look at the changes.
I removed the change on the `additional-spring-configuration-metadata.json` however I was following the pattern over here: https://github.com/Netflix/dgs-framework/blob/0108a2fc97c920ae08f3981a933a7596b5e01b8d/graphql-dgs-spring-webmvc-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L8

If there is any other change please let me know. 

I added a small addition to the documentation comment for the `introspection-enabled` property.

Thanks

andimarek:
@sbilello I am curious: can you explain a bit why you want to disable introspection? What is the use case? 

For example if you want to protect your API you would protect with authz (including Introspection), I assume.
Thanks a lot

sbilello:
@andimarek https://www.apollographql.com/docs/apollo-server/api/apollo-server/#introspection
There was also a security best practice talk provided by the `Apollographql` folks recommending to disable the introspection in production for security purposes.

berngp:
@sbilello I was unable to find the actual recommendation in the link you provided, that said they do have [this blog post](https://www.apollographql.com/blog/graphql/security/why-you-should-disable-graphql-introspection-in-production/).

From an implementation perspective I'm leaning more towards having a `GraphqlFieldVisibility` as a _bean_ where DGS provides one by default, the `DefaultGraphqlFieldVisibility`, and developers can override it by providing their own implementation.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -145,8 +153,23 @@ open class DgsAutoConfiguration(
...
+        prefix = AUTO_CONF_PREFIX,
+        name = ["introspection-enabled"], havingValue = "false", matchIfMissing = false
+    )
+    open fun noIntrospectionFieldVisibility(): GraphqlFieldVisibility {
+        return NO_INTROSPECTION_FIELD_VISIBILITY
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    open fun defaultFieldVisibility(): GraphqlFieldVisibility {
```
sbilello, line 171: @berngp this is the default Bean

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -145,8 +153,23 @@ open class DgsAutoConfiguration(
...
+    open fun schema(dgsSchemaProvider: DgsSchemaProvider, fieldVisibility: GraphqlFieldVisibility): GraphQLSchema {
+        return dgsSchemaProvider.schema(null, fieldVisibility)
+    }
+
+    @Bean
+    @ConditionalOnProperty(
+        prefix = AUTO_CONF_PREFIX,
+        name = ["introspection-enabled"], havingValue = "false", matchIfMissing = false
+    )
+    open fun noIntrospectionFieldVisibility(): GraphqlFieldVisibility {
```
sbilello, line 165: @berngp this is the bean enabled in case we want to disable the introspection

sbilello:
Screen Shot 2021-06-16 at 6 03 17 PM
@berngp I changed the implementation following the bean approach

berngp (approved):
Thanks for the contribution @sbilello !

StringKe:
How to control introspection enable/disable at runtime?

Developers need to carry a special token/header to use the introspection feature.

congotej:
Hi @StringKe! Thank you for your question. Currently, our framework does not support dynamically changing configuration settings at runtime.

PR #415, opened 2021-06-17
Query field arguments

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

This further moves serialization of projections to the client (away from codegen). The serialization is now prepared for projection fields that have arguments.

As an example, the following schema as input arguments on projection fields defined (see the `reviews` and `username` fields.

```
type Show {
    id: Int
    title: String
    releaseYear: Int
    reviews(minScore:Int, since: DateTime): [Review]
    artwork: [Image]
}

type Review {
    username(toUpperCase: Boolean): String
    starScore: Int
    submittedDate: DateTime
}
```

Codegen now generates a second method for each field, which takes input arguments. 

```java
  GraphQLQueryRequest graphQLQueryRequest = new GraphQLQueryRequest(ShowsGraphQLQuery.newRequest().titleFilter("").build(),
                new ShowsProjectionRoot()
                        .title()
                        .reviews(4, OffsetDateTime.now())
                        .username(true)
                        .starScore(), scalars);
```

The code in this PR serializes these fields correctly, optionally using scalars configured in the `GraphQLQueryRequest`.


paulbakker:
Codegen change for this: https://github.com/Netflix/dgs-codegen/pull/172

berngp (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
```diff
@@ -56,76 +53,12 @@ class GraphQLQueryRequest(
         }
```
srinivasankavitha, line 54: Why is this check below needed?

srinivasankavitha (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/GraphQLQueryRequest.kt
paulbakker: Only a subprojection has a `root()`.

Issue #375, opened 2021-05-31
feature: Integration with WebFlux ReactiveSecurityContextHolder


driverpt:
## Describe the Feature Request
Currently we are trying to access `ReactiveSecurityContextHolder` to fetch the Principal, since we are dependant on Roles to show the proper Data.

e.g.:
```
  @DgsQuery(field = "stuffByFilter")
  public CompletionStage<List<Stuff>> leaves(@InputArgument("filter") final StuffFilter filter) {
    return ReactiveSecurityContextHolder.getContext()
      .map(SecurityContext::getAuthentication)
      .map(Authentication::getPrincipal)
      .map(Object::toString)
      .doOnNext(log::info)
      .flatMapMany(principal -> randomStuff())
      .collectList()
      .toFuture();
  }
```
The above code returns empty list as no logging is generated and no breakpoints are hit.

## Describe Preferred Solution
Ideally we could get the current Principal available in the WebFlux Call chain via `ReactiveSecurityContextHolder`

driverpt:
Please let me know if i'm doing something wrong

hantsy:
Make sure you have set the authentication in the Security Context(by Spring Secuirty filter or manually) before fetching it.

AlexRiedler:
I have tested with having a seperate controller, and a DgsData with PreAuthorize annotation, it appears that DGS is losing the ReactiveSecurityContext somehow during the transition. Not sure exactly where yet, but this is probably something much needed for a lot of people.

AlexRiedler:
Okay, after messing around inside dgs-framework, I think I have a path forward but would love input from the maintainers.

I modified DataFetcherResultProcessor:

```kt
 interface DataFetcherResultProcessor {
     fun supportsType(originalResult: Any): Boolean
     fun process(originalResult: Any, dfe: DataFetchingEnvironment): Any
 }
```

this is a relatively trivial change, allowing access to the dfe (or potential we should allow access to DgsDataFetchingEnvironment instead) inside of Mono/FluxDataFetcherResultProcessors.

I then had to implement my own custom Mono/Flux implementation converters since spring-security is not included in these modules.

```kt
class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
    override fun supportsType(originalResult: Any): Boolean {
        return originalResult is Mono<*>
    }

    override fun process(originalResult: Any, dfe: DataFetchingEnvironment): Any {
        if (originalResult is Mono<*>) {
[...]

PR #427, opened 2021-06-21
DataFetchResultProcessor: add DFE as an argument

AlexRiedler:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

This will allow custom contexts to inject data into the result by type
as necessary based on the DataFetchingEnvironment.

For instance, in WebFlux apps this will allow a way to inject Mono's
ContextView object into the result processor via custom contexts.
Allowing the passing of SecurityContext through the stack, etc...

Not sure if this is the best approach, or we are harming the interface
by adding the DFE. Would there be other scenarios where adding the
DFE would also be useful? I was thinking maybe something like using
schema annotations to filter values but there might be other ways to
do that?

Issue https://github.com/Netflix/dgs-framework/issues/375

Alternatives considered
----

May have overlooked something here, this feels like the easiest path; although it might not be the best.

This does not fix the entire issue attached, supporting the Mono's context through the stack to Mono/Flux return types for reactive applications. However, this is one step towards adding a ReactiveDgsContext which might be a solution for the mentioned ticket.

Example usage
----

```kt
@Component
 class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
     override fun supportsType(originalResult: Any): Boolean {
         return originalResult is Mono<*>
     }

     override fun process(originalResult: Any, dfe: DgsDataFetchingEnvironment): Any {
         if (originalResult is Mono<*>) {
             val webContext = (dfe.getDgsContext().customContext as WebContext)
             return originalResult.contextWrite(webContext.monoContext).toFuture()
         } else {
             throw IllegalArgumentException("Instance passed to ${this::class.qualifiedName} was not a Mono<*>. It was a ${originalResult::class.qualifiedName} instead")
         }
     }
 }

 @Component
 class FluxDataFetcherResultProcessor : DataFetcherResultProcessor {
     override fun supportsType(originalResult: Any): Boolean {
         return originalResult is Flux<*>
     }

     override fun process(originalResult: Any, dfe: DgsDataFetchingEnvironment): Any {
         if (originalResult is Flux<*>) {
             val webContext = (dfe.getDgsContext().customContext as WebContext)
             return originalResult.contextWrite(webContext.monoContext).collectList().toFuture()
         } else {
             throw IllegalArgumentException("Instance passed to ${this::class.qualifiedName} was not a Flux<*>. It was a ${originalResult::class.qualifiedName} instead")
         }
     }
 }

 @Component
 class MonoContextBuilder : DgsReactiveCustomContextBuilderWithRequest<WebContext> {
     override fun build(
         extensions: Map<String, Any>?,
         headers: HttpHeaders?,
         serverRequest: ServerRequest?
     ): Mono<WebContext> {
         return Mono.deferContextual { context -> Mono.just(WebContext(context)) }
     }
 }
 
 class WebContext(val monoContext: ContextView) {
 }

```

### DIFF
--- a/graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/ReactiveDataFetcherResultProcessor.kt
+++ b/graphql-dgs-reactive/src/main/kotlin/com/netflix/graphql/dgs/reactive/internal/ReactiveDataFetcherResultProcessor.kt
@@ -16,6 +16,7 @@
 
 package com.netflix.graphql.dgs.reactive.internal
 
+import com.netflix.graphql.dgs.DgsDataFetchingEnvironment
 import com.netflix.graphql.dgs.internal.DataFetcherResultProcessor
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
@@ -26,7 +27,7 @@ class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
         return originalResult is Mono<*>
     }
 
-    override fun process(originalResult: Any): Any {
+    override fun process(originalResult: Any, dfe: DgsDataFetchingEnvironment): Any {
         if (originalResult is Mono<*>) {
             return originalResult.toFuture()
         } else {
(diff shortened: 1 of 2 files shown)


paulbakker:
This makes a lot of sense, thanks for the PR @AlexRiedler!

PR #428, opened 2021-06-22
Mark RestController attributes as open val to allow CGLIB proxying

lucatk:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Following up #275, we discovered another small requirement for change in order to be able to use this library in combination with Spring AOP's CGLIB proxying.
As I know, it is not the purpose of these properties to be open, however we see CGLIB proxying as a dealbreaker in some large projects, and therefore would be very grateful if these changes could be incorporated to finally allow coexistence of DGS & AOP, seeing as this could benefit many others aswell, as this is a difficult cause of issue to explore.


berngp:
Per @paulbakker 's comments [here](https://github.com/Netflix/dgs-framework/pull/275#issuecomment-821585665).

PR #439, opened 2021-06-25
Client Codegen migration to graphql-dgs-codegen-client-core

berngp:
This commit is part of an effort to migrate the _DGS Client Components_ that should belong to the  _DGS Codegen_ project.
This will allow us to evolve the both projects independently. In other words, we will not need to release a new version of the DGS Framework if we need changes to the _DGS Codegen_ components.

This PR is coupled with [Codegen PR 174](https://github.com/Netflix/dgs-codegen/pull/174).


srinivasankavitha:
This change will require users to explicitly remove the dependency they have on the dgs client module (if they are not using the client directly, and only using it for the query construction), correct? 
Although it won't hurt to leave it in I suppose?

srinivasankavitha (approved):

berngp:
> This change will require users to explicitly remove the dependency they have on the dgs client module (if they are not using the client directly, and only using it for the query construction), correct?

@srinivasankavitha, the idea is to move the _codegen_ specific code to just reside in [graphql-dgs-codegen-client-core](https://github.com/Netflix/dgs-codegen/tree/master/graphql-dgs-codegen-client-core). If a developer wants to leverage the client they will still need to import the `com.netflix.graphql.dgs:graphql-dgs-client` dependency.

Note that the `graphql-dgs-codegen-client-core` doesn't bring transitively the `com.netflix.graphql.dgs:graphql-dgs-client` dependency. This was done to avoid the plugin _upgrade_, under the hood, a pinned version of the `graphql-dgs-client` that the user might have defined.

PR #451, opened 2021-06-29
Enable Projection Types to define the Schema Type explicitly

berngp:
This provides a mechanism to explicitly define a _GraphQL Schema Type_
that will be used by a _projection_. This is useful when the projection
is not expressing the _GraphQL Schema Type_ via its Java class name.

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/ProjectionSerializer.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/ProjectionSerializer.kt
@@ -19,16 +19,24 @@ package com.netflix.graphql.dgs.client.codegen
 import java.util.*
 
 class ProjectionSerializer(private val inputValueSerializer: InputValueSerializer) {
+
     fun serialize(projection: BaseProjectionNode, isFragment: Boolean = false): String {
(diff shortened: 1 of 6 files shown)


srinivasankavitha (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/codegen/ProjectionSerializer.kt
```diff
@@ -19,16 +19,24 @@ package com.netflix.graphql.dgs.client.codegen
...
 class ProjectionSerializer(private val inputValueSerializer: InputValueSerializer) {
+
     fun serialize(projection: BaseProjectionNode, isFragment: Boolean = false): String {
         if (projection.fields.isEmpty() && projection.fragments.isEmpty()) {
             return ""
         }
 
         val prefix = if (isFragment) {
-            "... on ${projection::class.java.name.substringAfterLast(".").substringAfterLast("_").substringBefore("Projection")} { "
+            val schemaType = projection.schemaType.orElse(
```
paulbakker, line 29: The name `schemaType` is a bit too generic. I don't immediately have a better alternative though, let's see if we can come up with something while talking this through, otherwise we keep this name.

Issue #413, opened 2021-06-17
feature: Add Kotlin Coroutines support


hantsy:
Add Kotlin coroutines support for  WebFlux/Kotlin project(replace the reactive `DataFetcher` with Kotlin coroutines),  and the following cases in a Kotlin project(both MVC and WebFlux).

* DataLoader
* Subscription Handler

AlexRiedler:
I did some investigation into this, part of this work would require us to be able to supply a coroutine context as well since WebFluxSecurity context will probably need to be injected. GraphQL-Kickstart has this implemented https://github.com/graphql-java-kickstart/graphql-java-tools/blob/b6cfd647a6ded1b7f38a04a7fd2d9ea3b0287a7b/src/main/kotlin/graphql/kickstart/tools/resolver/MethodFieldResolver.kt#L194 but we might want to abstract the field resolvers in the dgs framework code to be more re-usable objects. Just my 2-cents. (I can also provide an example project if need be on how graphql-kickstart did this).

paulbakker:
Basic support added in #472. Please try it out and review!

paulbakker:
Merged in #472

hantsy:
Is there a *Kotlin Coroutines* based `DataLoader` ?

PR #472, opened 2021-07-07
Add basic support for coroutines in datafetchers

paulbakker:
A datafetcher defined as a coroutine is async, without using CompletableFuture.
This does *not* change the way dataloaders work, because the dataloader API requires CompletableFuture.

Issue #413.

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
@@ -18,13 +18,14 @@ package com.netflix.graphql.dgs.context
 
 import com.netflix.graphql.dgs.internal.DgsRequestData
 import graphql.schema.DataFetchingEnvironment
+import kotlinx.coroutines.CoroutineScope
 import org.dataloader.BatchLoaderEnvironment
 
 /**
  * Context class that is created per request, and is added to both DataFetchingEnvironment and BatchLoaderEnvironment.
  * Custom data can be added by providing a [DgsCustomContextBuilder].
  */
-open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?) {
+open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?, val coroutineScope: CoroutineScope? = null) {
 
     companion object {
         @JvmStatic
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
@@ -18,14 +18,13 @@ package com.netflix.graphql.dgs.context
 
 import com.netflix.graphql.dgs.internal.DgsRequestData
 import graphql.schema.DataFetchingEnvironment
-import kotlinx.coroutines.CoroutineScope
 import org.dataloader.BatchLoaderEnvironment
 
 /**
  * Context class that is created per request, and is added to both DataFetchingEnvironment and BatchLoaderEnvironment.
  * Custom data can be added by providing a [DgsCustomContextBuilder].
  */
-open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?, val coroutineScope: CoroutineScope? = null) {
+open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?) {
 
     companion object {
         @JvmStatic
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -431,7 +431,7 @@ class DgsSchemaProvider(
             }
         }
 
-        return if(method.kotlinFunction?.isSuspend == true) {
+        return if (method.kotlinFunction?.isSuspend == true) {
             val dgsContext = environment.getContext<DgsContext>()
             val launch = dgsContext.coroutineScope?.async {
                 return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -39,6 +39,8 @@ import graphql.schema.idl.TypeDefinitionRegistry
 import graphql.schema.idl.TypeRuntimeWiring
 import graphql.schema.visibility.DefaultGraphqlFieldVisibility
 import graphql.schema.visibility.GraphqlFieldVisibility
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.future.asCompletableFuture
 import org.slf4j.LoggerFactory
@@ -432,8 +434,8 @@ class DgsSchemaProvider(
         }
 
         return if (method.kotlinFunction?.isSuspend == true) {
-            val dgsContext = environment.getContext<DgsContext>()
-            val launch = dgsContext.coroutineScope?.async {
+
+            val launch = CoroutineScope(Dispatchers.Unconfined).async {
                 return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
             }
 
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -39,6 +39,8 @@ import graphql.schema.idl.TypeDefinitionRegistry
 import graphql.schema.idl.TypeRuntimeWiring
 import graphql.schema.visibility.DefaultGraphqlFieldVisibility
 import graphql.schema.visibility.GraphqlFieldVisibility
+import kotlinx.coroutines.async
+import kotlinx.coroutines.future.asCompletableFuture
 import org.slf4j.LoggerFactory
 import org.springframework.aop.support.AopUtils
 import org.springframework.context.ApplicationContext
@@ -59,6 +61,9 @@ import java.nio.charset.StandardCharsets
 import java.util.*
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
+import kotlin.coroutines.Continuation
+import kotlin.reflect.full.callSuspend
+import kotlin.reflect.jvm.kotlinFunction
 
 /**
  * Main framework class that scans for components and configures a runtime executable schema.
@@ -303,7 +308,7 @@ class DgsSchemaProvider(
     private fun invokeDataFetcher(method: Method, dgsComponent: Any, environment: DataFetchingEnvironment): Any? {
         val args = mutableListOf<Any?>()
         val parameterNames = defaultParameterNameDiscoverer.getParameterNames(method) ?: emptyArray()
-        method.parameters.forEachIndexed { idx, parameter ->
+        method.parameters.filter { it.type != Continuation::class.java }.forEachIndexed { idx, parameter ->
 
             when {
                 parameter.isAnnotationPresent(InputArgument::class.java) -> {
@@ -426,7 +431,16 @@ class DgsSchemaProvider(
             }
         }
 
-        return ReflectionUtils.invokeMethod(method, dgsComponent, *args.toTypedArray())
+        return if(method.kotlinFunction?.isSuspend == true) {
+            val dgsContext = environment.getContext<DgsContext>()
+            val launch = dgsContext.coroutineScope?.async {
+                return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
+            }
+
+            launch?.asCompletableFuture()
+        } else {
+            ReflectionUtils.invokeMethod(method, dgsComponent, *args.toTypedArray())
+        }
     }
 
     private fun getValueAsOptional(value: Any?, parameter: Parameter) =
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/CoroutineDataFetcherTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/CoroutineDataFetcherTest.kt
@@ -23,16 +23,16 @@ import graphql.GraphQL
 import io.mockk.every
 import io.mockk.impl.annotations.MockK
 import io.mockk.junit5.MockKExtension
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.data.Percentage
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.springframework.context.ApplicationContext
 import java.util.*
+import java.util.concurrent.Executors
 import kotlin.system.measureTimeMillis
 
 @ExtendWith(MockKExtension::class)
@@ -40,23 +40,35 @@ class CoroutineDataFetcherTest {
     @MockK
     lateinit var applicationContextMock: ApplicationContext
 
+    val executor = Executors.newFixedThreadPool(8)
+
     @Test
     fun `Suspend functions should be supported as datafetchers`() {
         val fetcher = object : Any() {
             @DgsQuery
             suspend fun concurrent(@InputArgument from: Int, to: Int): Int = coroutineScope {
                 var sum = 0
+                withContext(executor.asCoroutineDispatcher()) {
+                    println("before $from")
+                    repeat(from.rangeTo(to).count()) {
+                        sum++
 
-                repeat(from.rangeTo(to).count()) {
-                    sum++
-                    delay(50)
+                        //Forcing a blocking call to demonstrate running with a thread pool
+                        Thread.sleep(50)
+                    }
+                    println("after $from")
+
+                    sum
                 }
-
-                sum
             }
         }
 
-        every { applicationContextMock.getBeansWithAnnotation(DgsComponent::class.java) } returns mapOf(Pair("concurrentFetcher", fetcher))
+        every { applicationContextMock.getBeansWithAnnotation(DgsComponent::class.java) } returns mapOf(
+            Pair(
+                "concurrentFetcher",
+                fetcher
+            )
+        )
         every { applicationContextMock.getBeansWithAnnotation(DgsScalar::class.java) } returns emptyMap()
 
         val provider = DgsSchemaProvider(applicationContextMock, Optional.empty(), Optional.empty(), Optional.empty())
@@ -73,7 +85,6 @@ class CoroutineDataFetcherTest {
         val context = DgsContext(
             null,
             null,
-            CoroutineScope(Dispatchers.Default)
         )
 
         val concurrentTime = measureTimeMillis {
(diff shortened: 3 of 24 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -303,7 +308,7 @@ class DgsSchemaProvider(
     private fun invokeDataFetcher(method: Method, dgsComponent: Any, environment: DataFetchingEnvironment): Any? {
         val args = mutableListOf<Any?>()
         val parameterNames = defaultParameterNameDiscoverer.getParameterNames(method) ?: emptyArray()
```
srinivasankavitha, line 310: Why is this needed?

srinivasankavitha (approved):

berngp (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
```diff
@@ -18,13 +18,14 @@ package com.netflix.graphql.dgs.context
...
 import graphql.schema.DataFetchingEnvironment
+import kotlinx.coroutines.CoroutineScope
 import org.dataloader.BatchLoaderEnvironment
 
 /**
  * Context class that is created per request, and is added to both DataFetchingEnvironment and BatchLoaderEnvironment.
  * Custom data can be added by providing a [DgsCustomContextBuilder].
  */
-open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?) {
+open class DgsContext(val customContext: Any? = null, val requestData: DgsRequestData?, val coroutineScope: CoroutineScope? = null) {
```
berngp, line 28: nit: Should we create yet another level inside the `DgsContext` that expresses library specific _contexts_. In this case we are mixing at the same level the _User Customizable Information_, _Request Data_, and the _Kotlin Coroutine Scope_.

paulbakker: Yeah I wasn't sure about that. I don't have a good alternative in mind though.

Do you have concrete ideas how to improve it?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -426,7 +431,16 @@ class DgsSchemaProvider(
             }
         }
 
-        return ReflectionUtils.invokeMethod(method, dgsComponent, *args.toTypedArray())
+        return if (method.kotlinFunction?.isSuspend == true) {
```
kilink, line 434: I think there's some helpers from spring-core we could use;

Here's one for checking if a method is suspending:

https://github.com/spring-projects/spring-framework/blob/b595dc1dfad9db534ca7b9e8f46bb9926b88ab5a/spring-core/src/main/java/org/springframework/core/KotlinDetector.java#L82

Here is one for invoking a suspending function:

https://github.com/spring-projects/spring-framework/blob/3dba3691a7cee4299241e1f12d1680788ea77731/spring-core/kotlin-coroutines/src/main/kotlin/org/springframework/core/CoroutinesUtils.kt#L60

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
kilink: If we used Using CoroutinesUtils.invokeSuspendingFunction, we could kick the can / not worry about the scope or dispatcher, as it just defaults to running on Dispatcher.Unconfined.

Separately I think Dispatchers.Default might be problematic here anyway, as it by default limits the number of threads to the number of CPUs.

berngp: I like @kilink's idea, that said, if we need to keep these kind of context I would hope that eventually we can leverage the API in the latest graphql-java snapshot that has a [DataFetchingEnvironment.getGraphQLContext](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/DataFetchingEnvironment.java#L101) where the [GraphQLContext](https://github.com/graphql-java/graphql-java/blob/e6367faada5da8b098125bd3f055abd40697b21f/src/main/java/graphql/GraphQLContext.java#L80) has a method that enables us to treat the context more as a map that can be namespaced, with let's say, a "DGS-KotlinCorutineContext".

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/CoroutineDataFetcherTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        val concurrentTime = measureTimeMillis {
+            val executionResult = build.execute(
+                ExecutionInput.newExecutionInput().context(context).query(
+                    """
+            {
+                first: concurrent(from: 1, to: 10)
+                second: concurrent(from: 1, to: 10)               
+                third: concurrent(from: 1, to: 10)               
+                fourth: concurrent(from: 1, to: 10)
```
awgr, line 87: is it guaranteed that `first` will correspond to `1` in the execution result?

paulbakker: What do you mean by corresponding to `1`? What is `1` here?

berngp (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
paulbakker: It turns out the helper methods are not available yet in the Spring version we're depending on - the methods are all "internal".
It makes sense to use this when we bump the Spring version though.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
paulbakker: @kilink I've done some more testing with the different dispatchers, and see the following:


With `Default`
```
2021-07-13 15:13:46.103  INFO 8474 --- [atcher-worker-2] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Ozark
2021-07-13 15:13:46.103  INFO 8474 --- [atcher-worker-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Stranger Things
2021-07-13 15:13:46.103  INFO 8474 --- [atcher-worker-3] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for The Crown
2021-07-13 15:13:46.103  INFO 8474 --- [atcher-worker-4] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Dead to Me
2021-07-13 15:13:46.104  INFO 8474 --- [atcher-worker-5] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Orange is the New Black
2021-07-13 15:13:51.112  INFO 8474 --- [atcher-worker-4] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Dead to Me
2021-07-13 15:13:51.113  INFO 8474 --- [atcher-worker-1] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Orange is the New Black
2021-07-13 15:13:51.112  INFO 8474 --- [atcher-worker-3] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Stranger Things
2021-07-13 15:13:51.112  INFO 8474 --- [atcher-worker-5] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Ozark
2021-07-13 15:13:51.113  INFO 8474 --- [atcher-worker-2] c.e.d.datafetchers.ReviewsDataFetcher    : After delay The Crown
2021-07-13 15:13:53.123  INFO 8474 --- [atcher-worker-2] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7020ms
2021-07-13 15:13:53.123  INFO 8474 --- [atcher-worker-3] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7045ms
2021-07-13 15:13:53.124  INFO 8474 --- [atcher-worker-4] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7021ms
2021-07-13 15:13:53.124  INFO 8474 --- [atcher-worker-1] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7021ms
2021-07-13 15:13:53.126  INFO 8474 --- [atcher-worker-5] c.e.d.i.ExampleTracingInstrumentation    : Total execution time: 7130ms
2021-07-13 15:13:53.127  INFO 8474 --- [atcher-worker-5] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7027ms
```

With `IO`
```
2021-07-13 15:14:47.925  INFO 8569 --- [atcher-worker-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Stranger Things
2021-07-13 15:14:47.925  INFO 8569 --- [atcher-worker-3] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Ozark
2021-07-13 15:14:47.925  INFO 8569 --- [atcher-worker-2] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for The Crown
2021-07-13 15:14:47.926  INFO 8569 --- [atcher-worker-5] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Dead to Me
2021-07-13 15:14:47.926  INFO 8569 --- [atcher-worker-6] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Orange is the New Black
2021-07-13 15:14:52.936  INFO 8569 --- [atcher-worker-3] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Ozark
2021-07-13 15:14:52.936  INFO 8569 --- [atcher-worker-6] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Orange is the New Black
2021-07-13 15:14:52.936  INFO 8569 --- [atcher-worker-2] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Dead to Me
2021-07-13 15:14:52.936  INFO 8569 --- [atcher-worker-5] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Stranger Things
2021-07-13 15:14:52.936  INFO 8569 --- [atcher-worker-1] c.e.d.datafetchers.ReviewsDataFetcher    : After delay The Crown
2021-07-13 15:14:54.942  INFO 8569 --- [atcher-worker-1] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7017ms
2021-07-13 15:14:54.943  INFO 8569 --- [atcher-worker-2] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7018ms
2021-07-13 15:14:54.943  INFO 8569 --- [atcher-worker-5] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7042ms
2021-07-13 15:14:54.944  INFO 8569 --- [atcher-worker-3] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7022ms
2021-07-13 15:14:54.946  INFO 8569 --- [atcher-worker-6] c.e.d.i.ExampleTracingInstrumentation    : Total execution time: 7121ms
2021-07-13 15:14:54.947  INFO 8569 --- [atcher-worker-6] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7021ms
```

`Unconfined`
```
2021-07-13 15:11:28.186  INFO 8281 --- [nio-8080-exec-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Stranger Things
2021-07-13 15:11:28.196  INFO 8281 --- [nio-8080-exec-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Ozark
2021-07-13 15:11:28.196  INFO 8281 --- [nio-8080-exec-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for The Crown
2021-07-13 15:11:28.196  INFO 8281 --- [nio-8080-exec-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Dead to Me
2021-07-13 15:11:28.197  INFO 8281 --- [nio-8080-exec-1] c.e.d.datafetchers.ReviewsDataFetcher    : before delay for Orange is the New Black
2021-07-13 15:11:33.205  INFO 8281 --- [DefaultExecutor] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Stranger Things
2021-07-13 15:11:35.214  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 7045ms
2021-07-13 15:11:35.215  INFO 8281 --- [DefaultExecutor] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Ozark
2021-07-13 15:11:37.222  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 9027ms
2021-07-13 15:11:37.223  INFO 8281 --- [DefaultExecutor] c.e.d.datafetchers.ReviewsDataFetcher    : After delay The Crown
2021-07-13 15:11:39.231  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 11035ms
2021-07-13 15:11:39.231  INFO 8281 --- [DefaultExecutor] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Dead to Me
2021-07-13 15:11:41.238  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 13042ms
2021-07-13 15:11:41.238  INFO 8281 --- [DefaultExecutor] c.e.d.datafetchers.ReviewsDataFetcher    : After delay Orange is the New Black
2021-07-13 15:11:43.241  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Total execution time: 15136ms
2021-07-13 15:11:43.243  INFO 8281 --- [DefaultExecutor] c.e.d.i.ExampleTracingInstrumentation    : Async datafetcher 'Show.reviews' took 15046ms
```

As the logs indicate, when using `unconfined` we don't actually seem to get async behavior. I'm still not completely understanding all of it, so I might just be missing something, but this doesn't seem to be the behavior we want.
I think we should keep `Default`?

paulbakker: @kilink @berngp For the creation of the scope; instead of creating the scope on the `DgsContext`, per request, we could just create a new on for each (suspend) data fetcher: 

```
 val launch = CoroutineScope(Dispatchers.Default).async {
                return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
            }
```

The result seems to be the same looking at how it executes.

awgr: Leveraging some context from work I've done in the past related to the common pool,  and _unless_ things have changed significantly (or I'm mistaken!), it's my understanding that the common pool implementation backing the default dispatcher uses java's the common fork join pool (as we don't have a SM installed in the JVM), so we should _ideally_ use it primarily for non-blocking workloads (i.e. avoid using it for any subsequent requests or IO to e.g. a database, or other resource).

Why? Suppose we have some library code that sums integers over the common FJP - its execution will potentially be blocked by the data fetcher tasks that have been scheduled on that same pool, which reduces the predictability and consistency of that implementation.

Also I agree with @kilink that the max(cores - 1, 1) capacity of the default pool isn't (as far as I understand) ideal for the dgs framework use case; i.e. it doesn't seem particularly superior to many other default threading policies we could invent. 

Off the cuff, it seems like using unconfined as @kilink suggested ensures the threading policy for the coroutine is identical to the threading policy for the requests themselves, which ensures that the DGS framework does not introduce a policy which is a performance degradation to any specific DGS's implementation beyond the configuration of that web server. It seems like making it configurable is a plus for DGS implementers, as that ensures they can provide a pool specialized to their workload (if necessary), and this mirrors the configurability of our runtime environment to some extent.

If we wanted to specify a separate independent pool as a resource for the DGS framework, and let data fetchers use that pool, I think it could be fine as well, but that's a design decision I don't have sufficient data to really form an opinion about (and depends a bit on the workload implemented by the dgs itself, although is very likely to be a non-issue for _most_ DGS implementations).

IMO, the best option seems to be leaving the implementation of a pool up to the DGS owner who will know the most about their specific workloads, but providing a pool is a nice option as well as it's unlikely to negatively impact most workloads.

I think using the default pool is likely to be replaced by one of those strategies after some DGS owner encounters an issue caused by blocking usage of the common pool.


PS: There's another minor issue here that if parallelism is 1 (i.e. 2 cores, or 1 core) I _believe_ it will create a new thread for every task due to how the parallelism of the common pool is implemented, which may surprise some DGS owners.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/CoroutineDataFetcherTest.kt
awgr: Sorry, I mean, execution order of first, second, third, and fourth is not guaranteed, right? This isn't a critique, just making sure I understand the end-to-end implications of the async model for a data fetcher

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
kilink: CoroutinesUtils.invokeSuspendingFunction is public, and was added in Spring 5.2, which is what we depend on. You're right about the other one, isSuspendingFunction was added in 5.3 unfortunately.

kilink: Nevermind, I see even CoroutinesUtils.invokeSuspendingFunction was internal until 5.3...

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
paulbakker: After more experimenting and offline discussion, we landed on the following.

The `DgsSchemaProvider` uses Unconfined, which means the developer is now in control of creating async behavior. Since there wasn't really a benefit of storing the scope on `DgsContext` I have removed that.

```kotlin
val launch = CoroutineScope(Dispatchers.Unconfined).async {
      return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
}
```

If the suspend function doesn't further specify async behavior, the result is practically that the datafetcher runs synchronously. However, the developer can control this behavior by using `withContext` or `async` in the datafetcher, and for example, specify an `Executor`.
In Spring, the default available `Executor` would be a good start. The provided test shows using a manually created thread pool.

```


//For test purposes. In a real application, you should probably @Autowired Executor executor
val executor = Executors.newFixedThreadPool(8)

@DgsQuery
suspend fun concurrent(@InputArgument from: Int, to: Int): Int = coroutineScope {
     var sum = 0
     withContext(executor.asCoroutineDispatcher()) {
           println("before $from")
            repeat(from.rangeTo(to).count()) {
                 sum++

                  //Forcing a blocking call to demonstrate running with a thread pool
                  Thread.sleep(50)
              }
   
               println("after $from")

               sum
    }
 }

```

This code contains a blocking call (Thread.sleep). I runs in parallel with other datafetcher however, because of specifying the `executor`.

Note that when manually creating an Executor, you might be missing data that's stored in `ThreadLocal`, which is for example common for authorization data in Spring MVC. This is the same for using `CompletableFuture`.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/CoroutineDataFetcherTest.kt
paulbakker: > Sorry, I mean, execution order of first, second, third, and fourth is not guaranteed, right? This isn't a critique, just making sure I understand the end-to-end implications of the async model for a data fetcher

correct

PR #494, opened 2021-07-28
open DefaultDataFetcherExceptionHandler

sbilello:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

Changes in this PR
----

I would like to extend `DefaultDataFetcherExceptionHandler` to reduce boilerplate code

Alternatives considered
----

Copying and Paste in another `Class`

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt
@@ -27,7 +27,7 @@ import org.slf4j.LoggerFactory
  * Default DataFetcherExceptionHandler used by the framework, can be replaced with a custom implementation.
  * The default implementation uses the Common Errors library to return GraphQL errors.
  */
-class DefaultDataFetcherExceptionHandler : DataFetcherExceptionHandler {
+open class DefaultDataFetcherExceptionHandler : DataFetcherExceptionHandler {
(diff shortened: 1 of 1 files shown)


srinivasankavitha (commented):
Have you looked at these docs? This should let you add custom exception handling: https://netflix.github.io/dgs/error-handling/#mapping-custom-exceptions

paulbakker:
@srinivasankavitha I think the request is to make it possible to _extend_ the default behavior. 
If you create a custom exception handler, it replaces the default behavior currently.

@sbilello I'm not sure extending the default is the best option to achieve this. I'll look into some options.

sbilello:
@srinivasankavitha I wanted to extend the current exception handler with my logic and not copy and paste the current logic. Right now that class is instantiated by the `DgsAutoConfiguration` conditionally `@ConditionalOnMissingBean`, 
so my handler would replace the current one. I wanted to add logic to the existing one by extending it with a child class. Thanks for your review!

srinivasankavitha:
Actually what I meant to point out is that you can delegate to the default
exception handler in your custom exception handler. Thus you won't need to
copy any logic. The docs I linked do mentioned "either extend or delegate"
which is misleading. The recommended pattern is to just add your own
exception handler and delegate to the default where applicable as shown
here:
https://netflix.github.io/dgs/error-handling/#mapping-custom-exceptions

I will fix the docs as well.

sbilello:
Thanks, @srinivasankavitha!

PR #499, opened 2021-07-28
Fix assertSubscription helper method

kilink:
Verification was never being triggered on the StepVerifier instance, and
the assertion was looking for the wrong map key.


berngp:
Thanks @kilink !

PR #500, opened 2021-07-28
Avoid chaining collection operations

kilink:
Avoid chaining collection operations in some places where intermediate
collections would be allocated and thrown away. Also, switch to Kotlin
idioms, remove unnecessary null assertions, etc.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -96,12 +96,12 @@ class DgsSchemaProvider(
 
     fun schema(schema: String? = null, fieldVisibility: GraphqlFieldVisibility = DefaultGraphqlFieldVisibility.DEFAULT_FIELD_VISIBILITY): GraphQLSchema {
         val startTime = System.currentTimeMillis()
-        val dgsComponents = applicationContext.getBeansWithAnnotation(DgsComponent::class.java)
+        val dgsComponents = applicationContext.getBeansWithAnnotation(DgsComponent::class.java).values
         val hasDynamicTypeRegistry =
-            dgsComponents.values.any { it.javaClass.methods.any { m -> m.isAnnotationPresent(DgsTypeDefinitionRegistry::class.java) } }
+            dgsComponents.any { it.javaClass.methods.any { m -> m.isAnnotationPresent(DgsTypeDefinitionRegistry::class.java) } }
 
         var mergedRegistry = if (schema == null) {
-            findSchemaFiles(hasDynamicTypeRegistry = hasDynamicTypeRegistry).map {
+            findSchemaFiles(hasDynamicTypeRegistry = hasDynamicTypeRegistry).asSequence().map {
                 InputStreamReader(it.inputStream, StandardCharsets.UTF_8).use { reader -> SchemaParser().parse(reader) }
             }.fold(TypeDefinitionRegistry()) { a, b -> a.merge(b) }
         } else {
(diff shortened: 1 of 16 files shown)


awgr (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -445,31 +449,30 @@ class DgsSchemaProvider(
...
         }
     }
 
     private fun getValueAsOptional(value: Any?, parameter: Parameter) =
-        if (value == null && parameter.type.isAssignableFrom(Optional::class.java)) {
-            Optional.empty<Any>()
-        } else if (parameter.type.isAssignableFrom(Optional::class.java)) {
-            Optional.of(value!!)
+        if (parameter.type.isAssignableFrom(Optional::class.java)) {
+            Optional.ofNullable(value)
```
awgr, line 460: removing this unreachable branch here?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -445,31 +449,30 @@ class DgsSchemaProvider(
                 return@async method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
             }
 
-            launch?.asCompletableFuture()
+            launch.asCompletableFuture()
```
awgr, line 452: launch no longer nullable?

kilink: This might be left over from some refactoring when it was nullable? But yeah it's not nullable on master, so the `?.` isn't necessary.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
kilink: I don't know that it was unreachable, just that it could be collapsed into a single case by switching to `Optional.ofNullable`, and thus reads a bit cleaner to me (and avoids `!!`).

berngp (approved):
Thanks for cleaning the code @kilink

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -154,15 +155,16 @@ class DgsSchemaProvider(
     }
 
     private fun invokeDgsTypeDefinitionRegistry(dgsComponent: Any, registry: TypeDefinitionRegistry): TypeDefinitionRegistry? {
-        return dgsComponent.javaClass.methods.filter { it.isAnnotationPresent(DgsTypeDefinitionRegistry::class.java) }
+        return dgsComponent.javaClass.methods.asSequence()
```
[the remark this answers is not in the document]
kilink: [filter](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter.html) and [map](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map.html) both return lists, but at the end of the chain it's a `reduceOrNull` call anyway so those intermediate lists are just wasteful in terms of allocations.

To put it another way, if it were unrolled into an imperative style, it would be the equivalent of doing this:

```kotlin
val filtered = mutableListOf<Method>()
for (method in dgsComponent.javaClass.methods) {
    if (method.isAnnotationPresent(DgsTypeDefinitionRegistry::class.java)) {
        filtered += method
    }
}

val registries = mutableListOf<TypeDefinitionRegistry>()
for (method in filtered) {
    registries += method.invoke(/* ... */) as TypeDefinitionRegisty
}

var typeDefRegistry: TypeDefinitionRegistry? = null
for (registry in registries) {
    typeDefRegistry = typeDefRegistry?.merge(registry) ?: registry
}
return registry
```

In fact if you look at the generated bytecode, it's likely doing something similar under the hood. Putting `asSequence()` in there as the first step just means it's all done in 1 shot, so things are not iterated 3 times, and the intermediate List allocations don't even need to happen.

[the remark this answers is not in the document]
berngp: TIL, thanks for the explanation.

srinivasankavitha (approved):

PR #501, opened 2021-07-28
Avoid instantiating Jackson mappers per request

kilink:
An ObjectMapper was being instantiated per request, which can be
expensive.

### DIFF
--- a/graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
+++ b/graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
@@ -17,7 +17,7 @@
 package com.netflix.graphql.dgs.webflux.handlers
 
 import com.fasterxml.jackson.annotation.JsonProperty
-import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.graphql.dgs.reactive.DgsReactiveQueryExecutor
 import org.reactivestreams.Publisher
 import org.reactivestreams.Subscription
@@ -41,7 +41,7 @@ class DgsReactiveWebsocketHandler(private val dgsReactiveQueryExecutor: DgsReact
     private val resolvableType = ResolvableType.forType(OperationMessage::class.java)
     private val subscriptions = ConcurrentHashMap<String, MutableMap<String, Subscription>>()
     private val decoder = Jackson2JsonDecoder()
-    private val encoder = Jackson2JsonEncoder()
+    private val encoder = Jackson2JsonEncoder(decoder.objectMapper)
 
     companion object {
         const val GQL_CONNECTION_INIT = "connection_init"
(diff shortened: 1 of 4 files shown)


kilink:
It might make sense to share an ObjectMapper bean across components, but that would be a bigger change I suspect.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
```diff
@@ -74,9 +74,10 @@ class DgsReactiveWebsocketHandler(private val dgsReactiveQueryExecutor: DgsReact
                             )
                         )
                         GQL_START -> {
-                            val queryPayload =
-                                jacksonObjectMapper().convertValue(operationMessage.payload, QueryPayload::class.java)
-                            logger.debug("Starting subscription $queryPayload for session ${webSocketSession.id}")
+                            val queryPayload = decoder.objectMapper.convertValue<QueryPayload>(
+                                operationMessage.payload ?: error("payload == null")
```
awgr, line 78: any debugging context worth logging about the caller?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
```diff
@@ -112,23 +116,28 @@ class DgsReactiveWebsocketHandler(private val dgsReactiveQueryExecutor: DgsReact
...
 
                         GQL_STOP -> {
-                            subscriptions[webSocketSession.id]?.get(operationMessage.id)?.cancel()
-                            subscriptions[webSocketSession.id]?.remove(operationMessage.id)
-                            logger.debug("Client stopped subscription ${operationMessage.id} for connection ${webSocketSession.id}")
+                            subscriptions[webSocketSession.id]?.remove(operationMessage.id)?.cancel()
+                            logger.debug(
+                                "Client stopped subscription {} for connection {}",
+                                operationMessage.id, webSocketSession.id
+                            )
```
awgr, line 132: love this

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -42,11 +42,11 @@ import java.util.*
 @RestController
 open class DgsSSESubscriptionHandler(open val dgsQueryExecutor: DgsQueryExecutor) {
     open val logger = LoggerFactory.getLogger(DgsSSESubscriptionHandler::class.java)
+    private val mapper = jacksonObjectMapper()
```
awgr, line 45: why not inject these?

awgr:
LGTM

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
kilink: Yeah Ii think that is the way to go ultimatley, but it requires a bit more care. For instance, I imagine we'd want qualify the mapper used internally in DGS, and not just use the default ObjectMapper, and does it make sense to expose a builder / customizer as a hook point for people to fiddle with the configuration, etc.

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/handlers/DgsReactiveWebsocketHandler.kt
kilink: I am not sure what the correct thing to do here is; this is really just making explicit the behavior of the existing code, which would have passed the null value silently and blown up a few lines down with an NPE when it tried to access `queryPayload.query`. I guess logging a warning could make sense here.

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
paulbakker: Yes, we shouldn't inject the default one. Folks set those up for various reasons, not necessarily expecting the framework to use that as well. We've seen issues with this previously.

A DGS specific one would work.

Issue #438, opened 2021-06-25
feature: Add Subscription support to GraphQLClient


hantsy:
Currently, it is easy to handle general queries, but how to subscribe to a  `Subscription` with the `GraphQLClient`?

hantsy:
I mean  [Java GraphQL Client](https://netflix.github.io/dgs/advanced/java-client/) not `DgsQueryExecutor `.

srinivasankavitha:
Thanks for creating the issue. We will add it to our backlog.

paulbakker:
Merged in #522

PR #522, opened 2021-08-09
Websocket Subscription Client using Webflux

callumforrester:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Draft implementation of websocket client using webflux as requested in issue #438 discussed in #510 with @paulbakker, @osi and @berngp.
This is meant to showcase a particular approach so a merge is not necessarily expected. If the approach is deemed acceptable I will either make a new PR or make this one more production-ready, add tests etc.

Considerations/Questions:

* Are we happy using `WebSocketClient` from Webflux? It is somewhat lower-level than `WebClient` in that it is just meant to facilitate connection/disconnection and leave the protocol implementation to the user.
* This implementation gives 1 subscription per websocket connection. The protocol supports multiple subscriptions per websocket connection with a little extra bookkeeping, is that desirable?
* The data classes describing the protocol messages are now duplicated across `graphql-dgs-client` and `graphql-dgs-subscriptions-websockets`, is there anywhere they can be consoladated?
* Need to ensure backpressure is handled correctly.
* None of the existing client interfaces provide a flux, either a new interface is needed (as done here) or a way must be found to get a flux out of a `GraphQLResponse`. 

Alternatives considered
----

Using `spring-websocket` instead of webflux, see: https://github.com/Netflix/dgs-framework/discussions/510#discussioncomment-1126399


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
```diff
@@ -43,6 +45,19 @@ interface MonoGraphQLClient {
     ): Mono<GraphQLResponse>
 }
 
+interface ReactiveGraphQLClient {
+    fun reactiveExecuteQuery(
+        query: String,
+        variables: Map<String, Any>,
+    ): Publisher<GraphQLResponse>
```
osi, line 52: As this interface is already exposing Project Reactor types, I think it's safe to return `Flux` here?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/WebsocketGraphQLClient.kt
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    override fun reactiveExecuteQuery(
+        query: String,
+        variables: Map<String, Any>,
+        operationName: String?,
+    ): Publisher<GraphQLResponse> {
+        val uri = URI(url)
+
+        // Sinks are used as buffers, incoming messages from the server are
```
osi, line 61: I believe an implementation without sinks should be possible. If you want to buffer due to back pressure, I'd recommend using one of the operators on the stream.

osi:
re:
> The protocol supports multiple subscriptions per websocket connection with a little extra bookkeeping, is that desirable?

i believe so.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/WebsocketGraphQLClient.kt
callumforrester: Because `client.execute` must return a `Mono<Void>`, the alternative to sinks is something like

```kotlin
return Flux.create {
    emitter -> client.execute(uri) { session -> exchange(emitter, session) }
}
```

Would that be preferred?

osi: ah, yes, you're correct. i'm not sure that is necessarily better.

what i _would_ try to do is remove the naked `client.subscribe()` call and tie that into the overall flow. (perhaps by wrapping everything inside of the method in a `Flux.defer`)

the returned `Publisher` of this method should be able to be subscribed to multiple times without re-invoking `reactiveExecuteQuery`. the connection shouldn't be active until the returned `Publisher` is subscribed to.

callumforrester: Have created a single `Mono` for connection to the server which is cached to fulfil https://github.com/Netflix/dgs-framework/pull/522#issuecomment-895369056, the mono is chained to the flux of messages about a particular subscription, it is also deferred so connection does not occur until the first time the user calls `subscribe`

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/WebsocketGraphQLClient.kt
```diff
@@ -0,0 +1,280 @@
...
+
+data class QueryPayload(
+    @JsonProperty("variables")
+    val variables: Map<String, Any>?,
+    @JsonProperty("extensions")
+    val extensions: Map<String, Any> = emptyMap(),
+    @JsonProperty("operationName")
+    val operationName: String?,
+    @JsonProperty("query")
+    val query: String)
```
callumforrester, line 280: @osi These are near-duplicates of the structures in `graphql-dgs-subscriptions-websockets`. Is there somewhere common they can live?

osi: I'll defer to @paulbakker  - may be an opportunity for a new shared module with common subscription stuff.

callumforrester:
@paulbakker and @osi I'm now reasonably happy with the concurrency use cases but the auto-reconnect is proving to be more difficult. I have made a lot of assumptions about this client as a single-use object which I've been slowly unknotting for a few days.

I can think of a few different approaches, although I'm not sure how easy they are. For now I just want to think about hypothetical behavoir:
1. If the socket is closed, complete any subscription Fluxes. Get the client to check if the connection is stale when the user calls `subscribe()` and reconnect if necessary. 
2. If the socket is closed, attempt reconnection, if reconnection fails, emit an error. If it succeeds resubscribe.
3. If the socket is closed, emit an error. User has to make a new client object themselves
4. If the socket is closed, emit an error. User can resubscribe manually or automatically using `retry()`

I welcome any thoughts and comments

callumforrester:
@paulbakker/@osi After some more thought:

5. Although the apollo protocol allows multiple subscriptions per session not every client uses this facility. We could make each subscription on a new session as GraphQL playground does. This is also how SSE works, that would make the whole implementation considerably simpler and less likely to have bugs.

callumforrester:
I've pushed a best-effort implementation. I think the trouble is that buffered messages to/from the server may go to/from the wrong session

paulbakker:
@callumforrester I've merged this PR. It looked great, and we can always further improve it later.
Thanks a lot for this great work!

PR #536, opened 2021-08-15
add @DsgDirective

### LINKED ISSUE #419: feature: Add @DgsDirective to simplify the directive declaration
I followed the steps of the discussion posts and tried to add a simple uppercase directive in my sample project, but it does not work as expected.

The uppercase directive wring class.

```java
public class UppercaseDirectiveWiring implements SchemaDirectiveWiring {
    @Override
    public GraphQLFieldDefinition onField(SchemaDirectiveWiringEnvironment<GraphQLFieldDefinition> env) {

        var field = env.getElement();
        var parentType = env.getFieldsContainer();

        var originalDataFetcher = env.getCodeRegistry().getDataFetcher(parentType, field);
        var dataFetcher = DataFetcherFactories.wrapDataFetcher(originalDataFetcher,
                (dataFetchingEnvironment, value) -> {
                    if (value instanceof String s) {
                        return s.toUpperCase();
                    }
                    return value;
                }
        );

        env.getCodeRegistry().dataFetcher(parentType, field, dataFetcher);
        return field;
    }
}
```

And I created a bean `@DgsRuntimeWiring` in the data fetchers.  The complete codes is here: https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-codegen/src/main/java/com/example/demo/gql/PostsDataFetcher.java#L24

```java
@DgsComponent
public class PostsDataFetcher {

    @DgsRuntimeWiring
    RuntimeWiring.Builder runtimeWiring(RuntimeWiring.Builder builder) {
        return builder.directive("uppercase", new UppercaseDirectiveWiring());
    }
```

The partial content of the schema file:

```gql
directive @uppercase on FIELD_DEFINITION

scalar LocalDateTime
scalar UUID

type Post{
    id: ID!
    title: String! @uppercase
    content: String
    comments: [Comment]
    status: PostStatus
    createdAt: LocalDateTime
    authorId:String
    author:Author
}
```

When running the application, and test it on Graphiql, but it does not work. Not sure where is wrong.

This directive is working with my other Graphql example written in GraphQL Java and Spring Graphql.

Check: https://github.com/hantsy/spring-graphql-sample/tree/master/graphql-java/src/main/java/com/example/demo/gql/directives

## Expected behavior

Similar to the existing `@DgsScalar`, I hope to add a `@DgsDirective` and apply it on the user's `SchemaDirectiveWiring` class directly to simplify the registration of a directive. eg. 

```java
@DgsDirective(name="uppercase")
public class UppercaseDirectiveWiring implements SchemaDirectiveWiring {}
```

hantsy:
fixes #419

### DIFF
--- a/graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
+++ b/graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs;
+
+import org.springframework.stereotype.Component;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Mark a class as a custom Directive implementation that gets registered to the framework.
+ *
+ * @author Hantsy Bai <hantsy@gmail.com>
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Component
+public @interface DgsDirective {
+    String name() default "";
+}
--- a/graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
+++ b/graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
@@ -25,8 +25,6 @@ import java.lang.annotation.Target;
 
 /**
  * Mark a class as a custom Directive implementation that gets registered to the framework.
- *
- * @author Hantsy Bai <hantsy@gmail.com>
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -34,10 +34,7 @@ import graphql.language.InterfaceTypeDefinition
 import graphql.language.TypeName
 import graphql.language.UnionTypeDefinition
 import graphql.schema.*
-import graphql.schema.idl.RuntimeWiring
-import graphql.schema.idl.SchemaParser
-import graphql.schema.idl.TypeDefinitionRegistry
-import graphql.schema.idl.TypeRuntimeWiring
+import graphql.schema.idl.*
 import graphql.schema.visibility.DefaultGraphqlFieldVisibility
 import graphql.schema.visibility.GraphqlFieldVisibility
 import kotlinx.coroutines.CoroutineScope
@@ -53,6 +50,7 @@ import org.springframework.core.annotation.MergedAnnotations
 import org.springframework.core.io.Resource
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver
 import org.springframework.util.ReflectionUtils
+import org.springframework.util.StringUtils
 import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.ValueConstants
@@ -123,6 +121,7 @@ class DgsSchemaProvider(
             .mapNotNull { dgsComponent -> invokeDgsTypeDefinitionRegistry(dgsComponent, mergedRegistry) }
             .fold(mergedRegistry) { a, b -> a.merge(b) }
         findScalars(applicationContext, runtimeWiringBuilder)
+        findDirectives(applicationContext, runtimeWiringBuilder)
         findDataFetchers(dgsComponents, codeRegistryBuilder, mergedRegistry)
         findTypeResolvers(dgsComponents, runtimeWiringBuilder, mergedRegistry)
         findEntityFetchers(dgsComponents)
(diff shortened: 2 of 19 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -547,6 +546,22 @@ class DgsSchemaProvider(
...
+    private fun findDirectives(applicationContext: ApplicationContext, runtimeWiringBuilder: RuntimeWiring.Builder) {
+        applicationContext.getBeansWithAnnotation(DgsDirective::class.java).forEach { (_, directiveComponent) ->
+            val annotation = directiveComponent::class.java.getAnnotation(DgsDirective::class.java)
+            when (directiveComponent) {
+                is SchemaDirectiveWiring ->
+                    if (StringUtils.hasText(annotation.name)) runtimeWiringBuilder.directive(
+                        annotation.name,
+                        directiveComponent
+                    )
+                    else runtimeWiringBuilder.directiveWiring(directiveComponent)
```
WtfJoke, line 558:
```suggestion
                    if (StringUtils.hasText(annotation.name)) {
                        runtimeWiringBuilder.directive(
                            annotation.name,
                            directiveComponent
                        )
                    } else {
                        runtimeWiringBuilder.directiveWiring(directiveComponent)
                    }
```

Looking at the project, I think they preferr having if/else with curly braces {}. Personally I think its also better readable

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/WordFilterDirective.kt
```diff
@@ -0,0 +1,46 @@
...
+import graphql.schema.idl.SchemaDirectiveWiringEnvironment
+import java.util.function.BiFunction
+
+/**
+ *  An `@DgsDirective` example for test purpose.
+ *
+ *  @author Hantsy Bai <hantsy@gmail.com>
+ */
+@DgsDirective
+class WordFilterDirective : SchemaDirectiveWiring {
```
WtfJoke, line 31: If you provide 2 examples, it probably makes more sense to add another type of directive (for example INPUT_FIELD_DEFINITION or ARGUMENT_DEFINITION) as an example.

WtfJoke (commented):
I am not a contributor, but here are my 2cents :)

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Mark a class as a custom Directive implementation that gets registered to the framework.
+ *
+ * @author Hantsy Bai <hantsy@gmail.com>
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Component
+public @interface DgsDirective {
+    String name() default "";
```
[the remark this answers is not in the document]
hantsy: This should be applied in the schema file.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/WordFilterDirective.kt
hantsy: WordFilterDirective is an unnamed directive that will be applied globally.

graphql-dgs-mocking/build.gradle.kts
```diff
@@ -16,6 +16,8 @@
 
 dependencies {
     api("com.graphql-java:graphql-java")
-    implementation("com.github.javafaker:javafaker:1.+")
+    implementation("com.github.javafaker:javafaker:1.+") {
+        exclude("org.yaml", "snakeyaml")
+    }
```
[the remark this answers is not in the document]
hantsy: Moved to a new PR.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/WordFilterDirective.kt
WtfJoke: Is that something desirable? Genious question

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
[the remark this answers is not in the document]
WtfJoke: You could argue that the directive name is also applied in the schema file 😜 
I still could imagine it could be a nice addition. see my second point.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/WordFilterDirective.kt
hantsy: It is just a feature supported in GraphQL Java.

WtfJoke (approved):

paulbakker:
@hantsy I think this looks good in concept. There is a merge conflict currently.

Will do a bit more testing as well.

berngp:
@WtfJoke thanks for the review as well as the suggestions that you gave to create the PR!

hantsy:
@paulbakker  fixed the conflict. 

As @WtfJoke suggested, to introduce `@DgsDirective`, I think it is better to add a section in the **directive** related chapter to describe how to use this `@DgsDirective` to simplify the directive declaration.

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsDirective.java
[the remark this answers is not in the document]
paulbakker: The name is all that's used to register the directive in `graphql-java`, so I think the list of targets isn't actually useful on the annotation itself.

paulbakker (changes requested):
I just found that the CI workflow didn't trigger for this. It's supposed to trigger on `push`, but I guess that doesn't work for branches outside the repository. I've updated the workflow definition to also trigger on `pr` again, so hopefully it will happen on the next push.

Locally I found the build is failing with some test failures (because the `applicationContext` mocks are now missing a call). Would you want to take a look at that?

Also, the JavaDoc fails because of the `@author` you added. Mind removing that? We're not using it anywhere else.

Once the build is passing we'll merge this. I tested it by trying out some directives and it works great 👍

hantsy:
@paulbakker I added the `DgsDirective` stubs in every test that mocked the `applicaitonContext`,  it is a little tedious.  *Hope there is some way to avoid this*. 

And the new commit triggered the CI build, it it great and but there is an error stopped the building progress. 

```bash 
* What went wrong:
Execution failed for task ':graphql-dgs:ktlintMainSourceSetCheck'.
> A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.ConsoleReportWorkAction
   > KtLint found code style violations. Please see the following reports:
     - /home/runner/work/dgs-framework/dgs-framework/graphql-dgs/build/reports/ktlint/ktlintMainSourceSetCheck/ktlintMainSourceSetCheck.txt
```

hantsy:
>The name is all that's used to register the directive in `graphql-java`, so I think the list of targets isn't actually useful on the annotation itself.

The `@Target` indicates where this annotation (`@DgsDirective`) should be applied on, right?  I think the type is essential. Maybe method( similar to `@Bean` declarations) is also an alternative place in a `@Configuration` class.

paulbakker:
> > The name is all that's used to register the directive in `graphql-java`, so I think the list of targets isn't actually useful on the annotation itself.
> 
> The `@Target` indicates where this annotation (`@DgsDirective`) should be applied, right? I think the `type` target is essential at the moment. Maybe method( similar to `@Bean` declarations) is also an alternative place in a `@Configuration` class.

Sorry for the confusion. That was a reply to @WtfJoke's suggestion to add more arguments to the annotation. Basically, I was agreeing with what you have 👍

paulbakker:
Thanks for the work @hantsy !

Issue #400, opened 2021-06-13
bug: SnakeYaml android classfied jar is not found 


hantsy:
Latest DSG platform, when using Gradle as build tools, an error has occurred when building the application.

```
Could not find snakeyaml-1.28-android.jar (org.yaml:snakeyaml:1.28).
```

I have to exclude snakeyaml to overcome this issue. 

https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-codegen/build.gradle

But the maven project does not report this issue.

https://github.com/hantsy/spring-graphql-sample/blob/master/dgs

## Expected behavior

```gradle
implementation "com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter"
```
## Actual behavior

```gradle
implementation "com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter", {
      exclude group: 'org.yaml', module: 'snakeyaml'
}
```

berngp:
We have both a Java and Kotlin examples that run on every single pr as part of our build pipeline. I also see the dependency you mentioned available.  This was probably a transient issue resolving the artifact.

- https://github.com/Netflix/dgs-examples-java
- https://github.com/Netflix/dgs-examples-kotlin

hantsy:
When building the project, it only occurred in my Gradle project, not happened in the Maven project. 
I am not sure what happened in the Gradle dependency resolving.

paulbakker:
Closing because haven't seen this issue and can't reproduce it.

hantsy:
@paulbakker  Check my `gradlew build --scan` result of the **dgs-framework** on my **Windows**, check the result here: https://scans.gradle.com/s/vb3kn46n5whgk

And this error only occurred under Windows/Gradle. 

I also tried to [create some Maven projects](https://github.com/hantsy/spring-graphql-sample/tree/master/dgs), they did not repot this error.

hantsy:
@paulbakker  I have to add [this change](https://github.com/Netflix/dgs-framework/blob/c37ffcca0bb4f4c11909f27ba4d7afb2eb73e117/graphql-dgs-mocking/build.gradle.kts#L19-L21)  in the Gradle build script of dgs-mocking module to make sure it is built successfully under Windows.


```kotlin
    implementation("com.github.javafaker:javafaker:1.+") {
        exclude("org.yaml", "snakeyaml")
    }
```

hantsy:
[...]

PR #541, opened 2021-08-17
chore: remove snakeyaml dep from graphql-dgs-mocking

hantsy:
fixes issue #400

### DIFF
--- a/graphql-dgs-mocking/build.gradle.kts
+++ b/graphql-dgs-mocking/build.gradle.kts
@@ -16,6 +16,8 @@
 
 dependencies {
     api("com.graphql-java:graphql-java")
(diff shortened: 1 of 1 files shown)


hantsy:
As discussed in issue #400, if I do not exclude this `snakeyaml`(with a classifier **android**), it will stop me to build the whole **dgs-framework** under my **Windows** system.

It is also a barrier when creating some [sample codes](https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-codegen/build.gradle#L20), also encountered build issues if I do not exclude it.

paulbakker:
Thanks, @hantsy! I still don't understand why we're not seeing the issue consistently, but this is good anyway.

PR #549, opened 2021-08-19
Custom implementation of convertValue for input types

paulbakker:
We use Jackson's `convertValue` to convert the `Map` of input values created by `graphql-java` to a Java class for ease-of-use in data fetchers.
Using Jackson's `convertValue` has always been problematic. It serializes the input `Map` and then deserializes it again into the target `Class`. In this process, it doesn't know that `graphql-java` already took care of deserializing scalar values.

This means the following happened:
`Query input (as String)` -> deserialized (using scalars) to `Map` by `graphql-java` -> Jackson serializes the value again, without using scalars, and deserializes to a `Class`, again without using scalars.

In most cases, this happened to work, because Jackson could "accidentally" successfully serialize/deserialize to the same value. In some cases, such as more complex scalars (e.g. data/time), this didn't result in the correct value.
Even though it worked "most of the time", it has always been fundamentally incorrect.

Changes in this PR
----
In this PR a custom "converter" is added. Note that the converter only maps a `Map` to a Java type, it doesn't do any deserialization. The deserialization is done by `graphql-java`.
It accounts for both Java and Kotlin classes because they are instantiated differently through reflection.
We already had good test coverage for `@InputArgument` and all tests are passing, making this mostly a change that users won't notice, except in the special cases that previously didn't work.

Fixes: https://github.com/Netflix/dgs-framework/issues/70

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
@@ -16,14 +16,13 @@
 
 package com.netflix.graphql.dgs.internal
 
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.isKotlinClass
 import com.netflix.graphql.dgs.DgsDataFetchingEnvironment
 import com.netflix.graphql.dgs.InputArgument
 import com.netflix.graphql.dgs.context.DgsContext
 import com.netflix.graphql.dgs.exceptions.DgsInvalidInputArgumentException
 import com.netflix.graphql.dgs.exceptions.DgsMissingCookieException
 import graphql.schema.DataFetchingEnvironment
-import graphql.schema.GraphQLScalarType
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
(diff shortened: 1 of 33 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -61,14 +58,37 @@ class DataFetcherInvoker(
         method.parameters.asSequence().filter { it.type != Continuation::class.java }.forEachIndexed { idx, parameter ->
 
             when {
-                parameter.isAnnotationPresent(InputArgument::class.java) -> args.add(processInputArgument(parameter, idx))
-                parameter.isAnnotationPresent(RequestHeader::class.java) -> args.add(processRequestHeader(environment, parameter, idx))
-                parameter.isAnnotationPresent(RequestParam::class.java) -> args.add(processRequestArgument(environment, parameter, idx))
-                parameter.isAnnotationPresent(CookieValue::class.java) -> args.add(processCookieValueArgument(environment, parameter, idx))
+                parameter.isAnnotationPresent(InputArgument::class.java) -> args.add(
```
berngp, line 61: +1 to the col formatting ;)

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package com.netflix.graphql.dgs.internal
+
+import com.fasterxml.jackson.module.kotlin.isKotlinClass
+import kotlin.reflect.KClass
+import kotlin.reflect.full.primaryConstructor
+import kotlin.reflect.jvm.jvmErasure
+
+class InputObjectMapper {
```
berngp, line 24: `object` instead of class.

berngp (approved):
Changes look good to me. I'll do another pass but I think we should merge and have an RC.

PR #556, opened 2021-08-22
chore: add @JsonCreator to TypedGraphQLError

hantsy:
In my sample projects, when writing testing codes, I want to deserialize the **errors** result to *type-safe* `TypedGraphQLError`, it does not work.

I have to add a `@JsonCreator` to archive this purpose, eg.  [spring-graphql-sample/dgs-kotlin MutationTests](https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-kotlin/src/test/kotlin/com/example/demo/MutationTests.kt#L62) and [verification clause](https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-kotlin/src/test/kotlin/com/example/demo/MutationTests.kt#L73).


berngp (approved):
Thanks for the contribution!

hantsy:
@berngp If possible to enable the java compiler `-parameters` arguments **globally** and add Jackson parameter names module, thus we can erase all `@JsonProperty` annotations in the constructors.

berngp:
@hantsy, was already thinking of adding the `-parameters` flag already for a different reason. Thanks for pointing it out here as well.

PR #562, opened 2021-08-23
Avoid unnecessarily calculating the stacktrace when callint ExecutionStrategyInstrumentationContext

berngp:
Since DGS is complied via Kotlin, the bytecode will include a non-null
check on arguments which type is _none nullable_. In the case of `ExecutionStrategyInstrumentationContext`
this meant that on every call of any _resolver_ there will be a check of
nullability on the _throwable_. Although the exception was ignored it
incurred a heavy cost since calculating the exception includes
calculating the stack trace.

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
@@ -42,6 +42,14 @@ class DgsGraphQLMetricsInstrumentation(
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(DgsGraphQLMetricsInstrumentation::class.java)
+
+        private object DefaultExecutionStrategyInstrumentationContext : ExecutionStrategyInstrumentationContext {
+            override fun onDispatched(result: CompletableFuture<ExecutionResult>) {
+            }
+
+            override fun onCompleted(result: ExecutionResult, t: Throwable?) {
+            }
+        }
     }
 
     override fun createState(): InstrumentationState {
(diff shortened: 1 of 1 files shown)


paulbakker (approved):

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -42,6 +42,14 @@ class DgsGraphQLMetricsInstrumentation(
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(DgsGraphQLMetricsInstrumentation::class.java)
+
+        private object DefaultExecutionStrategyInstrumentationContext : ExecutionStrategyInstrumentationContext {
+            override fun onDispatched(result: CompletableFuture<ExecutionResult>) {
+            }
+
+            override fun onCompleted(result: ExecutionResult, t: Throwable?) {
```
osi, line 50: @berngp the addition of `?` here appears to be the largest portion of the change? can you elaborate more on why the null check there effects performance?

osi: ah, I think I may have answered my own question. Reading the javadocs for this method,
```
    /**
     * This is invoked when the instrumentation step is fully completed
     *
     * @param result the result of the step (which may be null)
     * @param t      this exception will be non null if an exception was thrown during the step
     */
    void onCompleted(T result, Throwable t);
```

.. if this was invoked in the non-error case, the Kotlin intrinsic `checkNotNullParameter` kicked in,

```
    public static void checkNotNullParameter(Object value, String paramName) {
        if (value == null) {
            throwParameterIsNullNPE(paramName);
        }
    }
```

and _that_ code will throw a NPE.

given that `result` is also nullable per the javadocs, shouldn't that also have a `?` ?

berngp: Exactly @osi. This was causing the computation on an exception every time we were instrumenting the execution. The exception went under the radar since there is no check on the return value of such _future_ since its treated as a _side effect_

berngp: > given that result is also nullable per the javadocs, shouldn't that also have a ? ?
I didn't observed any other unnecessary exceptions being created after the change but I'll verify. Thanks for mentioning it.

berngp: @osi https://github.com/Netflix/dgs-framework/pull/567

PR #567, opened 2021-08-24
Prevent calculating the stacktrace when calling ExecutionStrategyInstrumentationContext (rare case)

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

This is a followup on #562. 

We are describing the `result: ExecutionResult` parameter as nullable.
This will prevent an unnecessary _null check_ on the `result` parameter in case the value is `null`.
If we don't do this the `NullPointerException` will go unnoticed but will increase the latency in the request since a stacktrace 
will be computed. Arguably this will be a rare case so the latency impact will not be as drastic as the change that #562 addresses.


osi (approved):

berngp:
Thanks for pointing this out @osi !

PR #568, opened 2021-08-24
Add the @connection directive automatically when pagination module is used.

srinivasankavitha:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
This PR adds the connection directive automatically instead of having the user declare the directive in the SDL. This makes the feature easier to use.

### DIFF
--- a/graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
+++ b/graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
@@ -18,6 +18,7 @@ package com.netflix.graphql.dgs.pagination
 
 import com.netflix.graphql.dgs.DgsComponent
 import com.netflix.graphql.dgs.DgsTypeDefinitionRegistry
+import graphql.introspection.Introspection
 import graphql.language.*
 import graphql.schema.idl.TypeDefinitionRegistry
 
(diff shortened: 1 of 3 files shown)


graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
```diff
@@ -28,8 +29,13 @@ class DgsPaginationTypeDefinitionRegistry {
     fun registry(schemaRegistry: TypeDefinitionRegistry): TypeDefinitionRegistry {
         val definitions = schemaRegistry.types()
         val connectionTypes = parseConnectionDirective(definitions.values.toMutableList())
+
         val typeDefinitionRegistry = TypeDefinitionRegistry()
         typeDefinitionRegistry.addAll(connectionTypes)
+        val directive = DirectiveDefinition.newDirectiveDefinition().name("connection")
+            .directiveLocation(DirectiveLocation.newDirectiveLocation().name(Introspection.DirectiveLocation.OBJECT.name).build()).build()
```
paulbakker, line 36: Should this first check if it was already added manually to the schema?

graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
```diff
@@ -42,7 +48,7 @@ class DgsPaginationTypeDefinitionRegistry {
                 definitions.add(createEdge(it.name))
             }
 
-        if (types.find { it.name == "PageInfo" } == null) {
+        if (types.find { it.hasDirective("connection") } != null && types.find { it.name == "PageInfo" } == null) {
```
paulbakker, line 51: You can use `types.any(...)` instead of `.find`. It's the same, but returns a boolean, so it's a bit more readable.

graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
srinivasankavitha: Added the check.

graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
srinivasankavitha: done.

paulbakker (approved):

PR #603, opened 2021-09-03
Fix nullable enums in input arugments

jbaxleyiii:
Pull request checklist
----

  first
- [?] Make sure to have sufficient test cases

Pull Request type
----

- [x] Bugfix

Changes in this PR
----
Nullable enums as input arguments were broken because DGS tried to cast
null to be a value of the enum. With this commit, we check the null
status of the input argument or field (on an input object) before trying
to convert it to an enum

Also 👋🏼 @paulbakker @srinivasankavitha I hope y'all are doing well!


srinivasankavitha (approved):
Thanks got the PR! Great to hear from you :)

PR #604, opened 2021-09-04
Fix serialization of List of Enums as Input Arguments

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

When the target type of an input argument is a list of enumerations we need to transform the
list item to the enum type. Previously the value was left as a string causing unexpected errors.

As part of this change we are moving all the support classes for the  Input Argument serialization tests:

* If Java, to `com/netflix/graphql/dgs/internal/java/test/` and the name will have a `J` prefix. e.g. `JGreetingType`.
* If Kotlin, to `com/netflix/graphql/dgs/internal/kotlin/test/inputobjects.kt` and the name will have a 'K'  prefix. e.g. `KGreeingType`

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
@@ -54,22 +54,6 @@ object InputObjectMapper {
         return targetClass.primaryConstructor!!.call(*inputValues.toTypedArray())
     }
 
-    private fun convertList(input: List<*>, nestedTarget: KClass<*>): List<*> {
-        return input.map { listItem ->
-            if (listItem is Map<*, *>) {
-                if (nestedTarget.java == Object::class.java || nestedTarget == Any::class) {
-                    listItem
-                } else if (nestedTarget.java.isKotlinClass()) {
-                    mapToKotlinObject(listItem as Map<String, *>, nestedTarget)
-                } else {
-                    mapToJavaObject(listItem as Map<String, *>, nestedTarget.java)
-                }
-            } else {
-                listItem
-            }
-        }
-    }
-
     fun <T> mapToJavaObject(inputMap: Map<String, *>, targetClass: Class<T>): T {
         if (targetClass == Object::class.java) {
             return inputMap as T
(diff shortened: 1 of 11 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
```diff
@@ -1436,39 +1455,47 @@ internal class InputArgumentTest {
 
         val fetcher = object : Any() {
             @DgsData(parentType = "Query", field = "hello")
-            fun someFetcher(@InputArgument someInput: InputMessage): String {
-                return "Hello, this is a ${someInput.type} greeting"
+            fun hello(@InputArgument input: KInputMessage): String {
```
berngp, line 1458: In the past we were just asserting the representation of the output as a string. We need to assert that the collections actually contain the proper type.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
```diff
@@ -104,4 +88,24 @@ object InputObjectMapper {
 
         return instance
     }
+
+    private fun convertList(input: List<*>, nestedTarget: KClass<*>): List<*> {
+        return input.filterNotNull().map { listItem ->
+            if (nestedTarget.java.isAssignableFrom(listItem::class.java)) {
+                listItem
+            } else if (nestedTarget.java.isEnum) {
```
berngp, line 96: We need to ask if the target class is an enum, and if such we need to fetch the enum value. If we don't do this the resulting collection/list will have strings instead of the enums.

IvanDanilau:
I still have this issue. It works if List of enums is property of input type, but is a case of usage:
`@DgsQuery(field = QUERY.TestEnum)
  public boolean testEnum(@InputArgument(collectionType = TestEnum.class) List<TestEnum> enums) {
    return true;
  }` 
it still converted to the string type. What's I do wrong?

berngp:
Thanks @IvanDanilau , followup via https://github.com/Netflix/dgs-framework/issues/644.

PR #628, opened 2021-09-12
add query params to context

nf-super-egg:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

Adds query params to the DGS context


paulbakker:
I'm not sure about this. Query params aren't super common to use, and can easily be accessed from the WebRequest/ServerRequest. 

Also we have the `@RequestParam` for data fetchers already.
If add request parameters, why not also cookie values, and everything else that's on WebRequest/ServerRequest?

Is there anything that you _can't_ do without this change?

nf-super-egg:
@paulbakker yeah I guess you're right. I figured since we had http headers in there, it'd be nice to have query params as well, but there is no reason it can't be pulled from the WebRequest. 👍

PR #632, opened 2021-09-13
Move spring-security to Compile Only since its not a requirement.

### LINKED ISSUE #631: bug:  Make Spring Security dependency optional in the graphql-dgs
Tried to create a webflux example project, I did not add Spring Security in the dependencies, and but I found the Spring Security in the dependency tree. 

I found the `spring-security-core` dep in the `graphql-dgs` module,  https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/build.gradle.kts#L33

berngp:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Changes in this PR
----

Move spring-security as an optional module. Fixes #631

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt
@@ -22,6 +22,7 @@ import graphql.execution.DataFetcherExceptionHandlerParameters
 import graphql.execution.DataFetcherExceptionHandlerResult
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.util.ClassUtils
 
 /**
  * Default DataFetcherExceptionHandler used by the framework, can be replaced with a custom implementation.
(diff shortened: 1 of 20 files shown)


srinivasankavitha (approved):

berngp:
We should move [this](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt#L38-L43), ref below, to happen as part of a static init block/object or refactor the class to support a chain of handlers and then make the spring-security handler available _conditionally_ on the `org.springframework.security.access.AccessDeniedException` class.

```
        val springSecurityAvailable = try {
            Class.forName("org.springframework.security.access.AccessDeniedException")
            true
        } catch (ex: ClassNotFoundException) {
            false
        }
```

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DefaultDataFetcherExceptionHandler.kt
```diff
@@ -62,6 +56,21 @@ class DefaultDataFetcherExceptionHandler : DataFetcherExceptionHandler {
     }
 
     companion object {
+
         private val logger: Logger = LoggerFactory.getLogger(DefaultDataFetcherExceptionHandler::class.java)
+
+        private val springSecurityAvailable: Boolean by lazy {
```
berngp, line 62: @paulbakker @srinivasankavitha I'm updating how we calculate this as well such that the assertion that if SpringSecurity is available only happens ounce.

paulbakker: yeah that's a good improvement

paulbakker (approved):

PR #638, opened 2021-09-15
Prevent the creation of Loggers in the Hotpath.

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

This PR moves all loggers to be available through a `companion object`.
This will prevent instances of loggers to be created as part of the _request execution_/_hotpath_.
Although this was initially identified in the `DataFetcherInvoker`, via a JFR analysis, we are also
favoring consistency and apply the same pattern across all components.

From the image below the time associated with creating the logger per DataFetcherInvoker is visible on the right.

image

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -62,9 +62,6 @@ import org.springframework.web.multipart.MultipartFile
 @RestController
 open class DgsRestController(open val dgsQueryExecutor: DgsQueryExecutor) {
 
-    open val logger: Logger = LoggerFactory.getLogger(DgsRestController::class.java)
-    private val mapper = jacksonObjectMapper()
-
     // The @ConfigurationProperties bean name is <prefix>-<fqn>
     @RequestMapping(
         "#{@'dgs.graphql-com.netflix.graphql.dgs.webmvc.autoconfigure.DgsWebMvcConfigurationProperties'.path}",
@@ -196,5 +193,10 @@ open class DgsRestController(open val dgsQueryExecutor: DgsQueryExecutor) {
         }
 
         return ResponseEntity.ok(result)
+    }
+
+    companion object {
+        private val logger: Logger = LoggerFactory.getLogger(DgsRestController::class.java)
+        private val mapper = jacksonObjectMapper()
     }
 }
--- a/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
+++ b/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
@@ -25,6 +25,7 @@ import graphql.validation.ValidationError
 import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
+import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
@@ -41,7 +42,6 @@ import java.util.*
  */
 @RestController
 open class DgsSSESubscriptionHandler(open val dgsQueryExecutor: DgsQueryExecutor) {
-    open val logger = LoggerFactory.getLogger(DgsSSESubscriptionHandler::class.java)
     private val mapper = jacksonObjectMapper()
 
     @RequestMapping("/subscriptions", produces = ["text/event-stream"])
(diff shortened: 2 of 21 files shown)


graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -41,7 +42,6 @@ import java.util.*
  */
 @RestController
 open class DgsSSESubscriptionHandler(open val dgsQueryExecutor: DgsQueryExecutor) {
-    open val logger = LoggerFactory.getLogger(DgsSSESubscriptionHandler::class.java)
```
berngp, line 44: @lucatk, any chance you recall any reason to make the logger `open`?

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -62,9 +62,6 @@ import org.springframework.web.multipart.MultipartFile
 @RestController
 open class DgsRestController(open val dgsQueryExecutor: DgsQueryExecutor) {
 
-    open val logger: Logger = LoggerFactory.getLogger(DgsRestController::class.java)
```
berngp, line 65: @bono007, any chance you recall any reason to make the logger `open`?

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
lucatk: to allow CGLIB proxying of DGS' Bean classes, see #428.
I've not explored "companion objects" yet, so I can't verify whether or not this will break CGLIB proxying if you keep the companion object private.

berngp: This should not affect CGLIB proxies since the logger will not be an attribute of the reference but the class. You can see this all over the place in Java where loggers are private and static.

paulbakker:
Nice find!

paulbakker (approved):

srinivasankavitha (approved):

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
bono007: @berngp sorry just now getting back to this comment (been a crazy 24 hrs). I do not remember why. Its possible I had a reason and forgot by now OR I copy/pasta the logger from elsewhere.

bono007:
@berngp the above analysis was done w/ JFR? Anything special to get that going - or just follow their docs? I have been interested in looking into JFR more and the above reminds me about it. Thx for that.

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
berngp: No worries. Hope all is well now.

berngp:
@bono007 the specific graph above is generated by an internal tool we have. That said, I use JMC to do some further analysis on generated JFR files for jvms that I run locally or remote instances in test. There is a decent guide [here](https://docs.oracle.com/en/java/java-components/jdk-mission-control/8/user-guide/). Note that in Mac OS you will most likely need to start JMC from the command line. Tip, use JDK 11+.

JMC can attach to running instance and do _Flight Recording_ or you can use `jmcd` to create a jfr and then open it in JMC.

You could attach to a running JVM, `jps -f` to see the instances running. 
```
jcmd <pid> JFR.start duration=60s filename=myrecording.jfr settings=profile
```

You could also have the JVM generate a JFR at startup time if you are interested on how the jvm looks like when its booting up.
```
-XX:+FlightRecorder -XX:FlightRecorderOptions=stackdepth=1024 -XX:StartFlightRecording=duration=30s,filename=/some/path/to/mystartuprecording.jfr
```

My 2¢, to analyze the request path you want...
1. Have a jvm that has been already running and in a steady state.
2. Be able to generate the request you intend to analyze.
3. Be ready to capture a _flight recording_ which time window is sufficient to capture that particular request and doesn't extend too much. This will avoid noise in your JFR.
4. Coordinate when you start sampling/capturing for you JFR and send the request(s). Attempting to minimize noise in the jvm by stopping other things that are not related with that particular request you intend to analyze. If possible isolate the jvm and have only one request go through or have requests go trough in sequence where you wait for the previous to complete before you generate a new one.

PR #639, opened 2021-09-15
Fix bug in handling Optional Enum input arguments

paulbakker:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

`Optional<Enum>` didn't work anymore, the value would become `Optional<String>`.

### DIFF
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
@@ -1396,6 +1396,88 @@ internal class InputArgumentTest {
     }
 
     @Test
+    fun `Kotlin Optional enum @InputArgument`() {
+        val schema = """
+            type Query {
+                hello(type:GreetingType): String
+            }
+            
+            enum GreetingType {          
+                FRIENDLY
+                POLITE
+            }
+        """.trimIndent()
+
+        val fetcher = object : Any() {
+            @DgsData(parentType = "Query", field = "hello")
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
```diff
@@ -1395,6 +1400,93 @@ internal class InputArgumentTest {
...
+                FRIENDLY
+                POLITE
+            }
+        """.trimIndent()
+
+        val fetcher = object : Any() {
+            @DgsData(parentType = "Query", field = "hello")
+            fun someFetcher(@InputArgument type: Optional<KGreetingType>): String {
+                val typeValue: Any = type.get()
+                if (typeValue !is KGreetingType) {
```
berngp, line 1420: Why not just use a fluent assertJ check? 
```
assertThat(type).isNotEmpty.get().isInstanceOf(KGreetingType::class.java)
```

Will result in
```
Expecting:
  <"B">
to be an instance of:
  <com.netflix.graphql.dgs.internal.kotlin.test.KGreetingType >
but was instance of:
  <java.lang.String>
```

berngp (approved):

srinivasankavitha (approved):

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
paulbakker: yep, good idea. Updated.

PR #645, opened 2021-09-20
Support List of enums ar ArgumentInput types.

### LINKED ISSUE #644: bug: If the input is a list of enums, the deserialized list passed to the fetcher still contains strings.
If the input is a list of enums, the deserialized list passed to the fetcher still contains strings, e.g.:

```
Query {
  lookup(types: [MyTypeEnum!]): [Result]
}

enum MyTypeEnum {
  TYPE1,
  TYPE2
}
```

This will fail:
```
query TestQuery {
  lookup(types: [TYPE1])
}
```

However if you nest the enum inside another input object the deserialization does work correctly:
```
query {
  lookup(types: MyLookupType): [Result]
}

input MyLookupType {
  types: [MyTypeEnum!]!
}

enum MyTypeEnum {
  TYPE1,
  TYPE2
}
```

this will work, nested list of enums will be deserialized properly when calling the fetcher method:
```
Query TestQuery {
  lookup(types: { types: [TYPE1]})
}
```

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

This commit allows a list of enums to be used directly as an ArgumentInput.

Fixes #644

Out of Scope
----

This PR requires the developers to define a _Collection Type_, in this case the type of the Enum.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
@@ -190,22 +190,23 @@ class DataFetcherInvoker(
         val collectionType = annotation.collectionType.java
         val parameterValue: Any? = environment.getArgument(parameterName)
 
-        val convertValue: Any? = if (parameterValue is List<*> && collectionType != Object::class.java) {
-            try {
-                // Return a list of elements that are converted to their collection type, e.e.g. List<Person>, List<String> etc.
-                parameterValue.map { item -> convertValue(item, parameter, collectionType) }.toList()
-            } catch (ex: Exception) {
-                throw DgsInvalidInputArgumentException(
-                    "Specified type '$collectionType' is invalid for $parameterName.",
-                    ex
-                )
+        val convertValue: Any? =
+            if (parameterValue is List<*> && collectionType != Object::class.java) {
+                try {
+                    // Return a list of elements that are converted to their collection type, e.e.g. List<Person>, List<String> etc.
+                    parameterValue.map { item -> convertValue(item, parameter, collectionType) }.toList()
+                } catch (ex: Exception) {
+                    throw DgsInvalidInputArgumentException(
+                        "Specified type '$collectionType' is invalid for $parameterName.",
+                        ex
+                    )
+                }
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
```diff
@@ -1467,6 +1457,51 @@ internal class InputArgumentTest {
         verify { applicationContextMock.getBeansWithAnnotation(DgsComponent::class.java) }
     }
```
srinivasankavitha, line 1459: Do we already have a test case covering the nested enum type use case?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -239,8 +240,15 @@ class DataFetcherInvoker(
             } else {
                 InputObjectMapper.mapToJavaObject(parameterValue as Map<String, *>, targetType)
             }
-        } else if (parameter.type.isEnum && parameterValue !== null) {
-            (parameter.type.enumConstants as Array<Enum<*>>).find { it.name == parameterValue }
+        } else if ((parameter.type.isEnum || collectionType?.isEnum ?: false) && parameterValue != null) {
```
srinivasankavitha, line 243: What does this mean? `collectionType?.isEnum ?: false` Should this just be `collectionType?.isEnum`

berngp: It should just be `collectionType?.isEnum == true`, the Kotlin compiler doesn't like just `collectionType?.isEnum` 🤷

srinivasankavitha: Got it, I see.

srinivasankavitha (approved):

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
berngp: We have plenty of tests that include _enums_ as part of other complex types but not directly as an input argument.

Issue #437, opened 2021-06-25
feature: Add Multipart support to GraphQLClient


hantsy:
I can not find  an approach to use `GraphQLClient` to send a query which includes a `Upload`. 
or let me know how to do it?

paulbakker:
Example: https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-example-java/src/test/java/FileUploadMutationTest.java

Also, for questions, please create a question on the discussion forum instead of creating an issue.

hantsy:
@paulbakker I mean  [Java GraphQL Client](https://netflix.github.io/dgs/advanced/java-client/) not `DgsQueryExecutor `.

hantsy:
@paulbakker And I browsed the codes of `DgsQueryExecutor`, it delegates the query execution to the  `GrapQL` directly, not via `GraphQLClient`.

PR #658, opened 2021-09-25
refactored dgs client, and new Webclient based client.

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Added `WebClientGraphQLClient` as a convenience implementation. In this implementation the user is no longer responsible for making the HTTP request based on the `RequestExecutor` interface.
This new client should probably be the default for most use cases, unless you _really_ want to use a different HTTP client.

Because the existing `GraphQLClient` and `MonoGraphQLClient` interfaces had the `RequestExecutor` as an argument in each method, I deprecated those to better align with integrated clients like the new `WebClientGraphQLClient`.
The `DefaultGraphQLClient` is now also deprecated in favor of `CustomGraphQLClient` and `CustomReactiveGraphQLClient`, which both take `RequestExecutor` as a constructor argument.
We'll keep the old implementation around for a while, but it should be a super easy upgrade for folks.

Alternatives considered
----

I considered keeping the existing interfaces and implementation, and just add the `WebClientGraphQLClient` as a completely new, unrelated, option. That seems confusing though, so I opted for deprecating the methods that were problematic.


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
```diff
@@ -92,3 +190,28 @@ fun interface MonoRequestExecutor {
  */
 class GraphQLClientException(statusCode: Int, url: String, response: String, request: String) :
     RuntimeException("GraphQL server $url responded with status code $statusCode: '$response'. The request sent to the server was \n$request")
+
+internal object GraphQLClients {
+    internal val objectMapper: ObjectMapper = try {
```
berngp, line 195: We already bring Spring in the client and it has the ClassUtils class with `public static boolean isPresent(String className, @Nullable ClassLoader classLoader) `.

srinivasankavitha:
I find the `Custom` part of the CustomGraphQLClient and CustomReactiveGraphQLClient misleading.  If it's mainly to avoid clashing with the existing deprecated clients, how about BlockingGraphQLClient and ReactiveGraphQLClient for the names?

srinivasankavitha (changes requested):
I find the Custom part of the CustomGraphQLClient and CustomReactiveGraphQLClient not descriptive enough. If it's mainly to avoid clashing with the existing deprecated clients, how about BlockingGraphQLClient and ReactiveGraphQLClient for the names?

hantsy:
Although this new approach removes the RequestExecutor, but I think it still need to leave some room to customize it.

hantsy:
And consider adding a MultpartFile/Upload method in the new client, see #437

hantsy:
[GraphQL Java Kickstart WebClient](https://github.com/graphql-java-kickstart/graphql-spring-webclient) includes extra simple config for OAuth 2 etc.

paulbakker:
@srinivasankavitha @hantsy The "custom" part in the name comes from the fact that the user has to provide a `RequestExecutor`, so you have full control over the HTTP client and processing the HTTP request/response.

I thought about `BlockingGraphQLClient` and `ReactiveGraphQLClient`, but `ReactiveGraphQLClient` is already there for the subscription client stuff, so we can't use that.
I think we should push folks to use the new WebClient based implementation, and only use the "custom" ones if needed, maybe having "custom" in the name helps indicate that there are easier options?

paulbakker:
> [GraphQL Java Kickstart/GraphQL Spring WebClient](https://github.com/graphql-java-kickstart/graphql-spring-webclient) includes extra simple config for OAuth 2 etc.

I think that means we still need some way to add request headers. What about adding an optional callback as a constructor argument which lets you process the headers?

paulbakker:
> And consider adding a MultpartFile/Upload method in the new client, see #437

I agree, but out of scope for this change.

srinivasankavitha:
> @srinivasankavitha @hantsy The "custom" part in the name comes from the fact that the user has to provide a `RequestExecutor`, so you have full control over the HTTP client and processing the HTTP request/response.
> 
> I thought about `BlockingGraphQLClient` and `ReactiveGraphQLClient`, but `ReactiveGraphQLClient` is already there for the subscription client stuff, so we can't use that.
> I think we should push folks to use the new WebClient based implementation, and only use the "custom" ones if needed, maybe having "custom" in the name helps indicate that there are easier options?

Ok, I see. If the intention is to use the WebClient based implementation for the most part, the `custom` in the naming makes sense for the other cases.  Something we can just clarify in the docs.

srinivasankavitha (approved):

berngp (approved):

paulbakker:
Added some migration docs: https://github.com/Netflix/dgs-framework/releases/edit/untagged-0310b7a342ea92cecd98

Issue #640, opened 2021-09-16
feature: Add support for graphql-java-extended-validation


setchy:
## Describe the Feature Request

Just like DGS supports `graphql-java-extended-scalars`, it would be nice if there was support for the suite of schema validation directives provided by [graphql-java-extended-validation](https://github.com/graphql-java/graphql-java-extended-validation)

## Describe Preferred Solution

Add support for [graphql-java-extended-validation](https://github.com/graphql-java/graphql-java-extended-validation)

hantsy:
Note: Dgs used an old GraphQL Java 16.x, when using this `graphql-java-extended-validation`,  you have to use the 16.0 version and an older Hibernate validator, check [here](https://github.com/hantsy/spring-graphql-sample/blob/java17/dgs-codegen/build.gradle#L26-L29). 

I hope Dgs will upgrade to GraphQL Java 17 in the next minor version, there is a PR for it, https://github.com/Netflix/dgs-framework/pull/518

paulbakker:
We're just talking about `graphql-java-extended-validation` and want to look into it, so this is very timely. We'll follow up once we have looked into it a bit more.

paulbakker:
> Note: Dgs used an old GraphQL Java 16.x, when using this `graphql-java-extended-validation`, you have to use the 16.0 version and an older Hibernate validator, check [here](https://github.com/hantsy/spring-graphql-sample/blob/java17/dgs-codegen/build.gradle#L26-L29).
> 
> I hope Dgs will upgrade to GraphQL Java 17 in the next minor version, there is a PR for it, #518

We're still blocked by the [federation-lib](https://github.com/apollographql/federation-jvm/pull/122) release to update to 17. We raised the issue again with folks involved, so hopefully we can move on soon.

sbilello:
For the interested readers:

I integrated `graphql-java-extended-validation` library in our project and the steps to make it work are 2:

I forced  `hibernate-validator` to solve a run time problem:

In `build.gradle`

```gradle
configurations.all {
	exclude group: 'com.vaadin.external.google' , module: 'android-json'
	resolutionStrategy {
		force 'org.hibernate.validator:hibernate-validator:6.1.6.Final'
	}
}

```

I added a new Bean

```
@DgsComponent
public class GraphQLSchemaValidationConfiguration {

  @DgsRuntimeWiring
  public RuntimeWiring.Builder addValidationDirectives(RuntimeWiring.Builder builder) {
[...]

PR #659, opened 2021-09-25
Add graphql-dgs-extended-validation module

hantsy:
address #640


setchy:
Awesome @hantsy! 

Thinking out loud, I suspect a new `Advanced` documentation entry may be required, too - https://github.com/Netflix/dgs as well

hantsy:
@setchy I may add such a section later. But I am a none English speaker, writing great English docs is more difficult than coding.

srinivasankavitha:
Thanks for the PR @hantsy. We were just recently looking at adding this. Will review and test it out next week.

berngp (commented):
Thanks @hantsy, did a quick review and looks great. Will take a deeper dive tomorrow.

berngp (approved):

graphql-dgs-extended-validation/src/test/kotlin/com/netflix/graphql/dgs/autoconfig/BeanValidationSizeSmokeTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    fun createPostWithInvalidInput() {
+        val query = "mutation newPost(\$input: CreatePostInput!){ createPost(createPostInput: \$input) }"
+        var variables = mapOf(
+            "input" to mapOf(
+                "title" to "test",
+                "content" to "test content"
+            )
+        )
+        val executionResult = queryExecutor.execute(query, variables)
```
srinivasankavitha, line 46: Is there a specific error that we can assert and check for?

graphql-dgs-platform/build.gradle.kts
```diff
@@ -51,6 +51,9 @@ dependencies {
         api("com.graphql-java:graphql-java-extended-scalars") {
             version { require(Versions.GRAPHQL_JAVA_EXTENDED_SCALARS) }
         }
```
srinivasankavitha, line 53: I'm wondering of this should be added by default, or just be available as an opt-in by the user.

srinivasankavitha:
I see this error when I try it with a spring boot app: 
```
 Exception while executing data fetcher for /createPost: HV000183: Unable to initialize 'jakarta.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead

jakarta.validation.ValidationException: HV000183: Unable to initialize 'jakarta.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.buildExpressionFactory(ResourceBundleMessageInterpolator.java:217) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.<init>(ResourceBundleMessageInterpolator.java:83) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.<init>(ResourceBundleMessageInterpolator.java:48) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
```

You can reproduce it by adding the schema in your test to the `graphql-dgs-example-shared` module and running the example app. Running the createPost query will produce the error. Looks like the dependencies are not getting pulled in transitively.

srinivasankavitha (changes requested):

graphql-dgs-extended-validation/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsExtendedValidationAutoConfiguration.kt
```diff
@@ -0,0 +1,65 @@
...
+import graphql.validation.rules.ValidationRules
+import graphql.validation.schemawiring.ValidationSchemaWiring
+import org.springframework.beans.factory.ObjectProvider
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@ConditionalOnClass(graphql.validation.rules.ValidationRules::class)
+@ConditionalOnProperty(
```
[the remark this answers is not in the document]
setchy: I tend to agree, but would that make it inconsistent with the how extended-scalars or micrometer (any others???) are configured?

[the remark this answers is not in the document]
hantsy: This is a general approach to enable features in Spring Boot. For example, in the application I need some feature, but  in some tests for some specific reason it can be disabled with a property on tests.

graphql-dgs-extended-validation/src/test/kotlin/com/netflix/graphql/dgs/autoconfig/BeanValidationSizeSmokeTest.kt
hantsy: Added detailed error message assert.

hantsy:
> 
> 
> I see this error when I try it with a spring boot app:
> 
> ```
>  Exception while executing data fetcher for /createPost: HV000183: Unable to initialize 'jakarta.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
> 
> jakarta.validation.ValidationException: HV000183: Unable to initialize 'jakarta.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
> 	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.buildExpressionFactory(ResourceBundleMessageInterpolator.java:217) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
> 	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.<init>(ResourceBundleMessageInterpolator.java:83) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
> 	at org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator.<init>(ResourceBundleMessageInterpolator.java:48) ~[hibernate-validator-7.0.1.Final.jar:7.0.1.Final]
> ```
> 
> You can reproduce it by adding the schema in your test to the `graphql-dgs-example-shared` module and running the example app. Running the createPost query will produce the error. Looks like the dependencies are not getting pulled in transitively.

Currently I only tried 16.0 in my Dgs example(4.8.3).  The hibernate validator added required deps, no need extra settings. 

@berngp But from this error, I found another issue,  and checked Hibernate Validator 7.x doc, found it was updated to Jakarta EE 9 API level: https://in.relation.to/2021/01/06/hibernate-validator-700-62-final-released/, but Spring is still focus on Java EE 8/Jakarta EE 8, does not support the new namespace.

hantsy:
There is an issue reported the problem when using GraphQL 17 and Spring Boot 2.5.5 together, https://github.com/graphql-java/graphql-java-extended-validation/issues/52

graphql-dgs-extended-validation/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsExtendedValidationAutoConfiguration.kt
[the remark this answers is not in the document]
srinivasankavitha: Yes, I agree that this is a general approach. However, for this particular feature, I didn't see a use case for having the module be present but disabled via a property. That said, I am fine with keeping it as is. 
@setchy - as you also pointed out, we do have the properties for those features that are already in separate modules. So I agree with your point about consistency as well.

srinivasankavitha:
> There is an issue reported the problem when using GraphQL 17 and Spring Boot 2.5.5 together, [graphql-java/graphql-java-extended-validation#52](https://github.com/graphql-java/graphql-java-extended-validation/issues/52)

I think we'll defer this PR until there is a resolution for the above mentioned issue, agree? We just switched the framework to graphql-java-17 and released already.

berngp (changes requested):
This changes works well with the Spring Boot 2.3.x versions that DGS is currently using,
but we will soon be moving to Spring Boot 2.5. There is a blocker on `graphql-java-extended-validation` when used along with Spring Boot 2.5, see issue below. For this reason we will have to wait for this PR.

https://github.com/graphql-java/graphql-java-extended-validation/issues/52

paulbakker:
Thanks for the work @hantsy. I was pretty excited to see this! 

Too bad about the Hibernate Validator issue that seems to be blocking for now. Hopefully that can be resolved in library.

paulbakker:
Thanks to @berngp I found this thread on the Spring Framework that clarifies why this is an issue: https://github.com/spring-projects/spring-framework/issues/25354#issuecomment-812572313

hantsy:
@berngp  Current Spring only recognizes the Jakarta EE 8  API and ignore new *Jakarta EE 9 API* at all, so I think if using this validation extension in a Spring Boot 2.5 application, it could work, but there are some Jakarta EE 8 and 9 API coexisted in the final jar.

hantsy:
@srinivasankavitha Sent a PR on the upstream project, https://github.com/graphql-java/graphql-java-extended-validation/pull/53, the default source codes are switched back to Jakarta EE 8, and I used a `jakarta` classier for Jakarta EE 9 build. I would like a different artifactId for the jakarta build(some Jakarta EE projects used a `jakarta` postfix to differentiate from EE 8), but I can not set the `artifactId` dynamically in the maven publish plugin config.

hantsy:
@berngp  When adding all the deps (Jakarta EE 9) into build.gradle, the application starts up without any errors.  https://github.com/hantsy/spring-graphql-sample/blob/master/dgs-codegen/build.gradle#L26-L30

sbilello:
Is there a particular validation mechanism outside of this one adopted by Netflix?
I guess that this module is not used but I encountered the problem of upgrading my project and it looks like that it is a problem with its depth dependency hierarchy problem. Is there a different solution adopted or there is no POJO validation at all?

hantsy:
@sbilello you can use Bean Validation directly.

Generally, when using Dgs framework, if you build the models manually,  **Bean validation** is a good match. If you are using `codegen` plugin to generate models,  GraphQL specific validation directives are better. This is only my idea.

sbilello:
Yes, @hantsy but I didn't want to remove the graphql-java-extended-validation module.

hantsy:
@paulbakker There is no response from the upstream maintainers for weeks, if possible create a fork of *graphql-java-extended-validation* under Netflix? I think only need a Jakarta EE 8 version here.

paulbakker:
Forking would be a bit of a maintenance burden long-term, so I would really prefer to avoid that. I've asked on twitter who's maintaining the repo, hopefully we'll find someone who can merge/release.

paulbakker:
I spoke to Andi today, and he'll look into it after next week. They obviously have a lot on their plate, so let's just be patient. At least we now know the repo isn't abandoned.

hantsy:
@paulbakker Refreshed and update the version.

graphql-dgs-platform/build.gradle.kts
```diff
@@ -51,14 +51,12 @@ dependencies {
...
+        api("com.graphql-java:graphql-java-extended-validation") {
+            version { require(Versions.GRAPHQL_JAVA_EXTENDED_VALIDATION) }
+        }
         api("com.apollographql.federation:federation-graphql-java-support") {
             version { require(Versions.GRAPHQL_JAVA_FEDERATION) }
         }
-        api("com.graphql-java:graphql-java-extended-validation"){
-            // The version below will work with Jakarta EE 8
```
berngp, line 58: Why remove the comments?

berngp: Ok, they were added to the version file.

berngp: I don't see much value on keeping the versions in the version file. Ot is an unnecessary level of indirection. I'll address this in a different pr.

berngp (approved):

graphql-dgs-platform/build.gradle.kts
hantsy: comments moved to `Versions.kt`

PR #662, opened 2021-09-26
Update contributing doc

hantsy:
Update the some out-of-date content in the contributing doc.


hantsy:
Updated it according to the current available tasks, but not sure if it will be replaced with klint #630

berngp (approved):
Thanks, forgot to update the docs.

PR #676, opened 2021-10-05
Support custom deserializers in the GraphQLClient for deserializing the response.

srinivasankavitha:
Pull request checklist
----

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
This feature allows the user to set custom deserializers for processing the GraphQLResponse. We have had many requests to allow custom mappers to be injected since users would like to add modules for processing custom types such as DateTime etc. Rather than exposing the jackson object mapper in the interface, we expose a more generic abstraction. The user can now provide the deserializer as a lambda that the framework configures in the internal jackson mapper. 

The same mechanism will work for any custom scalars that implement Coercing<*, *>. The parseValue method can be set as the custom deserializers in this scenario.

Here is a code snippet from the tests showing the usage where the second argument represents the map of custom deserializers:
```
private val client = DefaultGraphQLClient(url,
        mapOf(
            OffsetDateTime::class.java to { t -> OffsetDateTime.parse(t.toString(), DateTimeFormatter.ISO_OFFSET_DATE_TIME) },
            Person::class.java to { t -> Person("Custom Person", 20)}
        )
    )
```
Alternatives considered
----
We considered allowing the user to inject their own  jackson object mapper in the GraphQLClient interface for customization
. Since this would couple the framework tightly with the  Jackson mapper, we opted to expose a more generic abstraction.

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
@@ -21,7 +21,7 @@ package com.netflix.graphql.dgs.client
  * The user is responsible for doing the actual HTTP request, making this pluggable with any HTTP client.
  * For a more convenient option, use [WebClientGraphQLClient] instead.
  */
-class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor) : GraphQLClient {
+class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor, private val customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) : GraphQLClient {
     override fun executeQuery(query: String): GraphQLResponse {
         return executeQuery(query, emptyMap(), null)
     }
@@ -40,6 +40,6 @@ class CustomGraphQLClient(private val url: String, private val requestExecutor: 
         )
 
         val response = requestExecutor.execute(url, GraphQLClients.defaultHeaders, serializedRequest)
-        return GraphQLClients.handleResponse(response, serializedRequest, url)
+        return GraphQLClients.handleResponse(response, serializedRequest, url, customDeserializer)
     }
 }
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
@@ -16,12 +16,14 @@
 
 package com.netflix.graphql.dgs.client
 
+import graphql.schema.Coercing
+
 /**
  * Blocking implementation of a GraphQL client.
  * The user is responsible for doing the actual HTTP request, making this pluggable with any HTTP client.
  * For a more convenient option, use [WebClientGraphQLClient] instead.
  */
-class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor, private val customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) : GraphQLClient {
+class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor, private val customDeserializer: List<Coercing<*, String>>? = null) : GraphQLClient {
     override fun executeQuery(query: String): GraphQLResponse {
         return executeQuery(query, emptyMap(), null)
     }
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomJsonDeserializer.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomJsonDeserializer.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.client
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+
+class CustomJsonDeserializer<T>(private val parse: (Any) -> Any?, vc: Class<*>) : StdDeserializer<T>(vc) {
+    override fun deserialize(jp: JsonParser, ctxt: DeserializationContext): T? {
+        val node: JsonNode = jp.codec.readTree(jp)
+        return parse(node.asText()) as T
+    }
+}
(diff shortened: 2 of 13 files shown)


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
```diff
@@ -21,7 +21,7 @@ package com.netflix.graphql.dgs.client
  * The user is responsible for doing the actual HTTP request, making this pluggable with any HTTP client.
  * For a more convenient option, use [WebClientGraphQLClient] instead.
  */
-class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor) : GraphQLClient {
+class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor, private val customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) : GraphQLClient {
```
paulbakker, line 24: This needs to be defined as a separate constructor, otherwise it will be a required argument for Java users.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClient.kt
```diff
@@ -67,7 +67,7 @@ interface GraphQLClient {
 
     companion object {
         @JvmStatic
-        fun createCustom(url: String, requestExecutor: RequestExecutor) = CustomGraphQLClient(url, requestExecutor)
+        fun createCustom(url: String, requestExecutor: RequestExecutor, customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) = CustomGraphQLClient(url, requestExecutor, customDeserializer)
```
paulbakker, line 70: same as with the constructor, you can't just add arguments. It would work for Kotlin, but not for Java

graphql-dgs-client/src/test/java/com/netflix/graphql/client/GraphQLResponseJavaTest.java
```diff
@@ -34,7 +34,7 @@
     RestTemplate restTemplate = new RestTemplate();
     MockRestServiceServer server = MockRestServiceServer.bindTo(restTemplate).build();
     String url = "http://localhost:8080/graphql";
-    DefaultGraphQLClient client = new DefaultGraphQLClient(url);
+    DefaultGraphQLClient client = new DefaultGraphQLClient(url, null);
```
paulbakker, line 37: this shouldn't be necessary. Don't change existing Java code, because that indicates it will be breaking for users.

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/WebClientGraphQLClientTest.kt
```diff
@@ -47,7 +47,7 @@ class WebClientGraphQLClientTest {
 
     @BeforeEach
     fun setup() {
-        client = MonoGraphQLClient.createWithWebClient(WebClient.create("http://localhost:$port/graphql"))
+        client = MonoGraphQLClient.createWithWebClient(WebClient.create("http://localhost:$port/graphql"), null)
```
paulbakker, line 50: This shouldn't be necessary

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomJsonDeserializer.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.netflix.graphql.dgs.client
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+
+class CustomJsonDeserializer<T>(private val parse: (Any) -> Any?, vc: Class<*>) : StdDeserializer<T>(vc) {
```
paulbakker, line 24: Needs javadoc

paulbakker (changes requested):

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/GraphQLResponseTest.kt
```diff
@@ -54,7 +55,13 @@ class GraphQLResponseTest {
     }
 
     private val url = "http://localhost:8080/graphql"
-    private val client = DefaultGraphQLClient(url)
+    private val client = DefaultGraphQLClient(
```
paulbakker, line 58: This should be a new test instead of adding it to existing tests.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
```diff
@@ -39,7 +39,7 @@ import reactor.core.publisher.Mono
  *    });
  */
 @Deprecated("This has been replaced by [CustomGraphQLClient], [CustomReactiveGraphQLClient] and [WebClientGraphQLClient]")
-class DefaultGraphQLClient(private val url: String) : GraphQLClient, MonoGraphQLClient {
+class DefaultGraphQLClient(private val url: String, private val customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) : GraphQLClient, MonoGraphQLClient {
```
paulbakker, line 42: DefaultGraphQLClient is already deprecated. I don't think we should make any changes to it.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
```diff
@@ -37,7 +38,7 @@ import org.slf4j.LoggerFactory
  * Representation of a GraphQL response, which may contain GraphQL errors.
  * This class gives convenient JSON parsing methods to get data out of the response.
  */
-data class GraphQLResponse(val json: String, val headers: Map<String, List<String>>) {
+data class GraphQLResponse(val json: String, val headers: Map<String, List<String>>, val customDeserializer: Map<Class<*>, (Any) -> Any?>? = null) {
```
paulbakker, line 41: Why is the return type nullable?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomJsonDeserializer.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun deserialize(jp: JsonParser, ctxt: DeserializationContext): T? {
```
paulbakker, line 25: Return type should not be nullable I think?

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/GraphQLResponseTest.kt
```diff
@@ -54,7 +55,13 @@ class GraphQLResponseTest {
...
+        url,
+        mapOf(
+            OffsetDateTime::class.java to { t -> OffsetDateTime.parse(t.toString(), DateTimeFormatter.ISO_OFFSET_DATE_TIME) },
+            Person::class.java to { t -> Person("Custom Person", 20) }
```
paulbakker, line 62: Person is not used anywhere, so the mechanism isn't being tested.

paulbakker:
since you don't have access to the json node in the lambda you provide, how would you implement use cases where the target class' fields don't match the json fields? Something like: https://www.baeldung.com/jackson-deserialization
That needs to be supported and verified with a test.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomJsonDeserializer.kt
paulbakker: T is not actually being defined anywhere, because the parameters of the method don't use it

PR #691, opened 2021-10-12
websocket client use websocket subprotocol header during handshake

amondnet:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Issue #

`WebSocketGraphQLClient` client use `Sec-WebSocket-Protocol` during handshake.

WebSocket RFC defines the use of [sub-protocols](https://datatracker.ietf.org/doc/html/rfc6455#section-1.9). During the handshake, the client and server can use the header `Sec-WebSocket-Protocol` to agree on a sub-protocol, i.e. a higher, application-level protocol to use. The use of a sub-protocol is not required, but even if not used, applications will still need to choose a message format that both the client and server can understand. That format can be custom, framework-specific, or a standard messaging protocol.

If you use [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) and  protocol is not specified, the [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws)  closes the connection.
https://github.com/apollographql/subscriptions-transport-ws/blob/231c1b0d25b20090033140b83c8a39871d4ed88c/src/server.ts#L143
```ts
if (socket.protocol === undefined ||
        (socket.protocol.indexOf(GRAPHQL_WS) === -1 && socket.protocol.indexOf(GRAPHQL_SUBSCRIPTIONS) === -1)) {
        // Close the connection with an error code, ws v2 ensures that the
        // connection is cleaned up even when the closing handshake fails.
        // 1002: protocol error
        socket.close(1002);

        return;
      }
```

Alternatives considered
----


srinivasankavitha:
Regarding this comment - `This pr makes dgs-client compatible with apollo-subscription + subscriptions-transport-ws. In this case, we only need to add the header. `, I am not confused. This is the original set up and should be working to begin with, no?  In other words the existing WebsocketClient should work with a server that is running `subscriptions-transport-ws` already. Are you saying it works with the dgs server but not the apollo server, and therefore this header is needed?

amondnet:
@srinivasankavitha Yes, dgs-client does not work with the apollo server running subscriptions-transport-ws.
I am getting error `onError(graphql.GraphQLException: Server closed the connection unexpectedly))` from dgs-client.
If protocol is not specified, apollo server will close the connection with error code `1002`. [1002 indicates that an endpoint is terminating the connection due to a protocol error.](https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1)

srinivasankavitha:
@amondet - thanks a lot for adding more test coverage for the changes. I think the example test with the apollo-server - `graphql-dgs-client-example-apollo` would be a good one to separate out into a separate project rather than have it be part of the framework. We have a separate section in our docs for example projects. Currently it requires docker installation and would be good to avoid that dependency in the framework.

srinivasankavitha:
The rest of the PR looks really good to me! Thanks for making the changes. I think we can merge once the apollo example is separated out to a different repo. We generally add OSS contribution sto examples in our docs here: https://netflix.github.io/dgs/examples/

graphql-dgs-client-example-apollo/src/test/java/subscription/integrationtest/SubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Collections;
+
+
+@Testcontainers
+public class SubscriptionIntegrationTest {
+    private static final Logger logger = LoggerFactory.getLogger(SubscriptionIntegrationTest.class);
+
+    @Container
+    private static final GenericContainer<?> apolloContainer = new GenericContainer<>(
+            "amond/graphql-dgs-client-example-apollo"
```
berngp, line 47: @amondnet I think this depends on the image that you are publishing to Dockerhub.
Although the test is interesting, and potentially worth to showcase as a pattern, I don't think it should reside in the framework.

paulbakker:
@amondnet Although I agree with @srinivasankavitha that the test container setup is probably a bit more than we want/need for the framework tests, it does look really interesting. May I suggest writing a blog post about that approach?

graphql-dgs-subscriptions-websockets-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/DgsWebSocketAutoConfig.kt
```diff
@@ -36,12 +37,21 @@ open class DgsWebSocketAutoConfig {
...
+    ) : WebSocketConfigurer {
 
         override fun registerWebSocketHandlers(registry: WebSocketHandlerRegistry) {
-            val defaultHandshakeHandler = DefaultHandshakeHandler()
-            defaultHandshakeHandler.setSupportedProtocols("graphql-ws")
-            registry.addHandler(webSocketHandler, "/subscriptions").setHandshakeHandler(defaultHandshakeHandler).setAllowedOrigins("*")
+            val handshakeHandler = DefaultHandshakeHandler()
+            registry.addHandler(webSocketHandler, "/subscriptions").setHandshakeHandler(handshakeHandler)
+                .addInterceptors(handshakeInterceptor)
+                .setAllowedOrigins("*")
```
berngp, line 49: We probably want to make `allowedOrigins` configurable.

graphql-dgs-client-example-apollo/src/test/java/subscription/integrationtest/SubscriptionIntegrationTest.java
amondnet: @berngp I removed that module.

amondnet:
@srinivasankavitha I removed `graphql-dgs-client-example-apollo` module.

srinivasankavitha:
Looks good! Thanks so much for all the work and your patience!

srinivasankavitha (approved):

berngp (approved):

Issue #726, opened 2021-11-08
bug: Federation _entities field/entity fetchers do not catch failed CompletionStage properly


victorlevasseur:
**Specs :** 
Java 11
DGS 4.9.3

## Expected behavior

Consider this entity fetcher, that throws a RuntimeException if the "Cool" type with id == "notCool" is requested by our Apollo Federation Gateway.

```java
  @DgsEntityFetcher(name = "Cool")
  public CompletableFuture<Cool> cool(Map<String, Object> attributes,
      DgsDataFetchingEnvironment env) {
    String id = (String) attributes.get("id");
    if (id.equals("notCool")) {
      return CompletableFuture.failedFuture(new RuntimeException("BAD!!!"));
    } else {
      return CompletableFuture.completedFuture(Cool.newBuilder().id(id).name("cool!").build());
    }
  }
```

The Apollo Gateway might try to do this request, containing the bad "Cool" value:

```graphql
query {
  _entities(representations: [
    {
      __typename: "Cool",
      id: "cool"
    },
    {
      __typename: "Cool",
      id: "notCool"
    },
    {
      __typename: "Cool",
      id: "cool3"
    }
  ]) {
    ... on Cool {
      id
      name
    }
  }
}
```

Following its best-effort strategy, the DGS GraphQL app should answer the following response, with a second element of the array represented as null (as it threw the RuntimeException) : 

```graphql
{
  "errors": [
    {
      "message": "java.lang.RuntimeException: BAD!!!",
      "locations": [],
      "path": [
        "_entities"
      ],
      "extensions": {
        "errorType": "INTERNAL"
      }
    }
  ],
  "data": {
    "_entities": [
      {
        "id": "cool",
        "name": "cool!"
      },
      null,
      {
        "id": "cool3",
        "name": "cool!"
      }
    ]
  }
}
```

## Actual behavior

Instead of returning partial results, it responds with:

```graphql
{
  "errors": [
    {
      "message": "java.lang.RuntimeException: BAD!!!",
      "locations": [],
      "path": [
        "_entities"
      ],
      "extensions": {
        "errorType": "INTERNAL"
      }
    }
  ],
  "data": null
}
```

## Workarounds and other cases

I saw that a recent issue tried to fix the exceptions in the entities fetchers. However, the fix seems to only apply to
[...]

PR #730, opened 2021-11-10
DefaultDgsFederationResolver now handles failed completion stages ret…

victorlevasseur:
…urned from entity fetchers properly

Pull request checklist
----

  first

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Issue #726 

It allows the DefaultDgsFederationResolver to properly handle exceptions thrown by failed completable futures.

I used the dataloader Try object to ease the handling of exception and to make sure to have a single path of execution for exceptions thrown in the entity fetcher body and also the exceptions thrown by failed completable futures. That's why the whole call to each entity fetchers are wrapped in Try.tryCallable (to handle exceptions thrown directly in the entity fetcher) and then unwrapped when combined with the result of the CompletableFuture (to handle exceptions thrown by a failed completable future).

I added a new tests to check that this use case now works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ build
 .idea/**/dynamic.xml
 .idea/**/uiDesigner.xml
 .idea/**/dbnavigator.xml
+.idea/**/runConfigurations.xml
+.idea/**/aws.xml
 
 # Gradle
 .idea/**/gradle.xml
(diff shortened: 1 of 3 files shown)


srinivasankavitha:
Thanks for the PR @victorlevasseur. Will test  this out and get back to you shortly.

.gitignore
```diff
@@ -23,6 +23,8 @@ build
 .idea/**/dynamic.xml
 .idea/**/uiDesigner.xml
 .idea/**/dbnavigator.xml
+.idea/**/runConfigurations.xml
+.idea/**/aws.xml
```
berngp, line 27: I don't think we need this here.

berngp (approved):
Changes look good, thanks for doing this. Suggest just to remove the .idea configuration changes.

PR #754, opened 2021-11-27
Add more IDE project specific files into gitignore

hantsy:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe):

Changes in this PR
----

These files are generated by IDEs when opening a Gradle based projects.

They are common files to be ignored in a Java project.

If you generate an empty project from Spring initializr or create an empty project on Github, it also generate a gitignore file include all IDE specific files which should not be managed by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,25 @@ buildSrc/.gradle/
 /.idea/kotlinScripting.xml
 scripts/__pycache__/
 /.idea/libraries-with-intellij-classes.xml
+.idea/google-java-format.xml
+
+# STS and Eclipse/VSCode Java Pack
+.apt_generated
+.apt_generated_test
+.classpath
+.factorypath
+.project
+.settings
+.springBeans
+.sts4-cache
+bin/
+
+# NetBeans
+/nbproject/private/
+/nbbuild/
+/dist/
+/nbdist/
+/.nb-gradle/
+
+# VS Code
+.vscode/


berngp:
I think most of these belong to a global git ignore and are not specific to this project.

srinivasankavitha:
I agree with @berngp. These files don't seem related to the dgs framework.

hantsy:
I do not set a global gitignore, when opening this project with VsCode, it
will generate Java related project files(similar to Eclipse project).

*Hantsy Bai*

Self-employed consultant, fullstack developer, agile coach

GitHub: https://github.com/hantsy

Twitter: ***@***.***
Medium: ***@***.***

srinivasankavitha:
I see, so these are generated when using VSCode only?

berngp:
If we could express that the idea of the PR is to better support other IDE's and ignore their project specific generated files, then I think its ok. As is the PR stands without any description.

hantsy:
These files are generated by IDEs when opening a Gradle based projects. 

They are common files to be ignored in a Java project. 

 If you generate an emtpy project from Spring initializr or create an empty project on Github, it also generate a gitignore file include all IDE specific files which should not be managed by Git.

berngp (approved):

srinivasankavitha (approved):

PR #779, opened 2021-12-10
Recommending Log4J versions >= 2.15

berngp:
Objective
========

Make sure that consumers of the DGS Starters don't get Log4J dependencies < 2.15

Ref. https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228

### Example

With this change the `graphql-dgs-spring-boot-starter` will expose `org.apache.logging.log4j:log4j-to-slf4j` version `2.15.0` instead of `2.13.3`.

```
org.apache.logging.log4j:log4j-to-slf4j:2.13.3 -> 2.15.0
+--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE
|    \--- compileClasspath
\--- org.springframework.boot:spring-boot-starter-logging:2.3.12.RELEASE
     +--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE (*)
     \--- org.springframework.boot:spring-boot-starter:2.3.12.RELEASE
          +--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE (*)
          +--- org.springframework.boot:spring-boot-starter-web:2.3.12.RELEASE
          |    +--- compileClasspath (requested org.springframework.boot:spring-boot-starter-web)
          |    +--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE (*)
          |    \--- org.springframework.boot:spring-boot-starter-websocket:2.3.12.RELEASE
          |         +--- compileClasspath (requested org.springframework.boot:spring-boot-starter-websocket)
          |         \--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE (*)
          \--- org.springframework.boot:spring-boot-starter-json:2.3.12.RELEASE
               +--- org.springframework.boot:spring-boot-dependencies:2.3.12.RELEASE (*)
               \--- org.springframework.boot:spring-boot-starter-web:2.3.12.RELEASE (*)
```

### DIFF
--- a/graphql-dgs-client/dependencies.lock
+++ b/graphql-dgs-client/dependencies.lock
@@ -436,6 +436,7 @@
                 "com.netflix.graphql.dgs:graphql-dgs-spring-webmvc-autoconfigure",
                 "com.netflix.graphql.dgs:graphql-dgs-subscription-types",
                 "com.netflix.graphql.dgs:graphql-dgs-subscriptions-sse-autoconfigure",
+                "com.netflix.graphql.dgs:graphql-dgs-subscriptions-websockets-autoconfigure",
                 "com.netflix.graphql.dgs:graphql-error-types"
             ],
             "project": true
@@ -458,6 +459,9 @@
         "com.netflix.graphql.dgs:graphql-dgs-subscriptions-sse-autoconfigure": {
             "project": true
         },
+        "com.netflix.graphql.dgs:graphql-dgs-subscriptions-websockets-autoconfigure": {
+            "project": true
+        },
         "com.netflix.graphql.dgs:graphql-error-types": {
             "firstLevelTransitive": [
                 "com.netflix.graphql.dgs:graphql-dgs",
@@ -466,7 +470,7 @@
             "project": true
         },
         "io.mockk:mockk": {
-            "locked": "1.12.0"
+            "locked": "1.12.1"
         },
         "io.projectreactor:reactor-core": {
             "locked": "3.4.10"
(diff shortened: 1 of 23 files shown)


srinivasankavitha (approved):

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -15,10 +15,5 @@
  */
 
 object Versions {
-    const val KOTLIN_VERSION = "1.5.21"
-    const val SPRING_VERSION = "5.2.13.RELEASE"
```
paulbakker, line 19: Any reason not to use these constants any more?

paulbakker (commented):
Would it be better to only upgrade log4j, do an emergency release, and then upgrade other stuff later?

buildSrc/src/main/kotlin/Versions.kt
berngp: We are keeping the KOTLIN_VERSION  there since it's used in a plugin and the Koltin BOM reference. For the others it is easier to reason if they are defined with the dependency they affect. For most, all except Spring Boot, there is only one place such versions were referenced.

berngp: Moved this part to a different PR.

PinkDev1:
> Would it be better to only upgrade log4j, do an emergency release, and then upgrade other stuff later?

Yes, 100%

berngp:
@PinkDev1 btw, we already did a release that covers only that.

Issue #450, opened 2021-06-28
feature: Add Bearer auth token support to subscriptions


metsfan:
## Describe the Feature Request
Right now it is not possible to have bearer authorization on a DGS subscription. Therefore if your subscription requires user authorization in order to succeed, it will not be possible. 

## Describe Preferred Solution
Allow some methodology for providing an authorization interceptor on the DGS subscription websocket

## Describe Alternatives
Allow library clients to provide an alternative bean for the websocket handler.

paulbakker:
Websockets don't have HTTP headers, which is why you can't use regular headers for this.

At Netflix we use a filter that basically takes the bearer out of a URL parameter. Since this isn't really standard, I'm not sure what the framework can/should do.

metsfan:
So the Apollo clients allow you to provide a "payload" on the initial websocket connection. In the GraphQL kickstart library you can define a bean which extends `ApolloSubscriptionConnectionListener` and allows you to read from the initialization payload and update the Spring security context. See a code sample below from my project (which currently uses GraphQL kickstart). 

If we could get something like this in DGS, it would be very helpful. 

However I am interested in how you are doing this in Netflix. How do you access this URL parameter? And is there any way to execute this as a filter so I can still use the `@PreAuthorize` annotation on my `@DgsSubscription` method? Sorry if this is an obvious question. 

```
@Component
class SubscriptionConnectionListener(
    private val tokenAuthorizationFilter: TokenAuthorizationFilter,
    private val authenticator: TokenBasedAuthenticator
) : ApolloSubscriptionConnectionListener {
    override fun onConnect(session: SubscriptionSession?, message: OperationMessage?) {
        super.onConnect(session, message)

        (message?.payload as? Map<*, *>)?.let { payload ->
            payload[AuthorizationHeader]?.toString()?.let {
                session?.userProperties?.put(AuthorizationHeader, it)
            }
        }
    }

    override fun onStart(session: SubscriptionSession?, message: OperationMessage?) {
        super.onStart(session, message)

        session?.userProperties?.get(AuthorizationHeader)?.toString()?.let { authorization ->
            tokenAuthorizationFilter.readToken(authorization)?.let { token ->
                authenticator.authenticateUser(token)
            }
        }
    }
}
```

metsfan:
This is what I mean btw. As you can see during hte GraphQL initialization, I've passed through a payload from the IOS client which has the Authorization header. All we need here is a listener here that passes through the payload. 
[...]

PR #814, opened 2022-01-05
Process HTTP session data on websocket handshake and load SecurityContext into Subscriptions

lucatk:
Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

This fixes a bug where we don't receive the correct SecurityContext in subscriptions, because the HTTP session etc. is only available during the handshake, meaning Spring Security logic (annotations, etc.) does not work.
The `HttpSessionHandshakeInterceptor` handles the extraction of HTTP headers during the handshake and makes it available as attributes on the `WebSocketSession`.
Then the correct SecurityContext is loaded prior to handling any WebSocket message.

Alternatives considered
----

- Create own WebSocket controller, but the DgsWebSocketHandler class needs to be `open` so we can extend it and add the loading of the SecurityContext for ourselves.

### DIFF
--- a/graphql-dgs-subscriptions-websockets/build.gradle.kts
+++ b/graphql-dgs-subscriptions-websockets/build.gradle.kts
@@ -23,6 +23,8 @@ dependencies {
     implementation("org.springframework:spring-websocket")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
 
+    compileOnly("org.springframework.security:spring-security-core")
+
     testImplementation("io.projectreactor:reactor-core")
     testImplementation("io.projectreactor.kotlin:reactor-kotlin-extensions")
 }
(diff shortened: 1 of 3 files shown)


srinivasankavitha:
Thanks for the PR! Are you able to add tests to our existing test suite to verify this functionality?

berngp (commented):
Approach seems reasonable, thanks for the fix!

berngp (approved):

lucatk:
Thanks for merging and releasing this ever so quickly!❤️

LexanRed:
Thanks for your effort! I am running in the same issues. 
Unfortunately, I still get the same problems.

What is the correct way to retrieve the Security Context in a method, called by a subscription?

We use ```Authentication authentication = SecurityContextHolder.getContext().getAuthentication();```
This sometimes works, but not always. I guess there must be a better way.

Any help is very much appreciated 😄

LexanRed:
Any input on this one? As described in https://github.com/Netflix/dgs-framework/issues/450, getting the security context fails sometimes. We managed to figure out, it fails when no mutation / query has been sent before. No session exists in this case. 
How do you manage to reliably retrieve the security context?

Feedcube:
Thanks a lot for fixing this issue @lucatk !
How did you manage to intercept the handshake to set the authentication object?
When i override the WebSocketConfigurer in an own class to add the interceptor, the registerWebSocketHandlers gets not called

PR #840, opened 2022-01-24
Issue-841: Fix NPE in equals method of TypedGraphQLError

rabbitvirus:
Related to Issue #841


berngp:
Thank you for the PR!

PR #842, opened 2022-01-24
Support custom input object mapper

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Issue #775
With this PR it's possible to provide your own `InputObjectMapper`. This can be useful for example to filter out fields from an input object, which is mostly to integrate with other frameworks/libraries, and is not intended to be used by most users.

Note that because `DefaultInputObjectMapper` calls itself recursively, for each field it goes a level deeper into an object structure it will call the custom mapper if available. The custom mapper in turn, might still call the `DefaultInputObjectMapper` after modifying the input map. It's easy to create a never ending recursion here, so be careful implementing this!

This extension does *not* make it possible to drop in a Jackson Object Mapper, since the input isn't JSON.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
@@ -33,6 +33,15 @@ import kotlin.reflect.jvm.javaType
 import kotlin.reflect.jvm.jvmErasure
 
 interface InputObjectMapper {
+    /**
+     * SPI intended for other frameworks/libraries that need to customize how input objects are mapped.
+     * Not intended to be used by most users.
+     *
+     * A custom mapper might call the DefaultInputObjectMapper, passing itself to the DefaultInputObjectMapper constructor.
+     * The DefaultInputObjectMapper will invoke the custom mapper each time it goes a level deeper into a nested object structure.
+     * This makes it possible to have a custom mapper that still mostly relies on the default one.
+     * Be careful to not create an infinite recursion calling back and forward though!
+     */
     fun <T : Any> mapToKotlinObject(inputMap: Map<String, *>, targetClass: KClass<T>): T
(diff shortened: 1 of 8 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/InputObjectMapper.kt
```diff
@@ -32,11 +32,25 @@ import kotlin.reflect.full.primaryConstructor
...
+    /**
+     * SPI intended for other frameworks/libraries that need to customize how input objects are mapped.
+     * Not intended to be used by most users.
+     *
+     * A custom mapper might call the DefaultInputObjectMapper, passing itself to the DefaultInputObjectMapper constructor.
+     * The DefaultInputObjectMapper will invoke the custom mapper each time it goes a level deeper into a nested object structure.
+     * This makes it possible to have a custom mapper that still mostly relies on the default one.
+     * Be careful to not create an infinite recursion calling back and forward though!
+     */
+    fun <T : Any> mapToKotlinObject(inputMap: Map<String, *>, targetClass: KClass<T>): T
```
berngp, line 45: Great feature @paulbakker, my only concern is ...

If I am a Java Application developer that doesn't use Kotlin in my project explicitly why would I need to be concerned on `mapToKotlinObject`.

berngp: Maybe we could have a `JavaInputObjectMapper` as well as a `KotlinInputObjectMapper`?

paulbakker: Since this isn't really targeted at application developers (it's really for library/framework integration), I think anyone who would implement this probably *should* care.

berngp (approved):

srinivasankavitha (approved):

golfingal72:
@paulbakker any plans to implement custom OUTPUT mapper?

PR #866, opened 2022-02-09
Move result processors for Flux/Mono to graphql-dgs

kilink:
Move MonoDataFetcherResultProcessor and FluxDataFetcherResultProcessor out
of graphql-dgs-reactive and into graphql-dgs, so they can be used without
WebFlux. DgsAutoConfiguration will now register them as beans conditional
on the presence of the Mono / Flux classes.

### DIFF
--- a/graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
+++ b/graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
@@ -20,13 +20,13 @@ import com.netflix.graphql.dgs.internal.CookieValueResolver
 import com.netflix.graphql.dgs.internal.DefaultDgsQueryExecutor
 import com.netflix.graphql.dgs.internal.DgsDataLoaderProvider
 import com.netflix.graphql.dgs.internal.DgsSchemaProvider
+import com.netflix.graphql.dgs.internal.FluxDataFetcherResultProcessor
+import com.netflix.graphql.dgs.internal.MonoDataFetcherResultProcessor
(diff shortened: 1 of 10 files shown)


graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt
```diff
@@ -190,11 +190,13 @@ open class DgsWebFluxAutoConfiguration(private val configProps: DgsWebfluxConfig
     }
 
     @Bean
+    @ConditionalOnMissingBean
```
kilink, line 193: I added @ConditionalOnMissingBean in both places out of caution, in case someone does happen to include both DgsWebFluxAutoConfiguration and DgsAutoConfiguration; in fact, a few of the integration tests were importing both.

graphql-dgs-spring-webflux-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/webflux/MalformedQueryContentTest.kt
```diff
@@ -18,25 +18,26 @@ package com.netflix.graphql.dgs.webflux
...
+import org.springframework.context.annotation.Import
 import org.springframework.test.web.reactive.server.WebTestClient
 import org.springframework.test.web.reactive.server.expectBody
 import org.springframework.web.reactive.config.EnableWebFlux
 
 @SpringBootTest
 @EnableWebFlux
 @AutoConfigureWebTestClient
-@EnableAutoConfiguration
+@Import(DgsWebFluxAutoConfiguration::class)
```
kilink, line 40: Without scoping this to DgsWebFluxAutoConfiguration, DgsAutoConfiguration was being discovered and also imported.

paulbakker:
Thanks a lot @kilink!

Issue #861, opened 2022-02-05
feature: Use class name by default for `@DgsDataLoader#name`, and add some inline methods to `DgsDataFetchingEnvironment`


jord1e:
~~Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue. Also consider discussing your idea on
the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first.~~

## Describe the Feature Request
The the annotated class's name as the `name` field of the `@DgsDataLoader` annotation __by default__
When annotating methods use something like `SimpleClassName.methodName`

## Describe Preferred Solution
Using
```kotlin
@DgsDataLoader
class PersonByIdDataLoader : MappedBatchLoader<String, Try<Person>>{
    override fun load(keys: Set<String>): CompletionStage<Map<String, Try<Person>>> {
        TODO("Not yet implemented")
    }
}
```

We can then do:
```kotlin
val dl = dfe.getDataLoader<String, Try<Person>>(PersonByIdDataLoader::class.java)
// or
val dl = dfe.getDataLoader<String, Try<Person>>("PersonByIdDataLoader")
```

We may also wish to add some extra (inline-reified) methods to DgsDataFetchingEnvironment, for example:
```kotlin
inline fun <K, V, reified T : MappedBatchLoader<K, V>> DgsDataFetchingEnvironment.getMappedBatchLoader(): DataLoader<K, V> {
    return this.getDataLoader(T::class.java)
}
inline fun <K, V, reified T : BatchLoader<K, V>> DgsDataFetchingEnvironment.getBatchLoader(): DataLoader<K, V> {
    return this.getDataLoader(T::class.java)
}
```

Which gives us compile-time type checking:
incorrect:
![afbeelding](https://user-images.githubusercontent.com/30464310/152649360-30da0d63-2165-4cfa-adae-e68c3791c3e4.png)

correct:
![afbeelding](https://user-images.githubusercontent.com/30464310/152649642-98ceedb6-4843-4c9a-aba9-94a2d8829a6f.png)

## Describe Alternatives
For the annotation proposal: keep it as is

For the methods:
```kotlin
inline fun <K, V, reified T> DgsDataFetchingEnvironment.getDataLoader(): DataLoader<K, V> {
    return this.getDataLoader(T::class.java)
}
```

would work instead of 4 seperate functions, but this does not do any type checking, it merely avoids having to use `::class.java`:
[...]

PR #904, opened 2022-03-03
Automatic DgsDataLoader naming

jord1e:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

This PR automatically names dataloaders, which is useful when using the `DgsDataFetchingEnvironment#getDataLoader(Class)` method.

It also implements automatic naming of field loaders, which is arguably less useful. If anyone objects to this we can error out on nameless field dataloaders.

Something to consider is that metric and tracing solutions may expose (field?)/class/package names of the dataloader. In the future we can add a `DgsDataLoaderNamingFactory` or something that gives library consumers the ability to modify the name programmatically (e.g. - only use the class name instead of the FQN) - this PR does not implement this.

Partially related to https://github.com/Netflix/dgs-framework/issues/861

Alternatives considered
----

Name the loaders

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Netflix, Inc.
+ * Copyright 2022 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


srinivasankavitha:
Thanks for the PR. The changes look good to me, I need to test it out a bit before merging. Will do that this week.

jord1e:
Please look at the automatic field naming the closest, I am seriously debating leaving it out (and throwing a exception on *unnamed* field dataloaders) because refactoring the field name breaks code (fields cannot be referenced as parameters in methods)

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
```diff
@@ -37,10 +44,12 @@ import javax.annotation.PostConstruct
  */
 class DgsDataLoaderProvider(private val applicationContext: ApplicationContext) {
 
-    private val batchLoaders = mutableListOf<Pair<BatchLoader<*, *>, DgsDataLoader>>()
-    private val batchLoadersWithContext = mutableListOf<Pair<BatchLoaderWithContext<*, *>, DgsDataLoader>>()
-    private val mappedBatchLoaders = mutableListOf<Pair<MappedBatchLoader<*, *>, DgsDataLoader>>()
-    private val mappedBatchLoadersWithContext = mutableListOf<Pair<MappedBatchLoaderWithContext<*, *>, DgsDataLoader>>()
+    data class LoaderHolder<T>(val theLoader: T, val annotation: DgsDataLoader, val name: String)
```
jord1e, line 47: Should be private

srinivasankavitha:
Regarding https://github.com/Netflix/dgs-framework/pull/904#issuecomment-1064340252

I agree, I think generating the names automatically for the field loaders feels like it may result in more confusion than help. Let's remove that and keep just the generation of class data loader names.

Otherwise looks good to me. Thanks for the PR!

srinivasankavitha (changes requested):
We can merge the PR after you get a chance to remove the generation of names for field data loaders. Thanks!

jord1e:
Made the following changes:

- Fields annotated with nameless @DgsDataLoader now throw `DgsUnnamedDataLoaderOnFieldException` upon initialization
- Added javadoc/ktdoc
- DgsDataLoaderProvider refactoring to DRY up some code (did I just invent a new verb?)
- Replaced the `prefix+_+fqnClassName` with just `Class.getSimpleName` (this is unambiguous enough in my opinion)

PR #915, opened 2022-03-08
Remove cyclic dependency between SchemaProvider and DefaultDgsFederat…

paulbakker:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe):

Changes in this PR
----
Remove cyclic dependency between SchemaProvider and DefaultDgsFederationResolver.
I've introduced a `EntityFetcherRegistry` which is a bean shared by schemaProvider and the `DefaultDgsFederationResolver`, so that those beans don't depend on each other anymore.

PR #960, opened 2022-04-04
Bugfix/circular bean dependencies

MayCXC:
Pull request checklist
----

- [ n/a ] Consider creating a discussion on the [discussion forum](https://github.com/Netflix/dgs-framework/discussions)
  first
- [ tbd ] Make sure to have sufficient test cases

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Issue #961
The DGS federation docs describe a way to implement federation entities with a different class name than the graphql type they represent [here](https://netflix.github.io/dgs/federation/#customizing-the-default-federation-resolver), which causes a startup error using graphql-dgs-platform-dependencies 4.9.17:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

┌─────┐
|  myFederationResolver (field public com.netflix.graphql.dgs.internal.DgsSchemaProvider com.netflix.graphql.dgs.federation.DefaultDgsFederationResolver.dgsSchemaProvider)
↑     ↓
|  dgsSchemaProvider defined in class path resource [com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.class]
└─────┘
```

The most consistent way I could see to fix this was to just add DgsFederationResolver to DgsAutoConfiguration, and then use constructor injection in DefaultDgsFederationResolver instead of the Autowired annotation, à la dependency inversion principle. To test this fix I am using [this](https://github.com/MayCXC/dgs-federation-example/tree/test/dgs-auto-configuration) fork of the dgs federation example project, pushing this platform branch to a local maven repository, and changing the version the example uses [here](https://github.com/MayCXC/dgs-federation-example/blob/test/dgs-auto-configuration/inner-dgs/build.gradle.kts#L34) and [here](https://github.com/MayCXC/dgs-federation-example/blob/test/dgs-auto-configuration/outer-dgs/build.gradle.kts#L34). That fork also uses java for both subgraphs, and runs a newer version of Apollo gateway. I have not added tests that can catch this bug to framework itself, because the scope of spring dependency management testing seemed too general.

I also had to delete all of the dependencies.lock files, update buildSrc's `Versions.KOTLIN_VERSION` to `1.6.20`, and use the gradlew update 7.4.2 branch, to get this repo to build at all. Otherwise I get [this](https://youtrack.jetbrains.com/issue/KT-45545) build error, but that might just be a problem with my local environment. Those changes aren't committed to this branch.

Background
----
A use case I've been exploring is using DGS to make subgraphs for Apollo's federated gateway that persist and query Postgres with JPA/hibernate. Using the codegen plugin with interface generation enabled is very useful in this case, as a way to implement a federated graphql "Entity" and a JPA "Entity" with a single class. Then it is tempting to name that implementation differently than the graphql schema type it represents, to make it clear it is not a generated class, which lead to this bug.

Alternatives considered
----

The Spring [@Lazy](baeldung.com/spring-lazy-annotation) annotation doesn't work on closed kotlin classes, and setter injection seemed to make inheritance overly complicated in open classes, but either could be an alternative fix if modifying DgsAutoConfiguration is undesirable.


srinivasankavitha:
A fix for this issue was already released as part of `4.9.22` as part of #915. Could you please try the latest version?

MayCXC:
Oh yeah, looks like I was quite late with this PR. Thanks for fixing this 👍

PR #972, opened 2022-04-10
Get rid of unnecessary .let, use MediaType

kilink:
Get rid of unnecessary .let calls where the varaible was already smart cast or known to not be null. Also, switch to MediaType class for request content type check.

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -31,6 +31,7 @@ import graphql.execution.reactive.SubscriptionPublisher
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpHeaders
+import org.springframework.http.MediaType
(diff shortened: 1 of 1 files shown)


graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -71,6 +72,11 @@ open class DgsRestController(
     open val mapper: ObjectMapper = jacksonObjectMapper()
 ) {
 
+    companion object {
```
berngp, line 75: Isn't the Kotlin convention to put the companion object at the bottom?

berngp (approved):

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
kilink: You're right, I guess it is, according to [this](https://kotlinlang.org/docs/coding-conventions.html#class-layout); guess it just feels weird to me to put constants / loggers at the end of a class.

PR #977, opened 2022-04-11
Deprecate InputArgument collectionType parameter

kilink:
The collectionType parameter to the @InputArgument annotation is not necessary, as generic types for parameters can be resolved on runtime. Update DataFetcherInvoker to rely on the actual parameter types, and also enhance the conversion performed to support arbitrary Collection and Map types.


berngp (approved):
Thanks @kilink!

PR #978, opened 2022-04-12
Use delegation pattern in DgsDataFetchingEnvironment

kilink:
Some of the overridden methods such as getArguments were incorrectly returning mutable types, when the underlying DataFetchingEnvironment implementation was returning immutable types. Switching to the delegation pattern avoids this issue and gets rid of some boilerplate code.


berngp (approved):
Thanks again @kilink

PR #979, opened 2022-04-12
Add support for defining a default DataFetcherFactory via autoconfig

jord1e:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

By default a `PropertyDataFetcher` will be configured when no data fetcher has been specified. You can provide a custom `DataFetcherFactory` to `GraphQLCodeRegistry` which gives you the ability to change this default behaviour.

I have been doing some work on dynamically generating the `PropertyDataFetcher` logic directly (instead of `PropertyDataFetcher`'s reflection behaviour)  in a class using the ASM bytecode library: https://github.com/jord1e/graphql-java-asm-datafetcher

By simply providing a `AsmDataFetcherFactory` the required classes get generated lazily, **increasing performance by 10x+**, matching handwritten data fetchers, in my initial testing (I will further work on this proof of concept to make it better match the default `PropertyDataFetcher` i.e. add generic Map<Object, Object> support).

This is not the only use-case of course but the ability to define a `DataFetcherFactory` as a bean is a low maintaince thing we can do.

Alternatives considered
----

N/A, Spring for GraphQL has a way to customize the GraphQL instance, I don't think DGS has this.


berngp (approved):
Interesting, thanks for the pr @jord1e

PR #982, opened 2022-04-13
Stop using deprecated context methods

kilink:
Switch to using GraphQLContext related methods in DataFetchingEnvironment
and ExecutionInput. Update DgsContext so that it can be passed directly
on the ExecutionInput.Builder's graphQLContext method, and add a static
helper to pull the DgsContext instance from the DataFetchingEnvironment.

### DIFF
--- a/graphql-dgs/src/test/java/com/netflix/graphql/dgs/internal/java/test/inputobjects/sortby/JMovieSortByField.java
+++ b/graphql-dgs/src/test/java/com/netflix/graphql/dgs/internal/java/test/inputobjects/sortby/JMovieSortByField.java
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.internal.java.test.inputobjects.sortby;
 
-enum JMovieSortByField {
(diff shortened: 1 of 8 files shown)


graphql-dgs/src/test/java/com/netflix/graphql/dgs/internal/java/test/inputobjects/sortby/JMovieSortByField.java
```diff
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.internal.java.test.inputobjects.sortby;
 
-enum JMovieSortByField {
+public enum JMovieSortByField {
```
kilink, line 19: This being package private was causing compiler warnings

paulbakker:
Thanks! We discussed this with the graphql-java folks when it was being added but never got around to changing it.

paulbakker (approved):

berngp (approved):

PR #599, opened 2021-09-02
Add framework specific DgsNoOpPreparsedDocumentProvider

jord1e:
Pull request checklist
----

- [x] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
- [x] Consider creating a discussion on the [discussion forum](https://github.com/Netflix/dgs-framework/discussions)
  first
- [x] Make sure the PR doesn't introduce backward compatibility issues
- [x] Make sure to have sufficient test cases

Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe):

Changes in this PR
----

Replaces the graphql-java [`NoOpPreparsedDocumentProvider`](https://github.com/graphql-java/graphql-java/blob/d783ec7ee30369ca2fdecdfd0b74e8bf6d96baa8/src/main/java/graphql/execution/preparsed/NoOpPreparsedDocumentProvider.java) with a Kotlin `object DgsNoOpPreparsedDocumentProvider` because the graphql-java one is annotated with [@Internal](https://github.com/graphql-java/graphql-java/blob/8530366f24ba316075a63402473cb2a38ca36ab3/src/main/java/graphql/Internal.java):

> ```java
> /**
>  * This represents code that the graphql-java project considers internal code that MAY not be stable within
>  * major releases.
>  *
>  * In general unnecessary changes will be avoided but you should not depend on internal classes being stable
>  */
> ```

Test adopted from https://github.com/graphql-java/graphql-java/blob/8530366f24ba316075a63402473cb2a38ca36ab3/src/test/groovy/graphql/execution/preparsed/NoOpPreparsedDocumentProviderTest.groovy

Alternatives considered
----

N/A

PR #989, opened 2022-04-16
Remove DgsNoOpPreparsedDocumentProvider

kilink:
Switch to the NoOpPreparsedDocumentProvider from graphql-java.


berngp (approved):

jord1e:
I implemented this in #599 because the GraphQL-Java one is annotated with `@Internal`, those classes are not public API and thus should never be used

https://github.com/graphql-java/graphql-java/blob/d783ec7ee30369ca2fdecdfd0b74e8bf6d96baa8/src/main/java/graphql/execution/preparsed/NoOpPreparsedDocumentProvider.java#L9

This PR reverts to using the internal non-public implementation. It should probably be reverted

kilink:
Thanks for catching that, I didn't notice it was annotated with `@Internal`. However, I still think it's not necessary to duplicate the `NoOpPreparsedDocumentProvider` in DGS. All we have to do is not pass in anything to the builder returned by `GraphQL.newGraphQL`, and it ends up being the default one used anyhow. See #994

PR #992, opened 2022-04-18
Avoid passing in empty ChainedInstrumentation

kilink:
When no Instrumentation beans are supplied, avoid passing in a ChainedInstrumentation
instance containing an empty list of Instrumentations. Doing so would result in some
unnecessary overhead, as the builder's internal checkInstrumentationDefaultState method
will attempt to add its own default instrumentation.

Not that I did actually see this show up in flame graphs, for a case where I had no instrumentation registered, it was spending time in ChainedInstrumentation's internals, looking up state in a HashMap, etc. This is because by passing in an empty ChainedInstrumentation, we end up with the default DataLoaderDispatcherInstrumentation instance wrapped unnecessarily in a ChainedInstrumentation.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
@@ -17,6 +17,7 @@
 package com.netflix.graphql.dgs.internal
 
 import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.jayway.jsonpath.Configuration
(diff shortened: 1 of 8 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
```diff
@@ -89,15 +90,13 @@ object BaseDgsQueryExecutor {
         val graphQLBuilder =
             GraphQL.newGraphQL(graphQLSchema)
                 .preparsedDocumentProvider(preparsedDocumentProvider)
-                .instrumentation(chainedInstrumentation)
                 .queryExecutionStrategy(queryExecutionStrategy)
                 .mutationExecutionStrategy(mutationExecutionStrategy)
-                .subscriptionExecutionStrategy(SubscriptionExecutionStrategy())
```
kilink, line 95: `SubscriptionExecutionStrategy(this.defaultExceptionHandler)` is what gets aded by default if no execution strategy is supplied for subscriptions; it seems better to rely on the builder's default, imo, unless we were doing this to avoid having the defaultExceptionHandler set?

berngp (approved):

berngp:
Thanks @kilink

PR #994, opened 2022-04-19
Remove usage of NoOpPreparsedDocumentProvider

kilink:
NoOpPreparsedDocumentProvider is marked as internal in graphql-java; however, it is
the default PreparsedDocumentProvider used by the GraphQL.Builder, so we can use it
by virtue of simply not passing anything in when no PreparsedDocumentProvider bean
is available.


berngp:
Thanks @kilink and @jord1e

jord1e:
Glad this is resolved.

I now know about [`ObjectProvider`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/ObjectProvider.html) :)

PR #999, opened 2022-04-21
Expose subscription ID to subscription data fetchers

kilink:
DgsSSESubscriptionHandler was generating the subscription ID returned to the client,
but there was no way for the subscription data fetcher to get access to this ID,
which is useful for logging and debugging purposes. To rectify that, change
DgsSSESubscriptionHandler to use a Flux directly, which allows us to propagate
the subscription id in the context.

Now that DgsSSESubscriptionHandler is opinionated about which Publisher implementation
it uses, a lot of the code was also simplified by avoiding adapting the subscription
Publisher to the SseEmitter ourselves; instead the controller method simply returns
a Flux directly and lets Spring handle that. Another change is that any error condition
occurring before the GraphQL query is executed successfully (e.g., a Publisher is returned)
will result in a non-SSE response; that is, the status code of the response will
be non-200, and the error message will no longer be written as a single server-sent event.

### DIFF
--- a/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
+++ b/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
@@ -22,20 +22,25 @@ import com.netflix.graphql.types.subscription.QueryPayload
 import com.netflix.graphql.types.subscription.SSEDataPayload
 import graphql.ExecutionResult
 import graphql.InvalidSyntaxError
+import graphql.language.OperationDefinition
+import graphql.parser.InvalidSyntaxException
+import graphql.parser.Parser
 import graphql.validation.ValidationError
 import org.reactivestreams.Publisher
-import org.reactivestreams.Subscriber
-import org.reactivestreams.Subscription
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType
-import org.springframework.http.ResponseEntity
+import org.springframework.http.codec.ServerSentEvent
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
-import org.springframework.web.servlet.mvc.method.annotation.SseEmitter
+import org.springframework.web.server.ServerErrorException
+import org.springframework.web.server.ServerWebInputException
+import reactor.core.publisher.Flux
 import java.nio.charset.StandardCharsets
-import java.util.*
+import java.util.Base64
+import java.util.UUID
+import com.netflix.graphql.types.subscription.Error as SseError
 
 /**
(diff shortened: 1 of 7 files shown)


graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -44,119 +49,74 @@ import java.util.*
...
+                "Invalid return type for subscription datafetcher. A subscription datafetcher must return a Publisher<ExecutionResult>. The query was {}",
+                query, exc
+            )
+            throw ServerErrorException("Invalid return type for subscription datafetcher. Was a non-subscription query send to the subscription endpoint?", exc)
         }
 
-        publisher.subscribe(subscriber)
+        val subscriptionId = UUID.randomUUID().toString()
+        return Flux.from(publisher)
+            .contextWrite { ctx -> ctx.put("subscriptionId", subscriptionId) }
```
kilink, line 93: I can move this key to a constant somewhere

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -44,119 +49,74 @@ import java.util.*
...
+                Flux.just(
+                    ServerSentEvent.builder(mapper.writeValueAsString(payload))
+                        .id(UUID.randomUUID().toString())
+                        .build()
+                )
+            }
+    }
 
-        return ResponseEntity.ok(emitter)
+    private fun isSubscriptionQuery(query: String): Boolean {
```
kilink, line 111: I changed the endpoint to explicitly check if the query is a subscription query before attempting to execute it, which seemed like a good idea to me.

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -44,119 +49,74 @@ import java.util.*
...
-            }
-
-            override fun onError(t: Throwable) {
-                logger.error("Error on subscription {}", sessionId, t)
-                val event = SseEmitter.event()
-                    .data(
-                        mapper.writeValueAsString(
-                            SSEDataPayload(
-                                data = null,
-                                errors = listOf(Error(t.message)),
```
kilink, line 115: I think this line was actually a bug. `Error` here is actually a typealias for `java.lang.Error`, so we'd end up sending back a full stack trace. I changed it to use `com.netflix.graphql.types.subscription.Error`, which just has a `message` field instead, since I think that is what was intended. Maybe the name of the class should be changed so that it doesn't so easily conflict with the kotlin `Error` typealias?

kilink:
After some discussion, we decided to handle propagating of the subscription ID separately from refactoring the code to leverage Flux. I'll close this PR in favor of #1001, and open a new one to handle the subscription ID.

PR #1001, opened 2022-04-22
Simplify DgsSSESubscriptionHandler by returning Flux

kilink:
DgsSSESubscriptionHandler was rolling its own Subscriber implementation in order
to adapt the Publisher returned by the GraphQL query to the SseEmitter; however,
Spring itself will handle this when the content type is "text/event-stream" and
we simply return a reactive type, such as Flux. Doing so greatly simplifies the
code in the controller, and allows for future enhancements such as propagating
context to the publisher.

Another change is that any error condition occurring before the GraphQL query is
executed successfully (e.g., a Publisher is returned) will result in a non-SSE response;
that is, the status code of the response will be non-200, and the error message
will no longer be written as a single server-sent event.

Finally, the way errors are written has been fixed. Previously the entire stack trace
was accidentally written to the event, now only the exception message is propagated.

### DIFF
--- a/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
+++ b/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
@@ -22,20 +22,25 @@ import com.netflix.graphql.types.subscription.QueryPayload
 import com.netflix.graphql.types.subscription.SSEDataPayload
 import graphql.ExecutionResult
 import graphql.InvalidSyntaxError
+import graphql.language.OperationDefinition
+import graphql.parser.InvalidSyntaxException
+import graphql.parser.Parser
 import graphql.validation.ValidationError
 import org.reactivestreams.Publisher
-import org.reactivestreams.Subscriber
-import org.reactivestreams.Subscription
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType
-import org.springframework.http.ResponseEntity
+import org.springframework.http.codec.ServerSentEvent
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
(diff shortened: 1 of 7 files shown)


graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -44,119 +49,73 @@ import java.util.*
...
-                logger.error(errorMessage)
-                emitter.send(errorMessage)
-                emitter.complete()
-                ResponseEntity.status(500).body(emitter)
+                "Error executing subscription query: ${executionResult.errors.joinToString()}"
             }
-        }
-
-        val subscriber = object : Subscriber<ExecutionResult> {
-            lateinit var subscription: Subscription
```
kilink, line 87: I think this actually wasn't thread safe, since the property wasn't volatile / access wasn't synchronized, although in practice I never saw any actual issues.

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -44,119 +49,73 @@ import java.util.*
...
+        } catch (exc: ClassCastException) {
+            logger.error(
+                "Invalid return type for subscription datafetcher. A subscription datafetcher must return a Publisher<ExecutionResult>. The query was {}",
+                query, exc
+            )
+            throw ServerErrorException("Invalid return type for subscription datafetcher. Was a non-subscription query send to the subscription endpoint?", exc)
         }
 
-        publisher.subscribe(subscriber)
+        val subscriptionId = UUID.randomUUID().toString()
```
srinivasankavitha, line 91: How does the execution Id get set here?

berngp: I don't think we can get the ExecutionId here.

berngp (approved):
Thanks for all these work @kilink !

srinivasankavitha (approved):

Issue #582, opened 2021-08-27
bug: Kotlin WebFlux - Request Context is not propagating to suspended Resolver function


darko1002001:
I am using Kotlin with Spring WebFlux + Spring Security. I want to be able to read the authenticated user information as part of the request context. There is a utility class called ReactiveSecurityContextHolder provided by Spring Security that read this information. Unfortunately the data stored in the request context is not available when executing the resolver methods.

There is an issue related to this and a proposed solution which helps with solving the case where a resolver returns a Mono or a Flux instance.

[https://github.com/Netflix/dgs-framework/issues/375](https://github.com/Netflix/dgs-framework/issues/375)

However this doesn't help when defining a Kotlin suspend function as the resolver method. In this case the context data is not available.

I think this is related to how the coroutines are executed here:

https://github.com/Netflix/dgs-framework/blob/cb762dae368583c6a4a0754b6a845f859d9f5b59/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt#L106

The simplest way to show the problem is to do the following:

- Register a custom Context Builder

```
@DgsComponent
class MonoContextBuilder : DgsReactiveCustomContextBuilderWithRequest<WebContext> {
    override fun build(
        extensions: Map<String, Any>?,
        headers: HttpHeaders?,
        serverRequest: ServerRequest?
    ): Mono<WebContext> {
        return Mono.deferContextual { context ->
            Mono.just(WebContext(context))
        }
    }
}
```

- create a separate Bean or Component of a DataFetcherResultProcessor. This is nearly identical to the one included with the library, except for copying over the request context we got above into the next request that executes the resolver. This will work if the return type of the resolver function is a Mono in this case.

```
@Component
class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
    override fun supportsType(originalResult: Any): Boolean {
        return originalResult is Mono<*>
    }

    override fun process(originalResult: Any, dfe: DgsDataFetchingEnvironment): Any {
        if (originalResult is Mono<*>) {
            val context = (dfe.getContext<DgsContext>().customContext as WebContext).requestContext
            return originalResult.contextWrite(context).toFuture()
        } else {
            throw IllegalArgumentException("Instance passed to ${this::class.qualifiedName} was not a Mono<*>. It was a ${originalResult::class.qualifiedName} instead")
        }
    }
}
```

Here is a working example for a resolver returning Mono:

```
   @DgsQuery
[...]

PR #1040, opened 2022-05-10
Preserve Reactor context in reactive data fetcher

gnoeley:
- Add field to DgsContext to store nullable ContextView instance when
creating a DgsContext in of the DefaultDgsReactiveGraphQLContextBuilder.
- Read Reactor context inside reactive data fetcher result processor
implementations and copy onto the subscriber context for the data fetcher
operation.
- Add example of setting and accessing the Reactor context inside a WebFlux
application; including a sample WebFilter and DgsQuery

Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

See comment https://github.com/Netflix/dgs-framework/pull/1040#issuecomment-1122616050

Alternatives considered
----

I had considered just wiring the existing solution from https://github.com/Netflix/dgs-framework/issues/375 in the Spring auto-configuration. However, this would end up conflicting with any other 'custom' result processors or custom context builders which users' themselves registered. It would also be specific to _just_ the Web Request context or Security Holder context.
There is one field which was I chose to add to `DgsContext`. However, this has not caused any compilation issues in sub-modules without that required dependency as the new argument is either omitted or remains as `null` in non-reactive projects (I'm happy for someone to double check this however!).

The chosen approach should allow for the context to be preserved transparently, outside of any other customisations and preserves the entire Reactor Context as it was available when building the initial `DgsContext`.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ReactiveDataFetcherResultProcessor.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ReactiveDataFetcherResultProcessor.kt
@@ -17,8 +17,10 @@
 package com.netflix.graphql.dgs.internal
 
 import com.netflix.graphql.dgs.DgsDataFetchingEnvironment
+import com.netflix.graphql.dgs.context.DgsContext
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
+import reactor.util.context.Context
 import java.lang.IllegalArgumentException
 
 class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
@@ -28,7 +30,7 @@ class MonoDataFetcherResultProcessor : DataFetcherResultProcessor {
 
     override fun process(originalResult: Any, dfe: DgsDataFetchingEnvironment): Any {
         if (originalResult is Mono<*>) {
-            return originalResult.toFuture()
+            return originalResult.contextWrite(reactorContextFrom(dfe)).toFuture()
         } else {
             throw IllegalArgumentException("Instance passed to ${this::class.qualifiedName} was not a Mono<*>. It was a ${originalResult::class.qualifiedName} instead")
         }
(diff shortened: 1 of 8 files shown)


gnoeley:
This proposed change would automatically allow access of the Reactor context from inside Data Fetcher invocations without having to manually provide a DgsContext customizer and override the default reactive data fetcher result processors. For example, the workaround described in this issue (https://github.com/Netflix/dgs-framework/issues/375) should no longer be required and the context would just 'work' for Spring WebFlux reactive applications.

This is a precursor to fixing an outstanding bug with Kotlin suspend function `ReactorContext` missing during invocations. See https://github.com/Netflix/dgs-framework/issues/582. I will try raise a follow up PR for that soon.

srinivasankavitha:
Thanks for the PR! The changes look good to me. I'm still debating whether we should be adding a reactive-specific context field in the DgsContext, the changes look safe but it does make the DgsContext reactive-aware.

gnoeley:
Pushed a potential resolution to avoid making DgsContext reactive-aware. I agree, it didn't seem like a good choice although the sub-class is a bit heavyweight too. 
I would have liked to have created this new class inside the `graphql-dgs-reactive` module but this was not possible due to the dependency requirement inside the `ReactiveDataFetcherResultProcessor` instances inside of `graphql-dgs`. This might be possible with a larger refactor but I don't have enough knowledge yet to make a good call on that.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ReactiveDataFetcherResultProcessor.kt
```diff
@@ -52,4 +51,4 @@ class FluxDataFetcherResultProcessor : DataFetcherResultProcessor {
 }
 
 private fun reactorContextFrom(dfe: DgsDataFetchingEnvironment) =
-    DgsContext.from(dfe).reactorContext ?: Context.empty()
+    ReactiveDgsContext.from(dfe)?.reactorContext ?: Context.empty()
```
gnoeley, line 54: Not sure if we should prefer throwing the `NullPointerException` here instead of all this null-handling. In theory with this change, there should always be a `ReactiveDgsContext` available in the environment. That relies on component configuration though, not the type checker... 🤔

gnoeley: TLDR; Seems that this could sometimes be null when not depending on `graphql-dgs-spring-webflux-autoconfigure`.

The result processors are configured in `DgsAutoConfiguration` (used by a `DgsSchemaProvider`) whereas this new `ReactiveDgsContext` will only be constructed inside of the `DefaultDgsReactiveGraphQLContextBuilder` class made available by `DgsWebFluxAutoConfiguration`. Leaving the null checks for now is likely required.

You can use Mono/Flux in regular Spring Web MVC though. This overall PR would currently only work for Spring WebFlux apps; without modifying the existing `DefaultDgsQueryExecutor` to also be reactive-aware. I'm happy with that trade-off personally but should perhaps be documented.

berngp: @gnoeley we have seen folks returning a _Mono_ or _Flux_ in Spring MVC, I don't think this approach will break those use-cases right?


Thanks for your changes, this PR LGTM

berngp (approved):

PR #1041, opened 2022-05-11
Fix missing Reactor Context in Kotlin suspend datafetchers

gnoeley:
Use same mechanism used by Spring Core to invoke a Kotlin
suspending function wrapped as a Mono or Flux as appropriate.
This ensures the Reactor Context is made available inside of the
Koltin coroutine context allowing access to values set outside of the
DGS framework (i.e. Spring Security / WebFilter).

- Add additional dependency to provide Koltin Reactor builder functions
required by Spring Core library.
- Update unit test to verify Kotlin suspending DataFetcher can access the
expected context.
- Update unit test to verify the query results to avoid false-positive test
results when an unexpected expection is thrown.
- Update unit test to remove custom ExecutorService to avoid race conditions.
This still allows for validating the delay() functions do not cause all
datafetchers to be processed sequentially on a single Unconfined Dispatcher.

Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
This change ensures that the Reactor Context is added to the Kotlin coroutine context when invoking suspending datafetcher methods. This will allow for passing things such as custom Reactor subscriber context from WebFilters or Spring SecurityContextHolder into the coroutine without patching onto the `DgsContext.custom` field.

Issue https://github.com/Netflix/dgs-framework/issues/582

Alternatives considered
----

Couldn't think of a more sensible way to implement this other than doing exactly what Spring Web already does to solve this exact same problem. Has resulted in one more (small) dependency though, hopefully this is an acceptable trade-off?

### DIFF
--- a/graphql-dgs/build.gradle.kts
+++ b/graphql-dgs/build.gradle.kts
@@ -35,6 +35,7 @@ dependencies {
     implementation("com.apollographql.federation:federation-graphql-java-support")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-jdk8")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-reactor")
 
     testImplementation("org.springframework.security:spring-security-core")
     testImplementation("io.projectreactor:reactor-core")
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
@@ -19,21 +19,14 @@ package com.netflix.graphql.dgs.internal
 import com.netflix.graphql.dgs.internal.method.ArgumentResolverComposite
 import graphql.schema.DataFetcher
 import graphql.schema.DataFetchingEnvironment
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.async
-import kotlinx.coroutines.future.asCompletableFuture
 import org.springframework.core.BridgeMethodResolver
+import org.springframework.core.CoroutinesUtils
 import org.springframework.core.KotlinDetector
 import org.springframework.core.MethodParameter
 import org.springframework.core.ParameterNameDiscoverer
 import org.springframework.core.annotation.SynthesizingMethodParameter
 import org.springframework.util.ReflectionUtils
-import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
-import java.util.concurrent.CompletableFuture
-import kotlin.reflect.full.callSuspend
-import kotlin.reflect.jvm.kotlinFunction
 
 class DataFetcherInvoker internal constructor(
     private val dgsComponent: Any,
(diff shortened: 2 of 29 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -116,17 +115,9 @@ class DataFetcherInvoker(
...
-                    method.kotlinFunction!!.callSuspend(dgsComponent, *args.toTypedArray())
-                } catch (exception: InvocationTargetException) {
-                    throw exception.cause ?: exception
-                }
-            }
-
-            launch.asCompletableFuture()
+        // Based on [org.springframework.web.reactive.result.method.InvocableHandlerMethod.invoke]
+        return if (KotlinDetector.isSuspendingFunction(method)) {
+            CoroutinesUtils.invokeSuspendingFunction(method, dgsComponent, *args.toTypedArray())
```
srinivasankavitha, line 120: Is this the main change that handles propagating the reactor context?

graphql-dgs/build.gradle.kts
```diff
@@ -35,6 +35,7 @@ dependencies {
     implementation("com.apollographql.federation:federation-graphql-java-support")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-jdk8")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-reactor")
```
srinivasankavitha, line 38: I think it's fine to add this new dependency.

berngp: @gnoeley , could you also run the dependency locks please?
```
./gradlew gL sL
```

berngp:
@gnoeley the #1013 PR refactored how we are resolving arguments and _invoking data fetchers_, I don't think the changes in the `DataFetcherInvoker` are needed anymore.

kilink:
> @gnoeley the #1013 PR refactored how we are resolving arguments and _invoking data fetchers_, I don't think the changes in the `DataFetcherInvoker` are needed anymore.

@berngp I think it still makes sense to switch to `CoroutinesUtils.invokeSuspendingFunction`, the only reason I hadn't done that already is that it requires pulling in a new dependency `kotlinx-coroutines-reactor`. If we're okay with that then it save us having to do similar stuff in `DataFetcherinvoker` to call a suspending function.

gnoeley:
Ok, couple things:
1. Looking at rebasing this on-top of #1013 today
2. I'll run the dependency lock 🔒
3. @berngp, the refactor would still not be propagating the `ReactorContext` to the coroutine as that is handled by the Kotlin extensions. Also, the return type also changes from a `CompletableFuture` to a `Mono` which causes the data processors to be applied from #1040 . Without that, the current refactor would still not fix the issue.
4. I need to update my commit author as it's not using my github.com name; please don't merge before then!

berngp:
Thanks for the explanation @gnoeley & @kilink , TIL.
I'll wait for your updates then @gnoeley .

graphql-dgs/build.gradle.kts
gnoeley: That's added updates for both `kotlinx-coroutines-reactor` and `spring-cloud-dependencies` which I'm assuming is expected?

gnoeley:
I think that's all points addressed if you'd like to take another look. Did attempt to split the dep.lock file updates into a separate set of commits (spring vs. kotlinx addition) but it was still 20+ files changed so didn't seem worth it in the end. 

Did simplify the test some by using the `MonoDataFetcherResultProcessor` and `ReactiveDgsContext` which has since been merged. Slightly more coupling though so, might prefer the anonymous `DataFetcherResultProcessor` used before?

berngp:
I would not worry on simplifying the test right now.

gnoeley:
> I would not worry on simplifying the test right now.

I'll do a final update using the anonymous class and remove the usage of `ReactiveDgsContext` again.

gnoeley:
> > I would not worry on simplifying the test right now.
> 
> I'll do a final update using the anonymous class and remove the usage of `ReactiveDgsContext` again.

That's done now btw

Issue #1006, opened 2022-04-24
feature request: Support multiple registered GraphQLContext contributors, allowing them to provide GraphQLContext key/value entries


theJC:
## Feature Request
In https://github.com/graphql-java/graphql-java/pull/2368, GraphQLContext was added to replace context, with the intent to allow code both in userspace and frameworks to contribute to, share and leverage the content in the GraphQLContext.

In https://github.com/Netflix/dgs-framework/pull/982, DGS has been adjusted to support using getGraphQlContext which allows the single customizable context builder to create the GraphQlContext.

This is a good first step.

There are several cases where customers would like to add additional information to the context that can then be referred by instrumentation and other downstream aspects, but unfortunately the custom context builder is a one-shot deal -- and I do not wish to override the way the that context is built in many applications that define their own DgsCustomContextBuilder, instead I'd like to add additional context to the GraphQlContext.

To properly support and leverage the intent of using GraphQLContext as a shared namespace for different pieces of information, an additional contract needs to be made available by customers of DGS to enable them to add/contribute to the context, from potentially different pieces of their framework code, without having to replace/override the single custom context builder, that often I find is defined by the final/consumer of said common frameworks.

## Describe Preferred Solution
I propose we add support to DGS to retrieve a set of GraphQLContext contributor classes (Spring Beans implementing a particular interface) that are invoked after the context builder has built the context, but before execution, so that they have the ability to provide additional contextual information (perhaps based on http headers, operationName, variables or extensions of the request) in the GraphQLContext under their own namespace.   

This contributor interface should have a method that accepts at least the same information that is passed to the custom context builders (see: [DgsCustomContextBuilderWithRequest.build](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsCustomContextBuilderWithRequest.kt#L28)), potentially also including the variables and operationName as well.

## Describe Alternatives
None come to mind at this point.

paulbakker:
I think this makes a lot of sense based on the new GraphQLContext. 

To help design this properly, could you describe some real-world use cases you have in mind that we could also use as test/example cases?

theJC:
I have a single real-world case in hand at the moment, to provide a single way to provide the proper support for federated tracing properly both for projects at my company who provided a custom context builder and for those projects that do not.

[...]

PR #1065, opened 2022-05-23
feature: Support multiple registered GraphQLContext contributors, allowing them to provide GraphQLContext key/value entries

theJC:
Changes in this PR
----
Implementation to add new feature as described in #1006 -- provide a mechanism to be able to register context contributors whom are able to add to the GraphQLContext and have access to the WebRequest/ServerRequest

Issue #1006


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/GraphQLContextContributor.kt
```diff
@@ -0,0 +1,30 @@
...
+import graphql.GraphQLContext
+import org.springframework.http.HttpHeaders
+import org.springframework.web.context.request.WebRequest
+
+interface GraphQLContextContributor {
+    fun contribute(
+        builder: GraphQLContext.Builder,
+        extensions: Map<String, Any>?,
+        headers: HttpHeaders?,
+        webRequest: WebRequest?
```
theJC, line 28: Anything else we should make available via this signature?
ie:
-         variables: Map<String, Any>,
-         extensions: Map<String, Any>?,
-         operationName: String?,

berngp: Aren't the headers available in the webrequest?

berngp:
I am wondering if we should invoke the context contributors when we know the query is valid. I.e an instrumentation will invoke the contributors when query validation has passed.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/GraphQLContextContributor.kt
theJC: Was just thinking that, that what I get from copying DgsCustomContextBuilderWithRequest ;)

theJC: Hmm.... I tried but then hit friction when looking at DefaultDgsQueryExecutor.getJsonResult ... that passes in headers but null for the webRequest....

theJC: https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutor.kt#L154-L155

theJC: Looks like I just need to add an additional signature executeAndExtractJsonPath on interface DgsQueryExecutor and its implementation

theJC: Done.

theJC: Anything else we should make available via this signature?
ie:

query: String
variables: Map<String, Any>
operationName: String?

theJC:
> I am wondering if we should invoke the context contributors when we know the query is valid. I.e an instrumentation will invoke the contributors when query validation has passed.

Query parsing consults the context for ParsingOptions.

One could anticipate that a GraphQLContext contributor may want to set parser options based on headers/caller/etc....

See: https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/ParseAndValidate.java#L58-L62

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/GraphQLContextContributor.kt
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * For each bean implementing this interface found, the framework will call the [build] method for every request.
+ * The [contribute] method is then able to use the [GraphQLContext.Builder] to provide additional entries to place in the context.
+ */
+interface GraphQLContextContributor {
+    fun contribute(
+        builder: GraphQLContext.Builder,
+        extensions: Map<String, Any>?,
+        webRequest: WebRequest?
```
theJC, line 30: Do you see value by adding additional items to this signature?
ie:

query: String
variables: Map<String, Any>
operationName: String?

graphql-dgs-example-java-webflux/src/test/java/ReactiveGraphQLContextContributorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.mock.web.reactive.function.server.MockServerRequest;
+import org.springframework.web.reactive.config.WebFluxConfigurationSupport;
+import reactor.core.publisher.Mono;
+
+import static com.netflix.graphql.dgs.example.context.MyContextContributor.CONTEXT_CONTRIBUTOR_HEADER_NAME;
+import static com.netflix.graphql.dgs.example.context.MyContextContributor.CONTEXT_CONTRIBUTOR_HEADER_VALUE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest(classes = {WebFluxConfigurationSupport.class, DgsAutoConfiguration.class, MovieDataFetcher.class, MyContextContributor.class, ExampleInstrumentation.class, DgsWebFluxAutoConfiguration.class, DgsPaginationAutoConfiguration.class})
```
theJC, line 36: See: https://github.com/Netflix/dgs-framework/issues/1066 for why WebFluxConfigurationSupport was added here

berngp:
The more I look at this the more I think we could implement the _context contributors_ as part of an Instrumentation and not as part of the `DgsQueryExecutor`.  The Instrumentation could be created such that it gets a ObjectProvider of _contributors_ and builds an ordered representation of such contributors via the `ObjectProvider`'s `orderedStream` API. The _contributors_ can then use the `@Ordered` annotations to express an order of precedence. I still think that we should invoke the _contributors_ after _validation of the query_ has passed. If we need to change the _Query Parser Options_ I would rather have a specific mechanism to override those, instead of doing it implicitly by a _context contributor_.

theJC:
@berngp - How do you recommend we make available the WebRequest/ServerRequest to said Instrumentation to then be able to be sent into any context contributors?

The primary use case that drove me to develop this capability was so that the context contributor(s) can have access to the headers on the request and based on the existance or value, be able to set something on the GraphQLContext.

berngp:
By the time the first instrumentation is being called the 'DgsContext' is already available vie the Instrumentation Environment. We could extract it from there, will get a poc. This will suggest that if we implement this feature via Instrumentations, we need to ensure the one that calls the context contributors executes before the others.

theJC:
@berngp  -- I've done as you suggested in https://github.com/Netflix/dgs-framework/pull/1145 Please take a look and verify you still feel that is a better direction to move towards, and if so, I'll close this PR in favor of that one.

> By the time the first instrumentation is being called the 'DgsContext' is already available vie the Instrumentation Environment. We could extract it from there, will get a poc. This will suggest that if we implement this feature via Instrumentations, we need to ensure the one that calls the context contributors executes before the others.

PR #1068, opened 2022-05-24
Wrap the DataLoaders after inspection for DgsDataLoaderRegistryConsumer

### LINKED ISSUE #1067: bug: `graphql-dgs-spring-boot-micrometer` Dataloader instrumentation does not work with `DgsDataLoaderRegistryConsumer` loaders
## Expected behavior

Instrumented Dataloaders implementing `DgsDataLoaderRegistryConsumer` in order to get a `DataLoaderRegistry` should have their `setDataLoaderRegistry` method called prior to execution.

## Actual behavior

`setDataLoaderRegistry` is never invoked on instrumented dataloaders. 

## Steps to reproduce

1. Create a dataloader that implements `DgsDataLoaderRegistryConsumer` and `setDataLoaderRegistry`.
```java
public class SomeDataLoader
    implements MappedBatchLoaderWithContext<String, ...>, DgsDataLoaderRegistryConsumer {
  // ...
  @Override
  public void setDataLoaderRegistry(DataLoaderRegistry dataLoaderRegistry) {
    this.dataLoaderRegistry = dataLoaderRegistry;
  }
}
```
2. Add in the instrumentation package into the pom.xml (I tested on 4.9.17, but after reading a bit of the code it appears to be an issue with future versions as well)
```xml
        <dependency>
            <groupId>com.netflix.graphql.dgs</groupId>
            <artifactId>graphql-dgs-spring-boot-micrometer</artifactId>
        </dependency>
```
3. Drop a breakpoint into the dataloader's load method (or otherwise add some application logic that requires the registry) and observe that `setDataLoaderRegistry` was never called.

## Mitigation

If you turn off DataLoader instrumentation with `dgs-graphql.data-loader-instrumentation.enabled=false` the data loader registry is set on the registry consumer as expected -- but you do lose out on dataloader metrics entirely

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

We need to test if a DataLoaders are an instance of `DgsDataLoaderRegistryConsumer`
before the reference is wrapped.

Fixes #1067

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/DgsDataLoaderInstrumentationProvider.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/DgsDataLoaderInstrumentationProvider.kt
@@ -29,7 +29,8 @@ class DgsDataLoaderInstrumentationProvider(
                     .to(BatchLoaderInterceptor(original, name, meterRegistrySupplier.get()))
             ByteBuddy()
                 .subclass(BatchLoader::class.java)
-                .method(ElementMatchers.named("load")).intercept(withBinders)
+                .method(ElementMatchers.named("load"))
+                .intercept(withBinders)
                 .make()
                 .load(javaClass.classLoader)
                 .loaded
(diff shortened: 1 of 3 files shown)


berngp:
As an action item we should improve the tests in the `graphql-dgs` module to account for _wrapped dataloaders_.

PR #1073, opened 2022-05-26
CSRF attack mitigation on requests with multipart/form-data, plain-text, or empty content-types

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

## TL;DR

* DGS will only accept requests with `content-type` of `application/json`, `application/graphql`, and `multipart/form-data`.
* DGS will enforce a _pre-flight_ header if the `content-type` is `multipart/form-data`. Acceptable _pre-flight_ headers are "x-apollo-operation-name",  "apollo-require-preflight", or "graphql-require-preflight" 

## Context

This commit addresses CSRF attacks that can leverage the execution of JS code attached to a
`content-type: multipart/form-data`, or other content-types which will not force the browsers to do a _preflight_ check
and enforce the CORS policy. **Application developers should provide a sensible CORS policy** as well as,
if they use cookies, a sensible cookie _SameSite_ policy.

DGS MVC supports the execution of GraphQL operations via HTTP POST requests with `content-type: multipart/form-data`.
Because they are POST requests, they can contain GraphQL mutations. Because they use `content-type: multipart/form-data`,
they can be "simple requests" which are not _preflighted_ by browsers.

Spring Boot applications using DGS that set `SameSite=None` cookies for authentication are then open to JS code from any origin can that can cause browsers to send `cookie-authenticated` Mutations to the GraphQL endpoint,
this will then be executed without checking your CORS policy first.  Although the attack won't be able to see the response to the mutation if your CORS policy is set up properly, the side effects of the mutation will still occur.

In addition, if the Spring Boot application using DGS relies on network properties for security (whether by explicitly
looking at the client's IP address or by only being available on a private network), then JS on any origin can cause browsers (which may be on a private network or have an allowed IP address) to send mutations to your GraphQL server,
which will be executed without checking your CORS policy first. (This attack does not require your server to use cookies.
It is in some cases prevented by some browsers such as Chrome.)

For additional context visit [Apollo Server 2 graphql-upload CSRF Page].

[Apollo Server 2 graphql-upload CSRF Page]: https://github.com/apollographql/apollo-server/security/advisories/GHSA-2p3c-p3qw-69r4

## Changes

With this commit DGS MVC will now enforce that all request targeting the
`/graphql` endpoint define their content type as either `multipart/form-data`, `application/json`, or `application/graphql`. this _limitation_ on supported content-types can be disabled, this is mentioned below).
In addition, if the content-type is `multipart/form-data`, `application/x-www-form-urlencoded`, or `text/plain`  there is the expectation that the request should also include one of the following _preflight_ headers:

* "x-apollo-operation-name"
* "apollo-require-preflight"
* "graphql-require-preflight"

This is *important* if you are using *file upload* since your clients will now have to send any of the headers mentioned above.

Although **not recommended** you can disable the _preflight_ check by setting `dgs.graphql.header.validation.enabled` to `false`.

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -31,6 +31,7 @@ import graphql.execution.reactive.SubscriptionPublisher
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpHeaders
+import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.RequestBody
@@ -69,18 +70,19 @@ import org.springframework.web.multipart.MultipartFile
 @RestController
 open class DgsRestController(
     open val dgsQueryExecutor: DgsQueryExecutor,
-    open val mapper: ObjectMapper = jacksonObjectMapper()
+    open val mapper: ObjectMapper = jacksonObjectMapper(),
+    open val dgsGraphQLRequestHeaderValidator: DgsGraphQLRequestHeaderValidator = DefaultDgsGraphQLRequestHeaderValidator()
 ) {
 
     companion object {
         private val logger: Logger = LoggerFactory.getLogger(DgsRestController::class.java)
-        private val GRAPHQL_MEDIA_TYPE = MediaType("application", "graphql")
     }
 
     // The @ConfigurationProperties bean name is <prefix>-<fqn>
+    // TODO Allow users to disable multipart-form/data
     @RequestMapping(
         "#{@'dgs.graphql-com.netflix.graphql.dgs.webmvc.autoconfigure.DgsWebMvcConfigurationProperties'.path}",
-        produces = ["application/json"]
+        produces = [MediaType.APPLICATION_JSON_VALUE]
     )
     fun graphql(
         @RequestBody body: String?,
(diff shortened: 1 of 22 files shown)


graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -222,3 +243,135 @@ open class DgsRestController(
...
+        @Suppress("MemberVisibilityCanBePrivate")
+        const val HEADER_X_APOLLO_OPERATION_NAME = "x-apollo-operation-name"
+        @Suppress("MemberVisibilityCanBePrivate")
+        const val HEADER_APOLLO_REQUIRE_PREFLIGHT = "apollo-require-preflight"
+        @Suppress("MemberVisibilityCanBePrivate")
+        const val HEADER_GRAPHQL_REQUIRE_PREFLIGHT = "graphql-require-preflight"
+
+        /**
+         * Our recommended set of CSRF prevention headers. Operations that do not provide a content-type such as `application/json`
+         * (in practice, this means GET operations) must include at least one of these headers.
```
glasser, line 313: This comment (possibly borrowed from Apollo Server :) ) is outdated given that multipart/form-data qualifies here too.

glasser (commented):
Thanks for doing this!

It's worth noting that Spring users may be a little more protected from at least the POST/mutation aspect of this than Node/Express users. That's because the default Spring CORS implementation [appears to actually reject requests which provided untrusted origins](https://github.com/spring-projects/spring-framework/blob/9e1d0057bc50ee3352e50b11da5008009dc3704a/spring-web/src/main/java/org/springframework/web/cors/DefaultCorsProcessor.java#L102-L122) whereas the Node cors middleware [does not do this](https://github.com/expressjs/cors/pull/110). This isn't a complete protection (there are ways to get browsers to send requests without an Origin header at all, though perhaps no ways in modern browsers to do this with POST requests) but it does help.

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
berngp: Thanks @glasser , will update the docs. The CSRF check is _heavily inspired_ by the Apollo Server implementation, and the comment was blatantly copied from it.

PR #1092, opened 2022-06-05
Addressing DgsGraphQLMetricsInstrumentation Latency increase

### LINKED ISSUE #1089: bug: Massive performance degradation since v4.10.4 if the graphql-dgs-spring-boot-micrometer module is active
Hey everyone, 

we recently updated DGS to v5.0.0 and noticed a massive increase in latencies. We did some testing and were able to target it down to the v4.10.4 release, so we downgraded to v4.10.3 to fix these latency issues in production.
After further investigation we figured out that the increased latencies are caused by the `graphql-dgs-spring-boot-micrometer`module. 
I created a simple [demo](https://github.com/hpuac/dgs-4-10-4-performance-degradation-demo) and measured the latencies. 
In the following table I averaged the latencies from 3 requests.

| DGS                                          | v4.10.2 | v4.10.3 | v4.10.4 | v5.0.0 | v5.0.1 |
|----------------------------------------------|---------|---------|---------|--------|--------|
| Without `graphql-dgs-spring-boot-micrometer` | 101ms   | 111ms   | 104ms   | 102ms  | 106ms  |
| With `graphql-dgs-spring-boot-micrometer`    | 145ms   | 147ms   | 842ms   | 867ms  | 864ms  |

It's clearly visible that there is a huge latency increase with `graphql-dgs-spring-boot-micrometer` enabled when updating from v4.10.3 to v4.10.4.
This shows best if the request contains many input arguments, so I'm sending 3000 numbers in my [test request file](https://github.com/hpuac/dgs-4-10-4-performance-degradation-demo/blob/main/request.sh#L13). 

## Expected behavior

With a release the latencies should not increase significantly.

## Actual behavior

With the update to v4.10.4 and later the latencies increase a lot if the `graphql-dgs-spring-boot-micrometer`module is active.

## Steps to reproduce

I created a repository with a minimal Sping + DGS setup to reproduce that issue:
https://github.com/hpuac/dgs-4-10-4-performance-degradation-demo

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [x] Build related changes

Changes in this PR
----

This commit addresses the increased latencies observed on DGS 4.10.4, 5.0.0, 5.0.1.
We are also adding support for JMH benchmarks across modules of DGS, that said we are only focusing in this commit on `graphql-dgs-spring-boot-micrometer`. To run benchmarks you can just call `./gradlew graphql-dgs-spring-boot-micrometer:jmh`.

Fixes #1089

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
@@ -15,8 +15,8 @@ import graphql.analysis.QueryVisitorFieldEnvironment
 import graphql.analysis.QueryVisitorStub
 import graphql.execution.instrumentation.*
 import graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted
+import graphql.execution.instrumentation.parameters.InstrumentationExecuteOperationParameters
 import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters
-import graphql.execution.instrumentation.parameters.InstrumentationExecutionStrategyParameters
 import graphql.execution.instrumentation.parameters.InstrumentationFieldFetchParameters
 import graphql.execution.instrumentation.parameters.InstrumentationValidationParameters
 import graphql.schema.DataFetcher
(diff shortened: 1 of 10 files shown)


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -177,18 +177,17 @@ class DgsGraphQLMetricsInstrumentation(
         }
     }
 
-    override fun beginExecutionStrategy(
-        parameters: InstrumentationExecutionStrategyParameters
```
srinivasankavitha, line 181: So switching to use InstrumenatationContext was teh issue?

berngp: Exactly

PR #1111, opened 2022-06-14
Switch to asynchronous response

1zg12:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

By switching to asynchronous response, I think this will help to
1. deliver the response faster. (instead of waiting for all data converted then respond, it could stream the response incrementally)
2. this could get rid of the OutOfMemory issue below

Issue #
https://github.com/Netflix/dgs-framework/discussions/1070

Alternatives considered
----

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -12,6 +12,9 @@
       <option name="IF_RPAREN_ON_NEW_LINE" value="false" />
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
+    <ScalaCodeStyleSettings>
+      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
+    </ScalaCodeStyleSettings>
     <codeStyleSettings language="kotlin">
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
       <option name="CALL_PARAMETERS_WRAP" value="0" />
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -12,9 +12,6 @@
       <option name="IF_RPAREN_ON_NEW_LINE" value="false" />
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
-    <ScalaCodeStyleSettings>
-      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
-    </ScalaCodeStyleSettings>
     <codeStyleSettings language="kotlin">
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
       <option name="CALL_PARAMETERS_WRAP" value="0" />
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -41,6 +41,7 @@ import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.context.request.WebRequest
 import org.springframework.web.multipart.MultipartFile
+import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody
 
 /**
  * HTTP entrypoint for the framework. Functionality in this class should be limited, so that as much code as possible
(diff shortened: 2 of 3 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -12,6 +12,9 @@
       <option name="IF_RPAREN_ON_NEW_LINE" value="false" />
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
+    <ScalaCodeStyleSettings>
```
berngp, line 15: I don't think you intended to add the _scala code style_ to a kotlin project.

berngp:
I think it is an interesting idea but I'm hesitant to enable this by default. According to [the docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/StreamingResponseBody.html), when using _ StreamingResponseBody_ ...

> it is highly recommended to configure explicitly the TaskExecutor used in Spring MVC for executing asynchronous requests. Both the MVC Java config and the MVC namespaces provide options to configure asynchronous handling. If not using those, an application can set the taskExecutor property of [RequestMappingHandlerAdapter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.html).

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -221,22 +242,29 @@ open class DgsRestController(
...
-                "Serialized JSON result in {}ms"
-            )
-        } catch (ex: InvalidDefinitionException) {
-            val errorMessage = "Error serializing response: ${ex.message}"
-            val errorResponse = ExecutionResultImpl(GraphqlErrorBuilder.newError().message(errorMessage).build())
-            logger.error(errorMessage, ex)
-            mapper.writeValueAsString(errorResponse.toSpecification())
-        }
 
-        return ResponseEntity.ok(result)
```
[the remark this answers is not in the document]
1zg12: I have a sample project in java with `StreamingResponseBody`, which seems like working as expected.

https://github.com/1zg12/stream/blob/d278d8ba182bad3237fb39ee37833a3626633038/src/main/java/com/example/stream/controller/RestController.java#L90

Did a screen capture as well: 
https://www.youtube.com/watch?v=MNfySp6bqDQ&ab_channel=5minutescounted

However, both are in java. I am not an expert on Kotlin, still trying with the syntax.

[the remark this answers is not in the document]
1zg12: it's working well to stream large data set as well
image

.idea/codeStyles/Project.xml
1zg12: Reverted.

1zg12:
> I think it is an interesting idea but I'm hesitant to enable this by default. According to [the docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/StreamingResponseBody.html), when using _StreamingResponseBody_ ...
> 
> > it is highly recommended to configure explicitly the TaskExecutor used in Spring MVC for executing asynchronous requests. Both the MVC Java config and the MVC namespaces provide options to configure asynchronous handling. If not using those, an application can set the taskExecutor property of [RequestMappingHandlerAdapter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.html).

I have that in the sample Java project. Have added in this PR as well. I think through that `Note`, Spring is likely suggesting to bulkhead the synchronous and asynchronous threads.

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/conf/WebMvc.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.netflix.graphql.dgs.mvc.conf
+
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
+import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+
+@Configuration
+open class WebMvc : WebMvcConfigurer {
```
berngp, line 25: Not convinced we should do this in the framework. This will also be problematic to the folks that want to explicitly set a different configuration. All in all, I don't think is a good idea for us to have it here.


Side note, the name of the configuration is not very expressive on what it is doing. T

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -91,23 +92,32 @@ open class DgsRestController(
         @RequestParam(name = "map") mapParam: String?,
         @RequestHeader headers: HttpHeaders,
         webRequest: WebRequest
-    ): ResponseEntity<String> {
+    ): ResponseEntity<StreamingResponseBody> {
+
```
berngp, line 96: I think, if we want to support this, is a feature that folks need to opt-in and not have by default. I.e. We could have a `DgsRestController` and a `DgsStreamingRestController`. People could opt in and then, as part of the process, define the `WebMvcConfigurer`'s `ThreadPoolTaskExecutor`.

berngp:
@1zg12 , thanks for doing all these work, as well as the sample project. I'm still hesitant on forcing folks to adopt this by default, folks will have to do something like what you do in your example [here](https://github.com/1zg12/stream/blob/d278d8ba182bad3237fb39ee37833a3626633038/src/main/java/com/example/stream/conf/WebConfig.java#L23-L50).

srinivasankavitha:
I agree with @berngp. I don't feel comfortable introducing this in the framework just yet.

Issue #906, opened 2022-03-04
feature: ability to customise the subscription endpoint


Ancient-Dragon:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue. Also consider discussing your idea on
the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first.

## Describe the Feature Request

I would like to be able to config the subscriptions endpoint. Currently it is only able to be: /subscriptions whereas the graphql endpoint can be configured using:
`
dgs:
  graphql:
    path: /api2/graphql
`

## Describe Preferred Solution

The preferred solution would be to a new property under graphql called subscription.path or something and then the subscription endpoint uses that. 

## Describe Alternatives

None

srinivasankavitha:
Thanks for the request. We will look into supporting this in a future release.

Ancient-Dragon:
Thank you!

PR #1117, opened 2022-06-21
feat: customise the subscription endpoint

TYzzt:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

add `dgs.graphql.websocket.path` to customise the subscription endpoint
Issue #906

Alternatives considered
----

### DIFF
--- a/graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
+++ b/graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
@@ -20,6 +20,5 @@ dependencies {
 
     implementation("org.springframework.boot:spring-boot-autoconfigure")
     implementation("org.springframework:spring-websocket")
-    implementation("jakarta.annotation:jakarta.annotation-api")
 
 }
--- a/graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
+++ b/graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
@@ -20,4 +20,6 @@ dependencies {
 
     implementation("org.springframework.boot:spring-boot-autoconfigure")
     implementation("org.springframework:spring-websocket")
+    implementation("jakarta.annotation:jakarta.annotation-api")
+
 }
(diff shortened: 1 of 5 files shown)


graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
```diff
@@ -20,4 +20,6 @@ dependencies {
 
     implementation("org.springframework.boot:spring-boot-autoconfigure")
     implementation("org.springframework:spring-websocket")
+    implementation("jakarta.annotation:jakarta.annotation-api")
```
berngp, line 23: Do we need this new dependency?

berngp:
Thanks for the PR @TYzzt , looks great! I'm just wondering if we really need the new dependency.

TYzzt:
> @berngp thanks reply! if there are more customIf there are more user-defined parameters, we can use `@PostConstruct` for easier verification,like https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcConfigurationProperties.kt
that's why need `jakarta.annotation:jakarta.annotation-api`

graphql-dgs-subscriptions-websockets-autoconfigure/build.gradle.kts
berngp: The project is being built with JDK 1.8, so we don't need this annotation to support `@PostConstruct`. I would argue that we should merge this PR without this dependency, and as a feature ticket, refactor the project such that we move to `"jakarta.annotation:jakarta.annotation-api"`. If you are having trouble building your project in the IDE just make sure it is using JDK 8 for the project.

berngp: In fact, I do wonder how things are working since it is known that using `PostConstruct` with an explicit dependency to `jakarta.annotation:jakarta.annotation-api` will not work. ref https://github.com/spring-projects/spring-boot/issues/23716

berngp: I correct myself, we are getting it via the `org.springframework.boot:spring-boot-starter`.

TYzzt: removed 'jakarta.annotation:jakarta.annotation-api' and build success! @berngp

berngp:
Thank you @TYzzt !

Issue #1105, opened 2022-06-10
bug: NPE in integration test when not using DgsAutoConfiguration


Georgian:
## Background
I'm updating DGS libraries from `4.9.25` to `5.0.2`, then running a `@SpringBootTest` integration test which uses `DgsQueryExecutor` directly to fire queries to the running test server.

## Expected behavior
No major code changes should be needed (hopefully).

## Actual behavior
I'm getting `java.lang.NullPointerException: null cannot be cast to non-null type org.springframework.web.context.request.NativeWebRequest` in `HandlerMethodArgumentResolverAdapter` (introduced in `4.10.3`) which handles my custom `@RequestHeader`; [more specifically at line 51](https://github.com/Netflix/dgs-framework/blob/a58aaaf08c2ff0e86bd763b845aae87a5301bc0e/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/internal/method/HandlerMethodArgumentResolverAdapter.kt#L51).

## Steps to reproduce
I've conveniently forked [this repo](https://github.com/Georgian/dgs-4-10-3-nativerequest-npe) from a previous bug. Simply run the `doMath` test.

With `@SpringBootTest(classes = [DgsAutoConfiguration::class])` it works fine.

## Workarounds
1. Downgrade to a pre - `4.10.3` version (hardly an option, since this is strictly a test problem)
2. Declare `DgsAutoConfiguration` class in test (inconvenient for me, because the test is HUGE and pulls in lots of components)

matejuh:
Same issue here. Makes `DgsQueryExecutor` unusable with newer DSG versions. It's possible to pass fake request, but not to all methods...

berngp:
Thanks again for reporting. We haven't been able to look into this issue yet.

fripoli:
is this fixed? I just used the sample fork mentioned by @Georgian with the latest version of the framework and the error is the same ...

```
java.lang.NullPointerException: null cannot be cast to non-null type org.springframework.web.context.request.NativeWebRequest
com.netflix.graphql.dgs.exceptions.QueryException: java.lang.NullPointerException: null cannot be cast to non-null type org.springframework.web.context.request.NativeWebRequest
```

AlexRiedler:
I worked around this locally using the following until it is fixed upstream (sorry for those that don't use kotlin)

```kt
inline fun <reified T> DgsQueryExecutor.executeAndExtractJsonPathAsObject(
[...]

PR #1153, opened 2022-07-15
fix: Use request as source for headers

matejuh:
After changes introduced in https://github.com/Netflix/dgs-framework/pull/1013 DgsQueryExecutor throws NPE. It does not use web layer, so request is missing. When custom request attributes eg. headers are needed they should be provided by specifically providing request which is not supported in all DgsExecutor methods. Also mixing using header in some parts and request in other parts is confusing. Request should be primary provider of headers and potentially other attributes which can be used to bind in DgsHandler.

fixes: https://github.com/Netflix/dgs-framework/issues/1105

Pull request checklist
----

- [ ] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Add posibility to add custom request to all `DgsQueryExecutor` methods.
`DgsContext` does not contain headers anymore, source for headers should be the request.

Alternatives considered
----

Move DefaultDgsGraphQLContextBuilder under interface or make `build` function open, so it would be possible to override it in custom implementations. It would be possible to use custom implementation in tests which would add a request. But this would work only for header not for other possible request params.

### DIFF
--- a/graphql-dgs-example-shared/src/main/java/com/netflix/graphql/dgs/example/shared/datafetcher/RequestHeadersDataFetcher.java
+++ b/graphql-dgs-example-shared/src/main/java/com/netflix/graphql/dgs/example/shared/datafetcher/RequestHeadersDataFetcher.java
@@ -19,16 +19,35 @@ package com.netflix.graphql.dgs.example.shared.datafetcher;
 import com.netflix.graphql.dgs.DgsComponent;
 import com.netflix.graphql.dgs.DgsData;
 import com.netflix.graphql.dgs.DgsDataFetchingEnvironment;
+import com.netflix.graphql.dgs.context.DgsContext;
+import com.netflix.graphql.dgs.internal.DgsWebMvcRequestData;
 import org.springframework.http.HttpHeaders;
 import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.context.request.WebRequest;
 
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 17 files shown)


srinivasankavitha:
Thanks for the PR. These changes are major breaking changes, so I'm a bit reluctant to proceed with this mechanism. We'll need some way of maintaining backwards compatibility with the current API.

matejuh:
I understand that these are backward incompatible changes... I was thinking about it and we can support original additional headers creating a custom request with these additional headers. Do you think it would be better? Currently we are not able to upgrade to newer dgs versions due to https://github.com/Netflix/dgs-framework/issues/1105 bug so I would like to solve the situation somehow. I the worst case we can move from `QueryExecutor` to `WebClient` but I would rather avoid that because it's only a workaround...

srinivasankavitha:
Yes, I think that is the approach we are leaning towards, i.e. creating a mock web request with the specified headers. 
cc @kilink  for context, since we were discussing the same changes.

Regarding being blocked on a fix, have you tried setting up the test with MockMVC, which would also handle creating of a mock web request?

srinivasankavitha:
FYI - @kilink is working on a fix for this. We will create a mock web request for this api when using it in tests.

matejuh:
OK thank you, I could update this PR. So I am closing it.

PR #1183, opened 2022-08-05
fix: GraphiQL may render broken and cannot be fully disabled

### LINKED ISSUE #1135: bug: Sometimes DGS_GRAPHQL_PATH is not set and it breaks the GraphiQL forcing me to redeploy the app
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue.

## Expected behavior
When I use the GraphiQL it should work

## Actual behavior
Sometimes, when I try to use the GraphiQL (Graphical Interface Web of GraphQL) it doesn't work.
When I watch the network tab it says 
/graphiql/%3CDGS_GRAPHQL_PATH%3E:1          Failed to load resource: the server responded with a status of 404 ()
Also when I try to perform a query or mutation I get this:
POST https://mysite.com/graphiql/%3CDGS_GRAPHQL_PATH%3E 404

## Steps to reproduce
I cannot reproduce 100% of the times, I think it's related with some order of execution or concurrency, because sometimes it happens and sometimes it's not. I'm deploying my application in an Elastic Beanstalk in AWS using CloudFormation for the infrastructure.
In localhost I couldn't reproduce it.
Until I not redeploy the app the problem continue there.

I'm using 5.0.5 version. It always happened with any version. I'm having this problem from time to time from the past year.

_Note: A test case would be highly appreciated, but we understand that's not always possible_

foo4u:
Pull request checklist
----

  first
- [x] Make sure to have sufficient test cases: tests cases already exist to cover not enabling GraphiQL for both MVC and WebFlux. However, the auto configuration for automatic resource discovery is not enabled on such tests. If you'd like this PR to be updated with static resource loading enabled I can do that.

Pull Request type
----

- [x] Bugfix

Changes in this PR
----

Fix a few issues with GraphiQL:
* Issue #1135
* Fix an issue where `dgs.graphql.graphiql.enabled` is not recognized as a valid configuration option with WebFlux
* Allow GraphiQL to be fully disabled, even if `spring.web.resources.add-mappings` it set to `true` (the default)

Alternatives considered
----

`DGS_GRAPHQL_PATH` is sometimes unset because there's a race condition between Spring's built in static resource handler and DGS' GraphiQL handlers in both Spring MVC and WebFlux. The alternative option to changing the path for `graphiql` would be to force applications to configure `spring.web.resources.add-mappings` to `false`; it defaults to `true`.

PR [#1167](https://github.com/Netflix/dgs-framework/pull/1167) from another contributor, also says it fixes #1135. However, it looks like a different potential issue related non-ASCII data in queries, rather than a fix for this bug.


foo4u:
@srinivasankavitha any way to get 👀 on this?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebfluxConfigurationProperties.kt
```diff
@@ -48,6 +48,7 @@ class DgsWebfluxConfigurationProperties(
      * Configuration properties for the GraphiQL endpoint.
      */
     data class DgsGraphiQLConfigurationProperties(
+        @DefaultValue("true") var enabled: Boolean = true,
```
srinivasankavitha, line 51: Where is this used?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/GraphiQlConfigurer.kt
```diff
@@ -35,7 +35,7 @@ class GraphiQlConfigurer(private val configProps: DgsWebfluxConfigurationPropert
         val graphiQLTitle = configProps.graphiql.title
         registry
             .addResourceHandler(configProps.graphiql.path + "/**")
-            .addResourceLocations("classpath:/static/graphiql/")
```
srinivasankavitha, line 38: For my understanding, could you explain why moving the location out of the static directory addresses the race condition?

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebfluxConfigurationProperties.kt
foo4u: It's used by Spring's [metadata generator](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#appendix.configuration-metadata.annotation-processor). The property isn't used directly by DGS b/c it's referred to in full dot notation as a `@Conditional`. However, it's included in the WebMVC equivalent of this class for Spring configuration autocompletion; see:

https://github.com/Netflix/dgs-framework/blob/8fa2368bf5e5ffb1a35a34e10e4bc328e02945d2/graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/DgsWebFluxAutoConfiguration.kt#L144-L147

graphql-dgs-spring-webflux-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webflux/autoconfiguration/GraphiQlConfigurer.kt
foo4u: > DGS_GRAPHQL_PATH is sometimes unset because there's a race condition between Spring's built in static resource handler and DGS' GraphiQL handlers in both Spring MVC and WebFlux. The alternative option to changing the path for graphiql would be to force applications to configure spring.web.resources.add-mappings to false; it defaults to true.

So, Spring attempts to serve all assets under `resources/static` as static files. This means the resources are served as-is from disk / JAR, and not processed by the DGS handler. Depending on bean registration order, the DGS handler may be registered after [Spring's static resource handler](https://docs.spring.io/spring-boot/docs/1.3.8.RELEASE/reference/html/boot-features-developing-web-applications.html#boot-features-spring-mvc-static-content).

From my understanding about Spring, it's best NOT to store templates that require server side rendering in `/resources/[static|public]`.

srinivasankavitha (approved):

PR #1188, opened 2022-08-12
Consume and produce bytes in DgsRestController

kilink:
Instead of reading the query and writing the response as Strings, use byte arrays instead.

Using Strings meant that Spring's StringHttpMessageConverter was used, which has a few
inefficiencies; for instance, it calculates the content length by encoding to the String
to a byte array.

Additionally, the Jackson documentation indicates that Strings are the least efficient input
and output format for ObjectMapper.


srinivasankavitha (approved):

berngp (approved):
Thanks!

Issue #966, opened 2022-04-06
bug: @DgsEntityFetcher does not delegate error handling correctly when using Mono


StefanVanDyck:
## Expected behavior
When implementing an entity fetcher using `@DgsEntityFetcher` I would expect all errors to be handled by our custom exception handler, implementing the `DataFetcherExceptionHandler` interface.

## Actual behavior
Errors are not delegated to the exception handler.
Unless the errors are first mapped to `InvocationTargetException`.

## Steps to reproduce
Implementing two entity-fetchers like so, should demonstrate the behavior.
This also happens when returning a CompletableFuture and using `toFuture()` on the stream before returning.
```
  @DgsEntityFetcher(name = "User")
  public Mono<User> entitiesUserNotHandled(Map<?, ?> values, DgsDataFetchingEnvironment dfe) {
    return Mono.error(new RuntimeException("TEST"))
  }
  
  @DgsEntityFetcher(name = "User")
  public Mono<User> entitiesUser(Map<?, ?> values, DgsDataFetchingEnvironment dfe) {
    return entitiesUserHandled(Map<?, ?> values, DgsDataFetchingEnvironment dfe)
        .onErrorMap(InvocationTargetException::new);
  }
```

StefanVanDyck:
Found the `onErrorMap` solution by stepping through the code and encountering this line:
https://github.com/Netflix/dgs-framework/blob/ac8c708352b3c58332c91bbe4feb18d3e45df078/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt#L127

But I'm not entirely sure about what I am looking at :)

srinivasankavitha:
Which version are you using? Have you tried using the latest release?

StefanVanDyck:
I upgraded to the latest(at this time) version: 4.9.25

denniseffing:
We encountered the same issue. Are there any known workarounds?

StefanVanDyck:
Adding the `.onErrorMap(InvocationTargetException::new)` like in the example above should work.

srinivasankavitha:
Is this an issue with the default exception handler or only when you are using a custom one? Because it is required to handle InvocationTargetException in the case of entities. I will close the issue in that case.

StefanVanDyck:
@srinivasankavitha The issue is that when using a custom exception handler, 
[...]

PR #1197, opened 2022-08-20
Make fed resolver always apply configured exception handler

raptros:
Pull request checklist
----

- [x] Consider creating a discussion on the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first

Pull Request type
----

- [x] Bugfix

Changes in this PR
----

Issue #966 

Ensures that a configured exception handler will be applied to exceptions that occurred within eg a Mono or CompleteableFuture execution.

The tests specifically for exception handling did not previously verify that the custom handler was called in all cases; added a check to verify it was called.

Had to make all of the tests in that suite provide a more fully configured DFE in order to get them to pass - there was an NPE in DefaultDataFetcherExceptionHandler when the ExecutionStepInfo wasn't present.

Alternatives considered
----

Considered trying to detect and special-case a not-fully configured DFE, but determined that the problem was the tests.

Considered not extracting causes from additional exception types, but after checking that CompletableFuture does wrap exceptions (and Mono does not), decided to add the extra extraction.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/federation/DefaultDgsFederationResolver.kt
@@ -40,6 +40,7 @@ import reactor.core.publisher.Mono
 import java.lang.reflect.InvocationTargetException
 import java.util.*
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionException
 import java.util.concurrent.CompletionStage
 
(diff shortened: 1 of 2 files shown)


srinivasankavitha:
Thanks for the PR contribution!

srinivasankavitha (approved):

raptros:
tyvm. I see that the checks have passed - what's the cycle for when yall merge things?

srinivasankavitha:
> 

We have a couple PRs coming up, so likely end of this week or early next week we'll do a release.

raptros:
Thank you!

Issue #1214, opened 2022-09-03
bug: Mocking of DataFetchers is broken in 5.2.1


nkonev:
After the introducing duplicate validator https://github.com/Netflix/dgs-framework/pull/1202
mock on DataFetchers stopped working.

Repro:
https://github.com/nkonev/dgs.issue.repro (5.1.1)

Update version from 5.1.1 to 5.2.1 in gradle, then run the tests again.

Example of output
```
Caused by: com.netflix.graphql.dgs.exceptions.InvalidDgsConfigurationException: Duplicate data fetchers registered for Query.car
```

nkonev:
Afterwords:
Here is very basic sample which is only demonstrating issue. Our use case is to provide mocks  for the one before last vertex in the long graph in order not to build all the chain but test the only two latest vertexes.

nkonev:
Presumably I've found that `MergedAnnotations.from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)` in DgsSchemaProvider returns both original and mock Methods
![Screenshot from 2022-09-05 23-20-17](https://user-images.githubusercontent.com/3160384/188504729-fe7e3090-7ecd-4c9e-b6b4-d14c7dd0132f.png)

PR #1202, opened 2022-08-26
Add duplicate data fetcher validation

willemmanuel:
Pull request checklist
----

- [x] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
- [ ] Consider creating a discussion on the [discussion forum](https://github.com/Netflix/dgs-framework/discussions)
  first
- [ ] Make sure the PR doesn't introduce backward compatibility issues
- [x] Make sure to have sufficient test cases

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Other (please describe):

Changes in this PR
----

During a refactor, my team recently introduced a bug to our GraphQL service by accidentally using a `@DgsData` annotation for the same field twice. I thought this might be a good validation to add when registering data fetchers to prevent undefined behavior. Throwing an `InvalidDgsConfigurationException` may affect backwards compatibility, so I'll defer to the team on whether this is acceptable!

PR #1216, opened 2022-09-05
Fix excess methods leading to schema duplicate in case mocked DataFecher

nkonev:
Fixes https://github.com/Netflix/dgs-framework/issues/1214 (appeared after https://github.com/Netflix/dgs-framework/pull/1202)

Checked with reproducer https://github.com/nkonev/dgs.issue.repro/tree/gh-1214

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -257,14 +257,13 @@ class DgsSchemaProvider(
         dgsComponents.forEach { dgsComponent ->
             val javaClass = AopUtils.getTargetClass(dgsComponent)
             ReflectionUtils.getUniqueDeclaredMethods(javaClass, ReflectionUtils.USER_DECLARED_METHODS).asSequence()
-                .filter { method ->
-                    MergedAnnotations
+                .map { method ->
+                    val mergedAnnotations = MergedAnnotations
                         .from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)
-                        .isPresent(DgsData::class.java)
+                    Pair(method, mergedAnnotations)
                 }
-                .forEach { method ->
-                    val mergedAnnotations =
-                        MergedAnnotations.from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)
+                .filter { (_, mergedAnnotations) -> mergedAnnotations.isPresent(DgsData::class.java) }
+                .forEach { (method, mergedAnnotations) ->
                     val filteredMergedAnnotations =
                         mergedAnnotations
                             .stream(DgsData::class.java)
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -62,6 +62,7 @@ import java.util.*
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 import java.util.concurrent.locks.ReentrantReadWriteLock
+import java.util.stream.Collectors
 import kotlin.concurrent.read
 import kotlin.concurrent.write
 
@@ -264,7 +265,12 @@ class DgsSchemaProvider(
                 .forEach { method ->
                     val mergedAnnotations =
                         MergedAnnotations.from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)
-                    mergedAnnotations.stream(DgsData::class.java).forEach { dgsDataAnnotation ->
+                    val filteredMergedAnnotations =
+                        mergedAnnotations
+                            .stream(DgsData::class.java)
+                            .filter { (it.source as Method).declaringClass == method.declaringClass }
+                            .collect(Collectors.toList())
+                    filteredMergedAnnotations.forEach { dgsDataAnnotation ->
                         registerDataFetcher(
                             typeDefinitionRegistry,
                             codeRegistryBuilder,


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -264,7 +265,12 @@ class DgsSchemaProvider(
                 .forEach { method ->
                     val mergedAnnotations =
                         MergedAnnotations.from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)
-                    mergedAnnotations.stream(DgsData::class.java).forEach { dgsDataAnnotation ->
+                    val filteredMergedAnnotations =
+                        mergedAnnotations
+                            .stream(DgsData::class.java)
+                            .filter { (it.source as Method).declaringClass == method.declaringClass }
```
srinivasankavitha, line 271: I'm not sure we want to just check if the duplicate data fetchers are in the same class. That would avoid cases where we actually have duplicates (non-mocks).

srinivasankavitha (changes requested):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
srinivasankavitha: Perhaps we should filter out mocks instead, if possible.

nkonev: > Perhaps we should filter out mocks instead, if possible.

a vice versa - we need mock themselves, because we mock datafetchers in order to get new (provided by mock) response

srinivasankavitha: Sorry, I was not clear. What I mean is we allow duplicates only if the other data fetchers are mocks.

nkonev: > I'm not sure we want to just check if the duplicate data fetchers are in the same class. That would avoid cases where we actually have duplicates (non-mocks).

I don't understand your point =( I added that check because `MergedAnnotations.from(method, MergedAnnotations.SearchStrategy.TYPE_HIERARCHY)` returns two annotations - annotation belonging to method of real class and annotation belonging to method of generated mock.
Although at step before `ReflectionUtils.getUniqueDeclaredMethods(javaClass, ReflectionUtils.USER_DECLARED_METHODS` returns **only** Method belonging to generated mock.

So doing filtering I filter out excess annotations of "real method"

srinivasankavitha: What I mean is the following desired behavior:
1. If we have multiple data fetchers, check if the duplicates include a mock
2. Throw an exception only when we have multiple real data fetchers
3. Disable the check and allow it if there is one real and the rest are mocks

Hope that clarifies things a bit better.

nkonev: One more question, so If I understood you correctly, you want to allow duplicates in case mocks.

Can this lead us to errors or any inconsistency in future, due to having duplicates when we run test and not having them in normal execution ?

srinivasankavitha: I think we should be ok. Based on the above conditional checks, only mock duplicates are allowed. So for tests that rely on real data fetchers, they should not be setting up mocks anyway.

srinivasankavitha: The issue with duplicates is that it is not guaranteed which data fetcher will be picked, btw. So tests need to be aware of that if mocking data fetchers. 

The other approach is to be strict like before and not even allow mocks to guarantee full consistency. That would be my strong preference generally.  But as you described in your use case, if your test set up requires mocks in addition to real data fetchers, this is the next best solution I can think of.

nkonev: Done @srinivasankavitha

srinivasankavitha: Thanks for the changes and iterating on this. I see the check for Mockito mock in the class name. I'm still not convinced this is needed though, and feels like we are unnecessarily introducing checks for the sake of tests. 

How is your test set up actually? Are you able to use the real data fetcher and just mock the calls within the data fetcher at a lower level?

I think we really do want to keep the duplicate data fetcher check across different classes, so ideally, it is the test that should be set up to accomadate this behavior.

nkonev: > feels like we are unnecessarily introducing checks for the sake of tests.

My previous solution was without those checks) And it should work correctly, because
> Although at step before ReflectionUtils.getUniqueDeclaredMethods(javaClass, ReflectionUtils.USER_DECLARED_METHODS returns only Method belonging to generated mock. So doing filtering I filter out excess annotations of "real method"


> How is your test set up actually?
It's described here https://github.com/Netflix/dgs-framework/issues/1214#issuecomment-1236009171

> Are you able to use the real data fetcher and just mock the calls within the data fetcher at a lower level? 
Also there https://github.com/Netflix/dgs-framework/issues/1214#issuecomment-1236009171

srinivasankavitha: Yeah, I did go through those comments earlier. It was not clear to me why you could not mock the calls inside the data fetcher (e.g. if you are making a call to a service etc.) and let all the data fetchers execute per normal.

In any case, I think after the discussion and the changes, your original changes might be better than checking for mocks.

nkonev: Thank you too)

So, speaking about the fix, what is my following step ? Should I remove the second commit ? Or will you just apply the first one ? Or something else ?

srinivasankavitha:
If you can remove the second commit, that would be great! Thanks!

srinivasankavitha (approved):

srinivasankavitha:
Thanks, looks good. If you could rebase when you get a chance, we can get this merged.

nkonev:
Done

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -264,7 +265,12 @@ class DgsSchemaProvider(
                 .forEach { method ->
                     val mergedAnnotations =
```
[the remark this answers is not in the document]
nkonev: Yes, its working.
I think I will update PR.

srinivasankavitha (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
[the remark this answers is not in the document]
ClaudenirFreitas: Cool, nice one!

timpamungkas:
Hello @srinivasankavitha , you might want to check this again.
I have mocks that works on `5.4.0` 

```
dependencyManagement {
    imports {
        mavenBom("com.netflix.graphql.dgs:graphql-dgs-platform-dependencies:5.4.0")
    }
}
```

But when change to `5.4.1` manually, or using platform bom
```
dependencyManagement {
    imports {
        mavenBom("com.netflix.graphql.dgs:graphql-dgs-platform-dependencies:latest.release")
    }
}
```

It failed.

```
aused by: com.netflix.graphql.dgs.exceptions.InvalidDgsConfigurationException: Duplicate data fetchers registered for Query.oneHello
	at app//com.netflix.graphql.dgs.internal.DgsSchemaProvider.registerDataFetcher(DgsSchemaProvider.kt:299)
	at app//com.netflix.graphql.dgs.internal.DgsSchemaProvider.findDataFetchers(DgsSchemaProvider.kt:273)
```

The test case is just simple hello world.

Thank you

PR #1232, opened 2022-09-14
Return specification Map from DgsRestController

kilink:
Instead of directly serializing the response to a String or bytes in DgsRestController, return the response as a Map as built by ExecutionResult.toSpecification. Doing this allows the Jackson HttpMessageConverter to serialize the result directly to the response's OutputStream.

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -227,8 +227,8 @@ open class DgsRestController(
         }
 
         return when (executionResult) {
-            is DgsExecutionResult -> executionResult.toSpringResponse(mapper)
-            else -> DgsExecutionResult.builder().executionResult(executionResult).build().toSpringResponse(mapper)
+            is DgsExecutionResult -> executionResult.toSpringResponse()
+            else -> DgsExecutionResult.builder().executionResult(executionResult).build().toSpringResponse()
         }
     }
 }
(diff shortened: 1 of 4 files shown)


graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
```diff
@@ -247,19 +244,6 @@ open class DgsRestController(
...
-            )
-        } catch (ex: InvalidDefinitionException) {
-            val errorMessage = "Error serializing response: ${ex.message}"
-            val errorResponse = ExecutionResultImpl(GraphqlErrorBuilder.newError().message(errorMessage).build())
-            logger.error(errorMessage, ex)
-            mapper.writeValueAsBytes(errorResponse.toSpecification())
-        }
-
-        return ResponseEntity(result, responseHeaders, HttpStatus.OK)
+        return ResponseEntity.ok().headers(responseHeaders).body(executionResult.toSpecification())
```
berngp, line 247: @srinivasankavitha what do you think about this change?
> Instead of directly serializing the response to a String or bytes, return the response as a Map as built by
> ExecutionResult.toSpecification. Doing this allows the Jackson HttpMessageConverter to serialize the result directly to the response's OutputStream.

berngp:
@kilink, I rebased against the latest master, could you please take a look?

graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
srinivasankavitha: Sorry for the delayed review. Looks good to me!

srinivasankavitha (approved):

Ancient-Dragon:
Hi @srinivasankavitha  / @berngp  any chance of this being merged in? We suspect this is causing an out of memory exception for us.

kilink:
@Ancient-Dragon I've rebased / merged this PR, it will be included in the next release.

Issue #1058, opened 2022-05-19
bug: Some invalid queries get exposed as "unknown" or "Intenal" instead of "ValidationError" or "BadRequest" in micrometer


hpuac:
## Expected behavior

When modifying the GraphQL request and sending a variable that is not in the query, I would expect it to get exposed as `ValidationError` or `BAD_REQUEST` in micrometer.

## Actual behavior

- When modifying the GraphQL request and sending a variable that is not in the query, it gets exposed as errorType `unknown` in micrometer. 
- When modifying a federated GraphQL request and sending a variable that is not in the query, it gets exposed as errorTyoe `INTERNAL` in micrometer.

This is preventing me to exclude client errors from our alerting that we base on these metrics.

## Steps to reproduce

I think the behaviour description is pretty self explanatory. Please let me know if something is not clear.
But I'd like to provide some additional information.
- When the error type "unknown" is exposed, the exception that `DgsGraphQLMetricsInstrumentation->instrumentExecutionResult` receives is `graphql.execution.InputMapDefinesTooManyFieldsException: The variables input contains a field name 'myDemoKey' that is not defined for input object type 'MyDemoInput' `. The `ErrorUtils.sanitizeErrorPaths()` call then sets the type to `unknown`.
- When the error type `INTERNAL` is exposed,  the exception that `DgsGraphQLMetricsInstrumentation->instrumentExecutionResult` receives is `TypedGraphQLError{message='java.lang.IllegalStateException: Expected key 'myDemoKey' in received federated values map does not exist.', locations=[], path=[_entities], extensions={errorType=INTERNAL}}`. The `ErrorUtils.sanitizeErrorPaths()` call then sets the type to `INTERNAL`.

berngp:
Thanks @hpuac, it should be consistent with `BAD_REQUEST` in both cases.

srinivasankavitha:
Closing this issue since it should be addressed in the latest release. Please reopen, if issue exists.

hpuac:
Hey @srinivasankavitha, sorry for the delay but I was now able to re-test it with DGS 5.0.5.
The following errors are still not exposed to micrometer as expected:
- Exposed as `INTERNAL`: `{"message":"com.netflix.graphql.dgs.exceptions.MissingFederatedQueryArgument: The federated query is missing field(s) __typename","locations":[],"path":["_entities"],"extensions":{"errorType"
:"INTERNAL"}}`
[...]

PR #1248, opened 2022-09-29
#1058: Correct error types for bad user input in micrometer

antholeole:
Pull request checklist
----

Pull Request type
----

- [X] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
In general I'd recommend reviewing commit by commit.

- `MissingFederatedQueryArgument` reported unknown to micrometer (fix [here](https://github.com/Netflix/dgs-framework/commit/cb1cde640045839298367e6b5f04087b88342e28)):  `MissingFederatedQueryArgument` and a few other Dgs exceptions now inherit from `DgsException`: if we catch a `DGSException`, we know that it is typed properly, so we don't have to special case on a bunch of exception types

- `The query is null or empty.` is not getting reported to micrometer (fix [here](https://github.com/Netflix/dgs-framework/commit/461ffb95126cb71d91fca4720b8b0e4d4b89f8e1)): this one was a bit tricky, since `BaseDgsQueryExecutor` would short circuit on an empty query with a custom error. I opted to remove this logic because an empty query is effectively the same thing as a query with just a single character but would turn up completely different errors; if we remove this special case logic, the error falls through, micrometer reports it and the user sees a `SyntaxError`. lmk if there is a better way to handle this (I tried manually passing to instrumentation but it was very clunky)
 
- `ValidationError` reporting as unknown to micrometer (fix [here](https://github.com/Netflix/dgs-framework/commit/7a5dddbd2baf9cbd0d84da7ba2c7deb244d657f0)): Special cased it

- Re-added micrometer to example project (fix [here](https://github.com/Netflix/dgs-framework/commit/412f7eeac12db2adc296cf8f1e5ac3abd5b4f829)): as @srinivasankavitha and I discussed, this should be re-added

lmk thoughts and additional test cases. The old test cases covered most of these changes, the only new feature is the `DGSException` which I added a test case for.


berngp:
Thanks for the PR. The idea of introducing a `DgsException` makes sense and simplifies the code handling them quite a bit in the [DefaultDataFetcherExceptionHandler.kt](https://github.com/Netflix/dgs-framework/pull/1248/files#diff-cbec39732fd94c71343fc88a8a9a2c97e7fc9946338e25f0bfc20ee0e8b16e8e). What I'll do is merge this PR in the [feature/refactor-metrics](https://github.com/Netflix/dgs-framework/tree/feature/refactor-metrics) branch and look at the failing text.

antholeole:
Hey! Sorry about the failing tests - I was going to take time to look at them but got caught up with midterms. If you want to take it over that’s fine, but it’s on the agenda for me. On Oct 10, 2022, at 4:19 PM, Bernardo Gomez Palacio ***@***.***> wrote:﻿
Merged #1248 into feature/refactor-metrics.

berngp:
@antholeole, the branch [feature/refactor-metrics](https://github.com/Netflix/dgs-framework/tree/feature/refactor-metrics) is based in your PR with additional small changes. There is a caveat around how we return errors when the request doesn't have a `query`. After discussing this with @srinivasankavitha, we both agree that the best approach is to change the code, in this case the `BaseDgsQueryExecutor` to return a known exception that will be caught by the `DgsRestController` and the `DefaultDgsWebfluxHttpHandler` to return 400. The response should follow the GraphQL standard of returning the following payload

```
{
                  "errors":[
                    {
                      "message": "GraphQL operations must contain a non-empty `query`.",
                      "locations":[],
                      "extensions": {
                        "errorType":"BAD_REQUEST"
                      }
                    }
                  ]
} 
```

How does it impact metrics?
If the request doesn't have a `query` it will not be captured as part of the `gql.query` metrics, since it is not really a query. That said, if we do return a 400, it will properly show in the Spring MVC and Spring Webflux [metrics](https://docs.spring.io/spring-boot/docs/2.6.x/reference/htmlsingle/#actuator.metrics.supported.spring-mvc).

If you want, and please let us know, you could implement the changes in the `BaseDgsQueryExecutor`, `DefaultDgsWebfluxHttpHandler` and `DgsRestController` to return a 400. This will impact the tests in both the`MalformedQueryContentTest.kt` used by webfux and mvc respectively. 

Please feel free to
1. checkout the `feature/refactor-metrics` branch
2. use that as your base branch
3. squahs your changes
4. send a PR targeting master.

ref.
https://github.com/Netflix/dgs-framework/compare/feature/refactor-metrics?expand=1

antholeole:
Sure, I'll take that approach - seems the most correct and clever. I'll try to have a PR submitted by the end of week, perhaps earlier

berngp:
Thank you!

Issue #1083, opened 2022-05-31
bug: Complex InputArgument serialized as Optional List of Complex Type


berngp:
## Context

Given the following GraphQL Schema...

```
type Query {
    hello(person:[Person]): String
}
            
input Person {
    name:String
}
```

And a _Data Fetcher_ that serializes to a `Optional<List<Person>>`, as below

```
@DgsData(parentType = "Query", field = "hello")
fun someFetcher(@InputArgument person: Optional<List<Person>>): String {
    return "Hello, ${person.get().joinToString(", ")}"
}
```

### Expectation

I should see that the optional arguments have a list of `Person` references, assuming we have a simple `Person` pojo.

### Actual behavior

The following error ensues
```
java.util.LinkedHashMap cannot be cast to com.netflix.graphql.dgs.internal.kotlin.test.Person
```

## Steps to reproduce

Test case.

```
    @Test
    fun `@InputArgument on an optional list of input types, with no input argument name`() {
        val schema = """
            type Query {
                hello(person:[Person]): String
            }
            
            input Person {
                name:String
            }
        """.trimIndent()

        val fetcher = object {
            @DgsData(parentType = "Query", field = "hello")
            fun someFetcher(@InputArgument person: Optional<List<Person>>): String {
                assertThat(person)
                    .isNotEmpty
                    .get()
                    .extracting { ls -> ls.map { it.name } }.asList()
                    .containsOnly("tester 1", "tester 2")
                return "Hello, ${person.get().joinToString(", ")}"
            }
        }

        withComponents("helloFetcher" to fetcher)

        val build = GraphQL.newGraphQL(provider.schema(schema)).build()
        val executionResult = build.execute("""{hello(person: [{name: "tester 1"}, {name: "tester 2"}])}""")!!

        assertThat(executionResult).isNotNull
        assertThat(executionResult.errors).isEmpty()
        assertThat(executionResult.isDataPresent).isTrue

        val data = executionResult.getData<Map<String, *>>()
        assertThat(data["hello"]).isEqualTo("Hello, tester 1, tester 2")
    }

```

k2k1422:
Hi @berngp i was tracking this bug more then a month. Is it going to be fixed in next releases.

k2k1422:
Hi @berngp 
can you help me how to resolve this issue, or by pass with any other thing.
I have using this optional field for a pojo. 

if we don't have solutions is we are going to resolve in subsequent releases

berngp:
[...]

PR #1254, opened 2022-10-01
BUG FIX (#1083): Complex InputArgument serialized as Optional List of Complex Type 

antholeole:
Pull request checklist
----

  first
- [X] Make sure to have sufficient test cases (pre-covered)

Pull Request type
----

- [X] Bugfix

Changes in this PR
----
fix for https://github.com/Netflix/dgs-framework/issues/1083

Bug: The type of the nested collection was getting unbounded to `List<*>` (from, say `List<Person>`) when `.toClass` was called on it so it triggered the exit condition on the recursion (since `List<Anything> == List<*>`)

Solution: Instead of calling `.toClass`, we should just use the constructor to create `TypeDescriptor` from `ResolveableType` and then Spring does all the heavy lifting 🥳🥳🥳

Also the test cases were a little off (they were printing the whole `Person` object not just name) but thanks for adding them, very helpful


antholeole:
CC @srinivasankavitha @paulbakker: Any way to get 👀 on this? small PR. tests were already pre-written by @berngp, just made them pass.

srinivasankavitha (approved):

Issue #1278, opened 2022-10-18
bug: "extensions" exposed in valid graphQL request 


antholeole:
## Steps to reproduce

Run the default DGS app with the following query:

```graphql
query {
  hello
}
```

or a mutation:

```graphql
mutation {
  addRating(input: {
    title: "Dahmer",
    stars: 1 # 2 scary 4 me
  }) {
    avgStars
  }
}
```

## Expected behavior

the server should respond with:

```json
{
  "data": {
    "hello": "hello, Stranger!"
  }
}
```

or 

```json
{
  "data": {
    "addRating": {
      "avgStars": 1
    }
  }
}
```

## Actual behavior

the server responds with:

```json
{
  "data": {
    "hello": "hello, Stranger!"
  },
  "extensions": {}
}
```

or:

```json
{
  "data": {
    "addRating": {
      "avgStars": 1
    }
  },
  "extensions": {}
}
```

I ran a `git bisect` and narrowed it down to commit `3cab92c3172128b37c277e89d491df3fe108bc12`. I can try to fix it after I'm done with `refactor-metrics`, unless someone beats me to it.

I didn't test for subscriptions but it's probably happening there too.

antholeole:
Actually looking back at this, this is not really a bug - it's leftover [from DGS_RESPONSE_HEADERS_KEY](https://github.com/Netflix/dgs-framework/commit/3cab92c3172128b37c277e89d491df3fe108bc12#diff-13e2b14fd77f54bc584e10e2f751a23820d20efc08b34a0d4e2abe08aee01154R246) getting removed from extensions.

I'll leave this open if you want me to fix it during my next PR, otherwise I'll close. Test case is easy, just add this to the test added during that PR:

```kotlin
        ...
        assertThat(result.headers["myHeader"]).contains("hello")

        assertThat(result.body).isInstanceOfSatisfying(ByteArray::class.java) { body ->
            val mapper = jacksonObjectMapper()
            val (_, _, extensions) = mapper.readValue(body, GraphQLResponse::class.java)
            assertThat(extensions).isNull()
        }
```

paulbakker:
Yeah it’s an unexpected side effect of my change. Would be good to fix for
sure.

PR #1280, opened 2022-10-19
[Bugfix] Refactor Metrics

antholeole:
Pull request checklist
----

  first

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

- Added `DgsExecutionResult`.  This makes it easy for developers to change the http status code, as well as add headers to the response should they wish do that (see the [example](graphql-dgs-example-java/src/main/java/com/netflix/graphql/dgs/example/datafetcher/MyInstrumentation.java)) 
- As per [this](https://github.com/Netflix/dgs-framework/pull/1248#issuecomment-1279530956), an empty query now returns a 400 response with the Apollo specification empty query message.
- Fixed a bug where `extensions` was being left on the response (https://github.com/Netflix/dgs-framework/issues/1278)
- Bunch of tests for everything above

This now does not break adding headers through the extension method due to [this](https://github.com/Netflix/dgs-framework/pull/1280/commits/8580ecde7b5893573bb33823b5d63b85845c5dde) change. Let me know if this is necessary or not, considering it was added less than a week ago and probably does not have wide adoption yet (not sure about internally).

In general, again, I'd recommend reviewing commit by commit - the initial commit (b352287) was updated by @/berngp so I'm assuming it passed review - all the other commits are my changes and are hopefully descriptive .

lmk thoughts.


antholeole:
need to clean this up a bunch but want to solicit feedback first...

thoughts on [DGSExecutionResult](graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsExecutionResult.kt)? It cleans up a bunch of stuff. 

- [adding a header through instrumentation is a little cleaner](https://github.com/Netflix/dgs-framework/pull/1280/files#diff-361a9d9e2e43484c6938d295b22f9331fba8e856dda4814150ca98d7ceb6c894L33-R38) and doesn't rely on a magic string or the extensions at all. Probably a matter of taste, as well as breaks the API so this might be overall not good.
- It also cleans up a ton of the logic in [DGSRestController](https://github.com/Netflix/dgs-framework/pull/1280/files#diff-13e2b14fd77f54bc584e10e2f751a23820d20efc08b34a0d4e2abe08aee01154L229-L261) and makes it easy for us to return custom headers / http status codes in the future. (as a side effect this solves https://github.com/Netflix/dgs-framework/issues/1278
- For instance, it makes [changing the http status code for empty query trivial](https://github.com/Netflix/dgs-framework/pull/1280/files#diff-45ffe682b2bcbc9136408be260c0d00232eab8604c830df91e5061c31a8e51c1R81-R90)

let me know your thoughts. I understand that it's another larger change and breaks the API for adding headers through the extensions (although it can maybe be monkey patched in? something like `if (has extension) make execution result w/ header` for backwards compatibility) so I understand it might be a tough sell.

berngp:
@antholeole looks good to me, could you squash changes and rebase from master. Thanks for the PR and thanks again for introducing the `DgsExecutionResult`.

berngp (approved):

antholeole:
@berngp rebased, instead of squashing my branch can you click "Squash and merge" when merging?

PR #1298, opened 2022-10-27
Rework the API expressed by DgsExecutionResult

berngp:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

* We are adopting a Builder pattern instead of depending on Kotlin delegation for the ExecutionResult. This was done to expose a cleaner API for Java Consumers.
* We are also moving the `DgsExecutionResult` to not be _internal_, since at is clear from the `MyInstrumentation` example developers will be using it explicitly.


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsExecutionResult.kt
```diff
@@ -133,4 +86,56 @@ class DgsExecutionResult @JvmOverloads constructor(
...
+     * Facilitate the construction of a [DgsExecutionResult] instance.
+     */
+    class Builder {
+        var executionResult: ExecutionResult = DEFAULT_EXECUTION_RESULT
+            private set
+        fun executionResult(executionResult: ExecutionResult) = apply { this.executionResult = executionResult }
+
+        fun executionResult(executionResultBuilder: ExecutionResultImpl.Builder) = apply { this.executionResult = executionResultBuilder.build() }
+
+        var headers: HttpHeaders = HttpHeaders()
```
kilink, line 121: Can this default to `HttpHeaders.EMPTY`, so it doesn't need to instantiate anything if the builder method isn't called?

berngp: It should, thanks @kilink !

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsExecutionResult.kt
```diff
@@ -14,86 +14,39 @@
...
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 
-class DgsExecutionResult @JvmOverloads constructor(
+class DgsExecutionResult(
     val executionResult: ExecutionResult,
-    val headers: HttpHeaders = HttpHeaders(),
-    val status: HttpStatus = HttpStatus.OK
+    val headers: HttpHeaders,
```
kilink, line 34: `HttpHeaders` is exposed via a public getter, should we wrap it with `HttpHeaders.readOnlyHttpHeaders` so we're not exposing mutable state?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsExecutionResult.kt
```diff
@@ -14,86 +14,39 @@
...
 import graphql.GraphqlErrorBuilder
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 
-class DgsExecutionResult @JvmOverloads constructor(
+class DgsExecutionResult(
     val executionResult: ExecutionResult,
```
kilink, line 33: I wonder if this property needs to be public, we are already delegating all methods to it

kilink (approved):

srinivasankavitha (approved):

antholeole:
Ah, I didn't consider these things, sorry! These are good changes, thanks for the refactor.

PR #1302, opened 2022-11-04
Fix mapping with Kotlin default arguments

kilink:
Add support for mapping to Kotlin objects with default parameters when not all arguments are specified in DefaultInputObjectMapper.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
@@ -16,44 +16,64 @@
 
 package com.netflix.graphql.dgs.internal
 
-import com.fasterxml.jackson.module.kotlin.isKotlinClass
 import com.netflix.graphql.dgs.exceptions.DgsInvalidInputArgumentException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.core.KotlinDetector
+import org.springframework.util.CollectionUtils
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
```diff
@@ -16,44 +16,64 @@
...
+
+        val parameters = constructor.parameters
+        val parametersByName = CollectionUtils.newLinkedHashMap<KParameter, Any?>(parameters.size)
+
+        for (parameter in parameters) {
+            if (parameter.name !in inputMap) {
+                if (parameter.isOptional) {
+                    continue
+                } else if (parameter.type.isMarkedNullable) {
+                    parametersByName[parameter] = null
```
kilink, line 53: This preserves the existing behavior (as exercised in the tests), but I am not sure this is desirable. It essentially means that we fill in nulls for classes like this when the argument isn't explicitly supplied:

```data class Foo(val nullableParam: String?)```

I would think that if that were the desired behavior, the user would write the class like:

```data class Foo(val nullableParam: String? = null)```

Anyway, I've left it here so as to not break the existing behavior.

srinivasankavitha: Yeah, I think this has been a point of contention in the codegen plugin as well. But I agree with the decision to preserve existing behavior.

srinivasankavitha (approved):
Looks good to me!

PR #1313, opened 2022-11-09
Make session security context available for both WebSocket protocols

HuseinJ:
Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Initial Issue see: https://github.com/Netflix/dgs-framework/pull/814 - This fixes a bug where we do not receive the SecurityContext when using `WebsocketGraphQlTransportWSProtocoll`. The logic which is used in `WebsocketGraphQLWSProtocolHandler` where the SecurityContext is loaded from the session, should be called regardless of which protocol is used. So i moved the logic up again to `DgsWebSocketHandler` where it was before so it gets called

Alternatives considered
----


srinivasankavitha (approved):
Thanks for the fixes!

HuseinJ:
@scon-io I think it should work. Unfortunately these changes are not released yet - You could try it out by building it locally.

srinivasankavitha:
We will be doing a release around end of next week, including these changes. Sent from my iPhoneOn Nov 24, 2022, at 1:17 AM, husein-ftapi ***@***.***> wrote:﻿
@scon-io I think it should work. Unfortunately these changes are not released yet - You could try it out by building it locally.

HuseinJ:
@srinivasankavitha Thank you very much for the info :-) I'm happy to hear that this will soon be released 😄

PR #1377, opened 2022-12-28
feat: allow users to override subscription handling

Ancient-Dragon:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):
Making method no-longer private
Changes in this PR
----
Our team would like to customise how subscriptions are handled, by extending the functionality offered in the SSE handler.

Issue # N/A

Alternatives considered
----
Duplicating the handler, but this could easily get out of date etc, didn't seem any harm in allowing users to override if required.


Ancient-Dragon:
@srinivasankavitha, please could you review this PR? Would be quite helpful for my team.

srinivasankavitha:
Thanks @Ancient-Dragon. Hwwever, have you read the class comments? We do not intend for users to extend this class since it is not an API. 
```
This class is defined as "open" only for proxy/aop use cases. It is not considered part of the API, and backwards compatibility is not guaranteed.
 * Do not manually extend this class.
 */```
 
 We would like to preserve the flexibility of changing this class without breaking backwards compatibility.

PR #1384, opened 2023-01-09
Use GraphQLContext instead of DgsContext for dataloaders

fotoetienne:
# Use GraphQLContext instead of DgsContext for dataloaders

As of graphql-java version [17](https://github.com/graphql-java/graphql-java/releases/tag/v17.0), [GraphQLContext](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/GraphQLContext.java) is now the approved context mechanism (replacing the previously used opaque Context) allowing passing of pertinent context via key/value pairs to provide frameworks and userspace code the capability to contribute, share and leverage various pieces of context independent of each other.

The [GraphQLContextContributor](https://netflix.github.io/dgs/advanced/graphqlcontext-leveraging/) interface allows DGS users to leverage GraphQLContext. However, GraphQLContext is current inaccessible from within DataLoaders. This PR make GraphQLContext available within dataloaders.

**Example Usage:**
```java
@DgsDataLoader(name = "exampleLoaderWithGraphQLContext")
public class ExampleLoaderWithGraphQLContext implements BatchLoaderWithContext<String, String> {
    @Override
    public CompletionStage<List<String>> load(List<String> keys, BatchLoaderEnvironment environment) {
        GraphQLContext graphQLContext = environment.getContext();
        return CompletableFuture.supplyAsync(() -> keys.stream().map((Function<String, String>) graphQLContext::get).collect(Collectors.toList()));
    }
}
```

*Note:* Previously, the DGS framework passed DgsContext to dataloaders as context. CustomContext is contained in DgsContext, and generally retrieved with a static helper. example:
```java
MyContext context = DgsContext.getCustomContext(environment);
```
The helper `DgsContext::getCustomContext` is able to pull MyContext from GraphQLContext, so this is non-breaking for users that employ the recommended helper method. However, this PR is potentially a breaking change for any user code that coerces dataloader context to DgsContext manually.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
@@ -33,6 +33,7 @@ import graphql.ExecutionInput
 import graphql.ExecutionResult
 import graphql.ExecutionResultImpl
 import graphql.GraphQL
+import graphql.GraphQLContext
 import graphql.GraphQLError
 import graphql.execution.ExecutionIdProvider
 import graphql.execution.ExecutionStrategy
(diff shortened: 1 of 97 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
```diff
@@ -108,23 +109,22 @@ object BaseDgsQueryExecutor {
         idProvider.ifPresent { graphQLBuilder.executionIdProvider(it) }
 
         val graphQL: GraphQL = graphQLBuilder.build()
-
-        val dataLoaderRegistry = dataLoaderProvider.buildRegistryWithContextSupplier { dgsContext }
+        val graphQLContextFuture = CompletableFuture<GraphQLContext>()
+        val dataLoaderRegistry = dataLoaderProvider.buildRegistryWithContextSupplier { graphQLContextFuture.get() }
```
srinivasankavitha, line 113: The changes look fine. I'm more concerned about the breaking change, and we do seem to jave quite a few usages that assume the type of context available in the `BatchLoaderEnvironment` is a `DgsContext`. One potential way to avoid breaking users, is to check whether `GraphQlContextContributors` is set in the BaseDgsQueryExecutor, and setting the context in the DataLoaderProvider accordingly. Then existing users will still get a DgsContext and the folks using the newer GraphQLContext will get that instead.

This assumes the user is choosing between either the `DgsContext` mechanism or using the newer `GraphQLContextContributor` mechanism to set up the custom context and not a mix of both approaches.

We can then get rid/deprecate the DgsContext in a future breaking release. Thoughts?

srinivasankavitha (changes requested):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
fotoetienne: That's a good idea. That should significantly reduce the chance of breaking existing code.

fotoetienne: Or what about conversely checking whether dgsContext.customContext is set?

srinivasankavitha: That would work, assuming it is not always set up with some default.

srinivasankavitha: Actually, not. Folks may be accessing the dgs context in the batch loader environment even if they aren't setting up a custom context,

srinivasankavitha: We discussed this PR offline and concluded we will make this a breaking change in DGS `6.0.0` release along with Spring Boot 3.0 update instead of trying to make it backwards compatible.

paulbakker (approved):

PR #1388, opened 2023-01-14
Use GrahpiQL with Explorer extensions instead of vanilla version (#1387)

armaniehs:
- Feature

Use GrahpiQL with Explorer extensions instead of vanilla version


srinivasankavitha (approved):
Thanks for the PR. The changes look good and we will test and merge in the next week.

PR #1403, opened 2023-01-29
Support default parameters for DgsData methods

kilink:
Add support for Kotlin default parameter values for DgsData methods.

This is a follow up to PR #1302, where we added support for Kotlin default parameters when mapping input arguments; with this PR we now also support default parameters for the DgsData methods themselves.

Example:

```kotlin
@DgsComponent
class DataFetcher {
    fun fetcher(@InputArgument input: List<String> = emptyList()): String {
        // ...
    }
}
```
If the schema does not require `input` and the query did not supply it, previously this would fail, even though the `fetcher` method is written in a way to handle the missing input. This PR updates the `DataFetcherInvoker` to call Kotlin methods reflectively using [callBy](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-callable/call-by.html) which allows supporting default parameters by omitting values in the passed in Map when the parameter is optional.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
@@ -19,14 +19,20 @@ package com.netflix.graphql.dgs.internal
 import com.netflix.graphql.dgs.internal.method.ArgumentResolverComposite
 import graphql.schema.DataFetcher
 import graphql.schema.DataFetchingEnvironment
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.reactor.mono
 import org.springframework.core.BridgeMethodResolver
-import org.springframework.core.CoroutinesUtils
-import org.springframework.core.KotlinDetector
 import org.springframework.core.MethodParameter
 import org.springframework.core.ParameterNameDiscoverer
 import org.springframework.core.annotation.SynthesizingMethodParameter
+import org.springframework.util.CollectionUtils
 import org.springframework.util.ReflectionUtils
+import java.lang.reflect.InvocationTargetException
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -62,10 +72,40 @@ class DataFetcherInvoker internal constructor(
...
-        return if (isSuspending) {
-            CoroutinesUtils.invokeSuspendingFunction(bridgedMethod, dgsComponent, *args)
+        return ReflectionUtils.invokeMethod(bridgedMethod, dgsComponent, *args)
+    }
+
+    private fun invokeKotlinMethod(kFunc: KFunction<*>, dfe: DataFetchingEnvironment): Any? {
+        val parameters = kFunc.parameters
+        val argsByName = CollectionUtils.newLinkedHashMap<KParameter, Any?>(parameters.size)
+
+        val paramSeq = if (parameters[0].kind == KParameter.Kind.INSTANCE) {
```
srinivasankavitha, line 82: Could you explain what lines #82 and #83 are doing?

kilink: We are checking if the `KFunction` takes an instance parameter, and supplying it if so; that is, the target object we want to call the method on, in this case the DGS component. That said, I can't really think if a scenario where we won't have an instance parameter, but we are kind of forced to deal with the possibility.

srinivasankavitha (approved):
Thanks for the PR! Looks good to me.  fyi - we now have 2 branches we merge PRs to, the master is on Spring Boot 3.0 and we have a spring-boot-2.7 branch as well.

PR #1421, opened 2023-02-15
Fix DgsQueryExecutor behavior with respect to HTTP headers

### LINKED ISSUE #1105: bug: NPE in integration test when not using DgsAutoConfiguration

kilink:
At some point, the various methods in DgsQueryExecutor that accept HttpHeaders were broken, due to the internal implementation of DGS switching to reading headers from the request. Work around this by adding a pluggable request customizer, DgsQueryExecutorRequestCustomizer, which allows for modifying the request inside DgsQueryExecutor. By default, an implementation of the customizer is installed that will copy any explicitly supplied headers into the request if it implements MockHttpServletRequest.

Fixes #1105 .

### DIFF
--- a/graphql-dgs-spring-boot-oss-autoconfigure/build.gradle.kts
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/build.gradle.kts
@@ -24,6 +24,8 @@ dependencies {
     compileOnly("com.github.ben-manes.caffeine:caffeine")
     compileOnly("io.micrometer:micrometer-core")
     compileOnly("io.projectreactor:reactor-core")
+    compileOnly("org.springframework:spring-test")
+    compileOnly("jakarta.servlet:jakarta.servlet-api")
 
     testImplementation("org.springframework.boot:spring-boot-starter-web")
     testImplementation("com.github.ben-manes.caffeine:caffeine")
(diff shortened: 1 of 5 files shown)


srinivasankavitha (approved):
Thanks! Looks good to me.

PR #1433, opened 2023-02-20
Remove explicit handling of Optional conversion

kilink:
Change InputObjectMapperConverter.matches to return false for Optional targets so that the ConversionService can delegate to ObjectToOptionalConverter from Spring. Gets rid of custom handling in AbstractInputArgumentResolver.


srinivasankavitha (approved):
Looks good to me!

PR #1445, opened 2023-02-28
feat(DgsSchemaProvider): use multiSourceReader for parsing schema files

coleturner:
Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----

This change makes it so that the DGS framework provides the source file name when it is specifying the source location of a given schema file. I followed @osi's notes and graphql-java does this conversion itself, if it's not already provided a MultiSourceReader. Specifying a MultiSourceReader will have graphql-java take the reader's source file name and use that when building a SourceLocation. 

Added test for validation.

### DIFF
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
@@ -42,6 +42,7 @@ import io.mockk.verify
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatNoException
 import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.Assertions.assertTrue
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
```diff
@@ -157,6 +158,21 @@ internal class DgsSchemaProviderTest {
...
+    @Test
+    fun `Should specify sourceName on SourceLocation`() {
+        val schemaProvider = schemaProvider(
+            schemaLocations = listOf("classpath*:schema/**/*.graphql*")
+        )
+
+        val schema = schemaProvider.schema()
+
+        for (type in schema.allTypesAsList) {
+            if (type.definition?.sourceLocation != null) {
```
coleturner, line 170: Both `definition` and `sourceLocation` can be null here. Built-in object types have null definitions, while built-in scalars have definitions that would appear to be default/uninitialized objects

osi (approved):

srinivasankavitha (approved):
Looks good!

gracecding (approved):

Issue #1435, opened 2023-02-23
bug:  Complete event not sent for SSE Subscriptions


yogeshdengle:
## Expected behavior

For SSE Subscriptions on the complete of the events the server sends a 'complete' event.
Currently once the Publisher sent to the DGS framework as part of the SSE Subscription finishes there is no indication on the subscription that the server is done sending the data.

## Actual behavior

the graphql-sse library (https://github.com/enisdenjo/graphql-sse) which is very commonly used in the graphql ecosystem expects a 'complete' event to be sent to indicate that the server is done sending events.
The spec is outlined here: 
https://github.com/graphql/graphql-over-http/blob/d51ae80d62b5fd8802a3383793f01bdf306e8290/rfcs/GraphQLOverSSE.md#complete-event-1

This causes the library to retry the subscription as it think it disconnected from the server. The "complete" event is used by the library to understand that this is a clean closure of the connection and not something unintended. 

## Steps to reproduce

Create a subscription that has a publisher that just sends out 3 events. Expectation is that after the 3 events we get a "complete" event.

Ancient-Dragon:
@yogeshdengle if you would like a java implementation (and can't upgrade to spring-boot 3 straight away you can just use this rest controller and remove the dgs sse package.

```java

package com.example.demo.services;

import com.example.demo.models.GraphqlServerSentEvent;
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.netflix.graphql.dgs.DgsQueryExecutor;
import com.netflix.graphql.types.subscription.Error;
import com.netflix.graphql.types.subscription.QueryPayload;
import com.netflix.graphql.types.subscription.SSEDataPayload;
import graphql.ExecutionResult;
import graphql.InvalidSyntaxError;
import graphql.language.Document;
import graphql.language.OperationDefinition;
import graphql.parser.InvalidSyntaxException;
import graphql.parser.Parser;
import graphql.validation.ValidationError;
import org.reactivestreams.Publisher;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.http.MediaType;
import org.springframework.http.codec.ServerSentEvent;
import org.springframework.web.bind.annotation.PostMapping;
import org.springframework.web.bind.annotation.RequestBody;
[...]

PR #1469, opened 2023-03-30
feat: follow graphql-sse spec, fixing complete event & errors

Ancient-Dragon:
Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Change the event type on exception to data as per the spec.

On completing the publisher, send the complete event

Fixes Issue #1435

Alternatives considered
----
N/a

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/SSESubscriptionGraphQLClient.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/SSESubscriptionGraphQLClient.kt
@@ -18,13 +18,20 @@ package com.netflix.graphql.dgs.client
 
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.netflix.graphql.types.subscription.QueryPayload
+import graphql.GraphQLException
 import org.springframework.http.MediaType
+import org.springframework.http.codec.ServerSentEvent
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.toEntityFlux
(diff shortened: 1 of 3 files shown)


srinivasankavitha:
The original SSE implementation in the framework is based on subscriptions-transport-sse : https://github.com/CodeCommission/subscriptions-transport-sse . This was prior to the introduction of graphql-sse protocol described here: https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md.

This seems to be the source of discrepancy. Your changes look good to me, but I think it would be better to introduce this as a new module to opt into call dgs-subscriptions-graphql-sse. This will prevent breakage of existing clients and also allow folks who want to use this newer standard SSE protocol to opt in.

With that strategy, we should be able to avoid making this a breaking release.

srinivasankavitha (changes requested):
I think this is the main change that may lead to breaking existing clients.

The original SSE implementation in the framework is based on subscriptions-transport-sse : https://github.com/CodeCommission/subscriptions-transport-sse . This was prior to the introduction of graphql-sse protocol described here: https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md.

This seems to be the source of discrepancy. Your changes look good to me, but I think it would be better to introduce this as a new module to opt into call dgs-subscriptions-graphql-sse. This will prevent breakage of existing clients and also allow folks who want to use this newer standard SSE protocol to opt in.

With that strategy, we should be able to avoid making this a breaking release.

srinivasankavitha:
Closing this PR since we agreed to only merge this into the master branch and not spring-boot-2.7.

PR #1540, opened 2023-05-26
feat(DgsObjectMapper): refactor initialization of object mappers

### LINKED ISSUE #1528: feature: use CustomObjectMapper to serialize subscription data
## Describe the Feature Request
A [CustomObjectMapper](https://netflix.github.io/dgs/advanced/custom-object-mapper/) can be supplied for serializing GraphQL responses. But that object mapper is not used to serialize subscription data. The [objectMapper](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-subscriptions-websockets/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandler.kt#LL109C53-L109C53) used to serialize subscription data is instantiated in the [companion object for the WebsocketGraphQLWSProtocolHandler](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-subscriptions-websockets/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandler.kt#L151) class.

This is causing our subscriptions to fail sending data because the objects we are trying to send contain `OffsetDateTime` fields that cause the ObjectMapper to throw exceptions because the `JavaTimeModule` is not registered with the subscription ObjectMapper.

## Describe Preferred Solution
Please consider injecting the CustomObjectMapper into the `WebsocketGraphQLWSProtocolHandler` class to use for serializing subscription data.

coleturner:
Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Changes in this PR
----
This change updates the DGS framework to use a singleton `DgsObjectMapper` for initializing object mappers used around the various modules. This allows us to register the Java Time Module anywhere the object mapper is used, as well as provide other modules when needed. This is an incremental change that does not refactor where modules are currently used, except to provide the Java time module everywhere.

Fixes #1528 
----


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsObjectMapper.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+
+class DgsObjectMapper {
+    companion object {
+        @Volatile
+        private var instance: ObjectMapper? = null
+
+        fun getInstance() =
+            instance ?: synchronized(this) {
```
srinivasankavitha, line 29: Don't think you need this since jacksonObjectMapper is already thread-safe.

graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt
```diff
@@ -59,7 +59,7 @@ open class DgsWebMvcAutoConfiguration {
     @Qualifier("dgsObjectMapper")
     @ConditionalOnMissingBean(name = ["dgsObjectMapper"])
     open fun dgsObjectMapper(): ObjectMapper {
-        return jacksonObjectMapper()
+        return DgsObjectMapper.getInstance()
```
srinivasankavitha, line 62: I think all you need to do is call registerJavaTimeModule here.

PR #1546, opened 2023-06-08
Rework DgsRestController to parse JSON directly from InputStream

kilink:
Rework DgsRestController to avoid buffering the request body in memory.

Update DgsRestController to accept an `InputStream` instead of using a `ByteArray`, which avoids buffering / copying the entire request into memory. Additionally, add a simple `InputQuery` class to deserialize to, which simplifies some of the code.

Other refactoring and changes in this PR:

- Split out the multipart upload handling into a separate RequestMapping handler.
- Update the DgsRestController tests to use MockMvc.
- Update MultipartVariableMapper to throw `VariableMappingException` instead of a vanilla `RuntimeException`, which allows DgsRestController to catch it and return a 4xx response.
- Fix some warnings in MultiVariableMapper.

### DIFF
--- a/graphql-dgs-client/dependencies.lock
+++ b/graphql-dgs-client/dependencies.lock
@@ -51,7 +51,7 @@
             "firstLevelTransitive": [
                 "com.netflix.graphql.dgs:graphql-dgs-subscription-types"
             ],
-            "locked": "20.2"
+            "locked": "20.3"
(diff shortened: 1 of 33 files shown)


srinivasankavitha (approved):
Thanks for the changes. Looks great to me!

PR #1548, opened 2023-06-10
DgsContext: Fix getRequestData with BatchLoaderEnvironment

Yrwein:
Pull request checklist
----

- [x] Consider creating a discussion on the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first

Pull Request type
----

- [x] Bugfix

Changes in this PR
----

`DgsContext.getRequestData(batchLoaderEnvironment: BatchLoaderEnvironment)` has been fixed to properly return DgsRequestData. It currently kicks an  exception:

```
java.lang.ClassCastException: class graphql.GraphQLContext cannot be cast to class com.netflix.graphql.dgs.context.DgsContext (graphql.GraphQLContext and com.netflix.graphql.dgs.context.DgsContext are in unnamed module of loader 'app')
	at com.netflix.graphql.dgs.context.DgsContext$Companion.getRequestData(DgsContext.kt:128) ~[graphql-dgs-7.1.0.jar:7.1.0]
	at ...
```

Alternatives considered
----

I believe the fix is fairly straight forward, but I'm not quite sure if the test is in the correct namespace etc. (DgsContext did not have a test class before.)

Please let me know if there's anything missing / to change. .) Thank you. .)

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
@@ -125,7 +125,7 @@ open class DgsContext(val customContext: Any? = null, val requestData: DgsReques
 
         @JvmStatic
         fun getRequestData(batchLoaderEnvironment: BatchLoaderEnvironment): DgsRequestData? {
-            val dgsContext = batchLoaderEnvironment.getContext<DgsContext>()
+            val dgsContext = from(batchLoaderEnvironment.getContext<GraphQLContext>())
             return dgsContext.requestData
         }
     }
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
```diff
@@ -125,7 +125,7 @@ open class DgsContext(val customContext: Any? = null, val requestData: DgsReques
 
         @JvmStatic
         fun getRequestData(batchLoaderEnvironment: BatchLoaderEnvironment): DgsRequestData? {
```
srinivasankavitha, line 127: I'm not sure if this change is completely safe/backwards compatible since the API documentation does not specify a type for the context. Perhaps it would be better to check the type first, i.e. whether it is DgsContext or GraphQLContext and return that accordingly?

srinivasankavitha (changes requested):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/context/DgsContext.kt
Yrwein: Thanks for the comment. I've done so (and fixed `getCustomContext(batchLoaderEnvironment)` too).

srinivasankavitha:
Thanks for the fixes. The changes look good to me. I'll merge it once the build passes.

Issue #1432, opened 2023-02-20
bug: DgsGraphQLMetricsInstrumentation variables parsing error


Jarda8:
There is a logic computing query complexity in `DgsGraphQLMetricsInstrumentation`. To do so, it creates a `QueryTraverser` with some parameters (see `QueryUtils.newQueryTraverser`). Among those builder parameters are `variables`, where raw (that means not yet parsed) query variables are expected, according to graphql-java doc. However, already parsed variables are used instead. That is a problem with scalars for example, where the raw value differs from the parsed one. It leads to an error being logged and query complexity not being computed for the instrumentation.

## Expected behaviour
The expected raw query variables should be passed into the `QueryTraverser` builder. Or the logic should change to avoid the redundant second variables parsing altogether.

## Actual behaviour
Already parsed query variables are passed to the `QueryTraverser` builder. That leads to an error, if parsed variable differs from the raw one.

## Steps to reproduce
Create a custom GraphQL scalar (e.g. a Value(value: String) data class) with the relevant coercing class implementing e. g. `Coercing<Value, String>`. Implement the required methods (only 'parseValue' should be required to implement properly the deserialisation). Then call a GraphQL query with a variable of the scalar type and see, that the `parseValue` is called twice - once with a raw value and then (through the Metrics Instrumentation) with an already parsed value.

srinivasankavitha:
Thanks for the detailed description. We will look into fixing this in a future release. We are currently tied up with other priorities, but will get this fixed this quarter.

ArvindR12:
I attempted recreating this issue in [this](https://github.com/Netflix/dgs-framework/pull/1551) PR, but could not. @Jarda8 I recreated the scenario you described, but I did not get the error. I linked my PR above so it's evident what I tried to do. The test case I set up successfully parsed a custom scalar that I implemented, and there are no errors in the logs.

kilink:
This was fixed in a1a6cd3, since we switched to passing in `coercedVariables` to the `QueryTraverser` builder. I can't reproduce the issue now, but I can if I revert that change from a1a6cd3.

PR #1551, opened 2023-06-12
Adding test to check for custom scalars using metrics

ArvindR12:
- [x] Bugfix

Changes in this PR
----

[Issue #1432](https://github.com/Netflix/dgs-framework/issues/1432)

This PR is an attempt to fix the issue linked above. I added a smoke test in `MicrometerServletSmokeTest.kt` to set up a test case for it. However, the test case is passing even when I recreated the scenario described in the issue. It's unclear at the moment if the issue is still outstanding or not. Creating this PR so I can link it to the issue and continue discussion there to see if I made a mistake in setting up the failure case.


kilink:
The issue was regarding queries with [variables](https://graphql.org/learn/queries/#variables) as well as custom scalars, and your example doesn't cover that, which is why it didn't trigger the issue. The associated issue has been fixed, so I will close this PR.

PR #1552, opened 2023-06-13
Use ApplicationContext to load resources in DgsSchemaProvider

kilink:
DgsSchemaProvider already has access to the ApplicationContext, so use that to load resources.

Additional minor changes:

- Use a single MultiSourceReader, which avoids the need for manually merging type registries.
- Use getBeansWithAnnotation extension method
- Handle null filename in findSchemaFiles
- Rework unit tests to use ApplicationContextRunner


srinivasankavitha (approved):
Thanks for the PR. Changes look good to me!

PR #1559, opened 2023-06-19
Get rid of usage of deprecated SimpleInstrumentation class

kilink:
Stop using the deprecated SimpleInstrumentation class and switch to SimplePerformantInstrumentation.


srinivasankavitha (approved):
Thanks for the PR! Looks good to me.

PR #1567, opened 2023-06-27
Add nullability annotations to DgsQueryExecutor

kilink:
Add nullability annotations to DgsQueryExecutor, an interface written in Java; note this is mostly to capture the "status quo" based on the method signatures of the default implementation in Kotlin in order to catch potential issues at compile time.

### DIFF
--- a/graphql-dgs-subscriptions-websockets/src/test/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandlerTest.kt
+++ b/graphql-dgs-subscriptions-websockets/src/test/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandlerTest.kt
@@ -54,6 +54,7 @@ import reactor.core.publisher.Mono
 @ExtendWith(MockKExtension::class)
 class WebsocketGraphQLWSProtocolHandlerTest {
 
+    private val objectMapper = jacksonObjectMapper()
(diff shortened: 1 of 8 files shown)


graphql-dgs-subscriptions-websockets/src/test/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandlerTest.kt
```diff
@@ -228,7 +229,7 @@ class WebsocketGraphQLWSProtocolHandlerTest {
 
         assertThat(dgsWebsocketHandler.sessions.size).isEqualTo(currentNrOfSessions + 1)
 
-        val returnMessage = jacksonObjectMapper().readValue<OperationMessage>(slot.captured.asBytes())
```
srinivasankavitha, line 231: Is this usage  problematic?

srinivasankavitha (approved):

graphql-dgs-subscriptions-websockets/src/test/kotlin/com/netflix/graphql/dgs/subscriptions/websockets/WebsocketGraphQLWSProtocolHandlerTest.kt
kilink: Not in and of itself, it just felt more natural to put the ObjectMapper instance in a field since it's used in more than one test.

PR #1581, opened 2023-07-21
 Fix the inference of the data. path in the response

berngp:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

It might be the case that data returned in the data field starts has a suffix that starts with `data`. This will incorrectly cause the GraphQLResponse to infer that it shouldn't add the `data.` suffix which will might cause confusion since the expected behavior is for the `data.` suffix to be inferred.

----

As an alternative we could leave it as is, that said, this will cause unexpected inconsistent behavior that is dependent on the payload of the response. Specifically the name of the field that is being queried.


srinivasankavitha (approved):

srinivasankavitha:
Thanks for the fix @berngp!

PR #1587, opened 2023-07-27
Fix breakage when extracting "data" as the path in GraphQLResponse

kilink:
Recently a change was made in b386856c8 that broke support for extracting the value from the response using the path "data". While users could simply use the "dataAsObject" method, there are many instances where users call extractValue with "data" as the path. Restore support for this, and add unit tests.


graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/GraphQLResponseTest.kt
```diff
@@ -238,4 +240,21 @@ class GraphQLResponseTest {
 
         server.verify()
     }
+
+    @Test
```
srinivasankavitha, line 244: We should probably add a test case for e.g. `response.extractValue("datasomething")`

PR #1610, opened 2023-08-25
Don't log queries for exceptions

gracecding:
Remove the query from error logging when an exception occurs, as this could cause security issues.


paulbakker:
Shouldn't this be configurable?

srinivasankavitha:
We want to make this the default. I agree it makes sense to have it be
configurable so it can be optionally turned on. We'll address that
separately.

gracecding:
Closing because we have decided to keep the logging as is.

Issue #1614, opened 2023-08-28
feature: add configuration toggle for query complexity analysis


tonyghita:
We use dgs-framework behind a GraphQL gateway that performs complexity analysis. This work is repeated by the dgs-framework application. The query complexity analysis is one of the more resource intensive functions in this application. We would like to skip the repeated work to give our application more headroom.

It seems the only option to disable complexity analysis we have currently is to disable all metrics.

Ideally, we could toggle analysis off/on with a configuration option. Alternatively, make it possible to provide a custom (i.e. no-op) implementation of complexity analysis.

PR #1615, opened 2023-08-29
Add property to disable query complexity metrics (#1614)

DanielThomas:
Implements #1614.


srinivasankavitha (approved):
Thanks for the PR and the quick fix!

PR #1645, opened 2023-09-22
Avoid allocating intermediate collections in some cases

kilink:
When combining iterables in various places, avoid excessive copying into intermediate collections.


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderInterceptor.kt
```diff
@@ -24,12 +23,8 @@ internal class BatchLoaderInterceptor(
                 logger.debug("Finished timer[{}] for BatchLoader.", ID)
                 timerSampler.stop(
                     Timer.builder(ID)
-                        .tags(
-                            listOf(
-                                Tag.of(GqlTag.LOADER_NAME.key, name),
-                                Tag.of(GqlTag.LOADER_BATCH_SIZE.key, result.size.toString())
-                            )
-                        ).register(registry)
+                        .tags(GqlTag.LOADER_BATCH_SIZE.key, result.size.toString(), GqlTag.LOADER_NAME.key, name)
```
srinivasankavitha, line 26: Is this changing how the tags are added? With your update above, do they all just get added individually as 4 tags instead of 2?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
```diff
@@ -144,16 +151,16 @@ open class DgsDataLoaderRegistry : DataLoaderRegistry() {
...
     }
 
     /**
      * This will be called [org.dataloader.DataLoader.dispatch] on each of the registered
      * [org.dataloader.DataLoader]s
      */
     override fun dispatchAll() {
-        scheduledDataLoaderRegistries.forEach {
-            it.value.dispatchAll()
+        scheduledDataLoaderRegistries.forEach { (_, value) ->
```
srinivasankavitha, line 162: I'm surprised this makes a difference in performance?

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/MappedBatchLoaderInterceptor.kt
```diff
@@ -23,12 +22,8 @@ internal class MappedBatchLoaderInterceptor(
                 logger.debug("Stopping timer[{}] for {}", ID, javaClass.simpleName)
                 timerSampler.stop(
                     Timer.builder(ID)
-                        .tags(
-                            listOf(
-                                Tag.of(GqlTag.LOADER_NAME.key, name),
-                                Tag.of(GqlTag.LOADER_BATCH_SIZE.key, result.size.toString())
-                            )
-                        ).register(registry)
+                        .tags(GqlTag.LOADER_BATCH_SIZE.key, result.size.toString(), GqlTag.LOADER_NAME.key, name)
```
srinivasankavitha, line 25: Same comment as earlier wrt tags.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderInterceptor.kt
kilink: It's still adding the 2 tags; this method takes a varargs String array and expects an even number of args representing the key / value pairs. Calling this instead avoids needing to create a List; it's going to create a Tags instance under the hood either way. . The tag keys are re-ordered to avoid needing to actually do any sorting as a sort of micro-optimization.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
kilink: It doesn't, I just changed it since I was touching the file anyway.

srinivasankavitha (approved):

Issue #1577, opened 2023-07-07
bug: Ignore unknown json properties when parsing request object into InputQuery type


jrogers:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue.

## Expected behavior

That request parsing be resilient to unknown properties in the graphql request. This seems to have been a regression in 7.x.

## Actual behavior

If a graphql request is called by an older graphql implementation and say sets the operation name in the JSON serialization as legacy `operation_name` the JSON parsing code will fail here and return bad request - https://github.com/Netflix/dgs-framework/blob/df1693edb934634a7775a7ff982e2caa339b0617/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt#L115
This is because it's is trying to strictly map to the new `InputQuery` type and does not ignore unknown properties. Previously the code prior to v7, would parse the request into the map, and it would simply end up not setting the operation name (which is fine).

Possible fix, add `@JsonIgnoreProperties(ignoreUnknown = true)` to the `InputQuery` type to make it more resilient - https://github.com/Netflix/dgs-framework/blob/df1693edb934634a7775a7ff982e2caa339b0617/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt#L82

Note that because the `operationName` field on the `InputQuery` type is nullable, not setting it at all works as expected.

## Steps to reproduce

Create a graphql request to any endpoint but change `operationName` property to `operation_name`, or even easier add an unknown/non-standard property to the request json. It will fail with 400.

gracecding:
We will look into this.

congotej:
Hi @jrogers! I am going to go ahead and close this, as the fix was just merged and should be out in the next release. Thanks for bringing this up!

PR #1650, opened 2023-09-25
#1577 bug: Ignore unknown json properties when parsing request object into …

congotej:
[bug: Ignore unknown json properties when parsing request object into InputQuery type](https://github.com/Netflix/dgs-framework/issues/1577#top)
#1577

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Fix for [Issue #1577](https://github.com/Netflix/dgs-framework/issues/1577)

- Help make the GraphQL request parsing more resilient when dealing with older GraphQL implementations that use different field names. This annotation tells the Jackson ObjectMapper to ignore any unknown properties in the JSON serialization, which would allow the code to continue parsing the request even if the operation name is set with a different field name like `legacy operation_name`.

Alternatives considered
----

- Can also only only add @JsonIgnoreProperties to the operationName instead of on the entire data class. This way, if the incoming GraphQL request has an "operation_name" field instead of "operationName," it will be ignored, and the parsing will continue without errors.

### DIFF
--- a/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
+++ b/graphql-dgs-spring-webmvc/src/main/kotlin/com/netflix/graphql/dgs/mvc/DgsRestController.kt
@@ -16,6 +16,7 @@
 
 package com.netflix.graphql.dgs.mvc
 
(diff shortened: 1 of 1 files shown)


srinivasankavitha (approved):
Looks good!

kilink (approved):

PR #1652, opened 2023-09-30
Detect and throw exception for dataloaders with same name

srinivasankavitha:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):

Changes in this PR
----

Issue #
It is hard to detect when we have multiple dataloaders registered with the same name. This PR introduces a change to detect that and throw an exception.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
@@ -16,14 +16,10 @@
 
 package com.netflix.graphql.dgs.internal
 
-import com.netflix.graphql.dgs.DataLoaderInstrumentationExtensionProvider
-import com.netflix.graphql.dgs.DgsComponent
-import com.netflix.graphql.dgs.DgsDataLoader
-import com.netflix.graphql.dgs.DgsDataLoaderOptionsProvider
-import com.netflix.graphql.dgs.DgsDataLoaderRegistryConsumer
-import com.netflix.graphql.dgs.DgsDispatchPredicate
+import com.netflix.graphql.dgs.*
 import com.netflix.graphql.dgs.exceptions.DgsUnnamedDataLoaderOnFieldException
 import com.netflix.graphql.dgs.exceptions.InvalidDataLoaderTypeException
+import com.netflix.graphql.dgs.exceptions.MultipleDataLoadersDefinedException
 import com.netflix.graphql.dgs.exceptions.UnsupportedSecuredDataLoaderException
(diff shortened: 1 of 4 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
```diff
@@ -87,50 +85,53 @@ class DgsDataLoaderProvider(
...
-        dataLoaders.values.forEach { dgsComponent ->
-            val javaClass = AopUtils.getTargetClass(dgsComponent)
-            val annotation = javaClass.getAnnotation(DgsDataLoader::class.java)
-            val predicateField = javaClass.declaredFields.asSequence().find { it.isAnnotationPresent(DgsDispatchPredicate::class.java) }
-            if (predicateField != null) {
-                ReflectionUtils.makeAccessible(predicateField)
-                val dispatchPredicate = predicateField.get(dgsComponent)
-                if (dispatchPredicate is DispatchPredicate) {
-                    addDataLoaders(dgsComponent, javaClass, annotation, dispatchPredicate)
+        dataLoaders.values.distinctBy { AopUtils.getTargetClass(it) } // filter duplicate beans for the same class (tests)
```
srinivasankavitha, line 120: This was needed because I was seeing multiple beans (generic and root) for the same class in the context of tests.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
```diff
@@ -87,50 +85,53 @@ class DgsDataLoaderProvider(
     }
 
     private fun addDataLoaderFields() {
-        applicationContext.getBeansWithAnnotation(DgsComponent::class.java).values.forEach { dgsComponent ->
-            val javaClass = AopUtils.getTargetClass(dgsComponent)
+        val dataLoaders = applicationContext.getBeansWithAnnotation(DgsComponent::class.java)
+        dataLoaders.values.distinctBy { AopUtils.getTargetClass(it) }
```
paulbakker, line 89: Why doing distinctBy here, because that would avoid/hide the problem? It should just fail if multiple with the same name are found I think

srinivasankavitha: So I added a comment earlier as well `This was needed because I was seeing multiple beans (generic and root) for the same class in the context of tests.`

What I was seeing is that in the context of tests, I see 2 beans for the same provider class - one was a Generic bean and the other a Root bean. I'm not sure if this is expected behavior though.

PR #1660, opened 2023-10-10
fix: in an input object a nested scalar is not mapped

gilteplitsky:
Bug Fix

When an input has a nested enum scalar in it
The scalar will not be mapped to the object


srinivasankavitha (approved):
Thanks for the PR! Looks good to me.

PR #1662, opened 2023-10-10
(Experimental) Added support for automatically running data fetchers on virtual threads

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

When running on JDK 21 this change wraps each data fetcher in a ComletableFuture, which is run on a new virtual thread. This enables parallel data fetching without needing to manually deal with CompletableFuture, and without the need to manually manage a thread pool for data fetchers.
The future currently needs to be enabled with `dgs.graphql.virtualthreads.enabled=true`. We might enable by default at a later point in time if this turns out to be beneficial.

- This doesn't (yet) affect data loaders
- DataFetchers that explicitly return CompletableFuture are ignored for this feature, they'll continue to use the `Executor` specified by the user. This way you can, if needed, be specific about using a thread pool for certain data fetchers if they are CPU bound.
- It uses a multi release JAR to be able to use Java 21 APIs for Virtual Threads, while still compiling the project with Java 17.
- The virtual thread factory can be provided by the user as a `@Bean`. This is important for integration with other frameworks for things like context propagation.

### DIFF
--- a/graphql-dgs/src/main/java/com/netflix/graphql/dgs/internal/VirtualThreadTaskExecutor.java
+++ b/graphql-dgs/src/main/java/com/netflix/graphql/dgs/internal/VirtualThreadTaskExecutor.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.internal;
+
+import org.jetbrains.annotations.NotNull;
+import org.springframework.core.task.AsyncTaskExecutor;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+
+/**
+ * Unimplemented version of the VirtualThreadTaskExecutor just for reachability.
+ * The actual implementation is in the java21 source folder, using the multi-release JAR feature.
+ */
+public class VirtualThreadTaskExecutor implements AsyncTaskExecutor {
+    public VirtualThreadTaskExecutor() {
+    }
+
+    @Override
+    public void execute(@NotNull Runnable task) {
+        throw new UnsupportedOperationException("VirtualThreadTaskExecutor is only supported on JDK 21+");
+    }
+
+    @Override
+    public void execute(@NotNull Runnable task, long startTimeout) {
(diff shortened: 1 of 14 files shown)


graphql-dgs/src/main/java/com/netflix/graphql/dgs/internal/VirtualThreadTaskExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+package com.netflix.graphql.dgs.internal;
+
+import org.jetbrains.annotations.NotNull;
+import org.springframework.core.task.AsyncTaskExecutor;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+
+/**
+ * Unimplemented version of the VirtualThreadTaskExecutor just for reachability.
```
srinivasankavitha, line 26: What happens if you don't provide this? Is it causing a missing Bean?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -113,4 +130,32 @@ class DataFetcherInvoker internal constructor(
...
+        }, taskExecutor)
+    }
+
+    /**
+     * Decides if a data fetcher method should be wrapped in CompletableFuture automatically.
+     * This is only done when a taskExecutor is available, and if the data fetcher doesn't explicitly return CompletableFuture already.
+     * Used when virtual threads are enabled.
+     */
+    private fun shouldWrapInCompletableFuture(kFunc: KFunction<*>): Boolean {
+        return taskExecutor != null && !kFunc.returnType.isSubtypeOf(typeOf<CompletionStage<Any>>())
```
srinivasankavitha, line 150: This condition covers CompletableFuture as well?

srinivasankavitha (approved):
We should also caveat the PR to state that it is only for data fetchers and not data loaders yet since it's not documented.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
paulbakker: yes, CompletionStage sits at the top of that type structure

graphql-dgs/src/main/java/com/netflix/graphql/dgs/internal/VirtualThreadTaskExecutor.java
paulbakker: No, it's injected here as an `Optional`, and the other code all checks if it's available or not. 

```
open fun methodDataFetcherFactory(argumentResolvers: ObjectProvider<ArgumentResolver>, @Qualifier("dgsAsyncTaskExecutor") taskExecutorOptional: Optional<AsyncTaskExecutor>
```

paulbakker: Because it's also conditional on Java 21, so it needs to be able to handle the case where it's not there.

PR #1692, opened 2023-11-03
Initialize additionalInformation with emptyMap in GraphQLErrorDebugInfo

kilink:


kilink:
This actually won't work without some modification, closing for now.

PR #1705, opened 2023-11-09
Update set up for ScheduledExecutorService.

srinivasankavitha:
This PR has the following changes:
1. Inject a ScheduledExecutorService bean for use across requests, qualified with @DgsScheduledExecutorService
2. Update to use the latest java-dataloader 3.2.2 to pick up [this fix] to avoid instantiating a new service per request (https://github.com/graphql-java/java-dataloader/pull/135)
3. Make the tickerMode configurable, default set to false

### DIFF
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -69,6 +69,8 @@ import org.springframework.mock.web.MockHttpServletRequest
 import org.springframework.web.context.request.NativeWebRequest
 import org.springframework.web.context.request.WebRequest
 import java.util.*
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
 
 /**
  * Framework auto configuration based on open source Spring only, without Netflix integrations.
@@ -153,9 +155,16 @@ open class DgsAutoConfiguration(
         return DefaultDataLoaderOptionsProvider()
     }
 
+    @Bean(destroyMethod = "shutdown")
+    @ConditionalOnMissingBean
+    @Qualifier("dgsScheduledExecutorService")
+    open fun dgsScheduledExecutorService(): ScheduledExecutorService {
+        return Executors.newSingleThreadScheduledExecutor()
+    }
+
     @Bean
-    open fun dgsDataLoaderProvider(applicationContext: ApplicationContext, dataloaderOptionProvider: DgsDataLoaderOptionsProvider): DgsDataLoaderProvider {
-        return DgsDataLoaderProvider(applicationContext, dataloaderOptionProvider)
+    open fun dgsDataLoaderProvider(applicationContext: ApplicationContext, dataloaderOptionProvider: DgsDataLoaderOptionsProvider, @Qualifier("dgsScheduledExecutorService") dgsScheduledExecutorService: ScheduledExecutorService): DgsDataLoaderProvider {
+        return DgsDataLoaderProvider(applicationContext, dataloaderOptionProvider, dgsScheduledExecutorService, configProps.dataloaderTickerModeEnabled)
     }
 
     /**
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
@@ -39,12 +39,13 @@ import graphql.execution.ExecutionStrategy
 import graphql.execution.instrumentation.Instrumentation
 import graphql.execution.preparsed.PreparsedDocumentProvider
 import graphql.schema.GraphQLSchema
+import org.dataloader.registries.ScheduledDataLoaderRegistry
 import org.intellij.lang.annotations.Language
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.util.StringUtils
-import java.util.Optional
+import java.util.*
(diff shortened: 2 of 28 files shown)


graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -154,8 +156,14 @@ open class DgsAutoConfiguration(
     }
 
     @Bean
-    open fun dgsDataLoaderProvider(applicationContext: ApplicationContext, dataloaderOptionProvider: DgsDataLoaderOptionsProvider): DgsDataLoaderProvider {
-        return DgsDataLoaderProvider(applicationContext, dataloaderOptionProvider)
+    @ConditionalOnMissingBean
+    open fun scheduledExecutorService(): ScheduledExecutorService {
+        return Executors.newSingleThreadScheduledExecutor()
```
kilink, line 161: I think this will cause problems if someone is trying to autowire an `Executor` bean, since `ScheduledExecutorService` also implements `Executor`.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -154,8 +156,14 @@ open class DgsAutoConfiguration(
     }
 
     @Bean
```
kilink, line 158: Should we set `destroyMethod="shutdown"`?

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
srinivasankavitha: Thanks, yeah I am changing it to add a quaifier for `dgsScheduledExecutorService`. Would that mitigate the issue you think?

kilink: Yeah, a Qualifier would avoid the problem

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
```diff
@@ -45,7 +48,9 @@ import kotlin.system.measureTimeMillis
  */
 class DgsDataLoaderProvider(
     private val applicationContext: ApplicationContext,
-    private val dataLoaderOptionsProvider: DgsDataLoaderOptionsProvider = DefaultDataLoaderOptionsProvider()
+    private val dataLoaderOptionsProvider: DgsDataLoaderOptionsProvider = DefaultDataLoaderOptionsProvider(),
+    private val scheduledExecutorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(),
```
paulbakker, line 52: I don't really know when/why to use `singleThreadScheduleExecutor` vs a pool of scheduled threads with `Executors.newScheduledThreadPool(int corePoolSize)`. It might be good to have this configurable, but I'm not sure.

paulbakker (approved):
Maybe the pool size for the scheduled executor should be configurable, but otherwise looks great.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
srinivasankavitha: Since this is set up as a bean in our autoconfig, I'm thinking the the user can override  the dgsScheduledExecutorService bean to set up differently?

kilink: I also thought about that and am not sure what a good default is. It probably depends on the load / RPS. If you have a lot of events scheduled at or around the same instant I can see one thread becoming a bottleneck; but I also think it's something that could be solved in a follow up.

srinivasankavitha: Thanks for the input folks! The default is just based on what the java-dataloader library already picked. Will merge this PR for now and we can make it configurable if needed in a follow up.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
```diff
@@ -118,7 +119,12 @@ object BaseDgsQueryExecutor {
                 .extensions(extensions.orEmpty())
                 .build()
             graphQLContextFuture.complete(executionInput.graphQLContext)
-            graphQL.executeAsync(executionInput)
+            graphQL.executeAsync(executionInput).thenApply {
```
kilink, line 122: `whenComplete` might be safer, although I don't know under what circumstances `executeAsync` ever really fails; maybe cancellation?

kilink (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
kilink: Right, but the default uses a new executor for every registry created. That's in and of itself not ideal and runs into its own scaling problem of spinning up too many threads, but it wouldn't run into this specific scaling problem where an executor is shared since it's essentially spinning up a thread per request. I suspect a default like corePoolSize set to the number of CPUs might make more sense. But if the default here is a problem we'll see it in testing, and can fix it in a follow up like you said.

Issue #1569, opened 2023-06-29
bug: Custom Scalars are not parsed when included in `_entities` query representations


gwardwell:
Unlike field arguments which parse custom scalar values on the way into a data fetcher, custom scalars are not parsed when included in `_entities` query representations.

#### Schema

```graphql
# Parses to `Long`
scalar Int64

type Book @key(fields: "bookId") {
  bookId: Int64!
}

type Query {
  bookById(bookId: Int64!): Book
}
```

#### Query

Query:

```graphql
query GetEntities($representations: [_Any!]!) {
  _entities(representations: $representations) {
    ... on Book {
      __typename
      bookId
    }
  }
}
```

Variables:

```json
{
  "representations": [
    {
      "__typename": "Book",
      "bookId": "7654323234565432345676543234567"
    }
  ]
}
```

#### Data Fetchers

```java
@DgsQuery
public CompletableFuture<DemoBook> book(
    // `bookId` is parsed to a `Long` by the custom scalar logic
    @InputArgument(value = "bookId") Long bookId
) {
  ...
}

@DgsEntityFetcher(name = "Book")
public CompletableFuture<Book> book(Map<String, Object> values) {
  // The `bookId` is not parsed to a `Long` by the custom scalar and is
  // instead typed as a generic `Object`, but is actually a String.
  String bookId = (String) values.get("bookId");
  ...
}
```

### Nested Keys

Nested keys may provide complication in solving this problem as only the root level of an entity representation includes the `__typename`. But all fields in the key are represented in the schema, making the schema a "source of truth" for mapping these nested types.

#### Schema

```graphql
# Parses to `Long`
scalar Int64

# Parses to `OffsetDateTime`
scalar DateTime

type Book @key(fields: "bookId publicationInfo { publicationDate }") {
  bookId: Int64!
  publicationInfo: PublicationInfo!
}

type PublicationInfo {
  publicationDate: DateTime!
}

type Query {
  bookById(bookId: Int64!): Book
}
```

#### Query

Query:

```graphql
query GetEntities($representations: [_Any!]!) {
  _entities(representations: $representations) {
    ... on Book {
      __typename
      bookId
      publicationInfo {
        publicationDate
      }
    }
  }
}
```

Variables:

```json
{
  "representations": [
    {
      "__typename": "Book",
      "bookId": "7654323234565432345676543234567",
      "publicationInfo": {
        "publicationDate": "2020-02-08 09:30:26"
      }
    }
  ]
}
```

[...]

PR #1710, opened 2023-11-13
fix issue 1569 -- parse custom scalars in _entities representations

Emily:
Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

This PR addresses #1569 by parsing custom scalars that are contained within entity representations (based on the `@key` fields in the local schema). 

This functionality is turned off by default, as it can cause backwards compatibility issues:
- in a future version it will be turned on by default
- currently it can be enabled by setting the `dgs.graphql.enableEntityFetcherCustomScalarParsing` property to `true`

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -26,13 +26,17 @@ import com.netflix.graphql.dgs.exceptions.NoSchemaFoundException
 import com.netflix.graphql.dgs.federation.DefaultDgsFederationResolver
 import com.netflix.graphql.dgs.internal.method.InputArgumentResolver
 import com.netflix.graphql.dgs.internal.method.MethodDataFetcherFactory
+import com.netflix.graphql.dgs.internal.utils.SelectionSetUtil
 import com.netflix.graphql.mocking.DgsSchemaTransformer
 import com.netflix.graphql.mocking.MockProvider
 import graphql.TypeResolutionEnvironment
 import graphql.execution.DataFetcherExceptionHandler
 import graphql.language.FieldDefinition
+import graphql.language.ImplementingTypeDefinition
 import graphql.language.InterfaceTypeDefinition
 import graphql.language.ObjectTypeDefinition
+import graphql.language.ScalarTypeDefinition
+import graphql.language.StringValue
 import graphql.language.TypeName
 import graphql.language.UnionTypeDefinition
 import graphql.parser.MultiSourceReader
@@ -90,7 +94,8 @@ class DgsSchemaProvider(
     private val defaultDataFetcherFactory: Optional<DataFetcherFactory<*>> = Optional.empty(),
     private val methodDataFetcherFactory: MethodDataFetcherFactory,
     private val componentFilter: ((Any) -> Boolean)? = null,
-    private val schemaWiringValidationEnabled: Boolean = true
+    private val schemaWiringValidationEnabled: Boolean = true,
+    private val enableEntityFetcherCustomScalarParsing: Boolean = false
 ) {
 
     private val schemaReadWriteLock = ReentrantReadWriteLock()
@@ -175,16 +180,6 @@ class DgsSchemaProvider(
             mergedRegistry = mergedRegistry.merge(existingTypeDefinitionRegistry.get())
         }
 
-        val federationResolverInstance =
-            federationResolver.orElseGet {
-                DefaultDgsFederationResolver(
-                    entityFetcherRegistry,
-                    dataFetcherExceptionHandler
-                )
-            }
-
-        val entityFetcher = federationResolverInstance.entitiesFetcher()
-        val typeResolver = federationResolverInstance.typeResolver()
         val codeRegistryBuilder = GraphQLCodeRegistry.newCodeRegistry().fieldVisibility(fieldVisibility)
         if (defaultDataFetcherFactory.isPresent) {
             codeRegistryBuilder.defaultDataFetcher(defaultDataFetcherFactory.get())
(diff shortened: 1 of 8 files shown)


Emily:
FYI @gwardwell @wbaldoumas

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsRolloutFlags.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs
+
+data class DgsRolloutFlags(
```
Emily, line 19: I'm not particularly happy with having to introduce this, but I can't think of a way to inject this config into `DefaultDgsFederationResolver` (and subclasses) without it.

srinivasankavitha: Instead of creating a bean for DgsRolloutFlags, you can create a bean for the DgsFederationResolver. This bean will be injected into the DgsSchemaProvider, so you can pass in the config to the DgsFederationResolver directly when creating the bean there in the DgsAutoConfiguration.

srinivasankavitha: I think you could just pass the config setting as a boolean flag into the DgsSchemaProvider and then pass that through to the federation resolver.

paulbakker, line 19: Instead of a flag to completely enable/disable the feature, could it be opt-in per entity fetcher? E.g. add it as a new optional parameter to `@DgsEntityFetcher`. Something like `@DgsEntityFetcher(name = "Show", enableCustomScalarCoercion=true)`.

That way it will be a bit more easy to gradually adopt this.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -508,10 +518,57 @@ class DgsSchemaProvider(
                         enableInstrumentation
 
                     entityFetcherRegistry.entityFetchers[dgsEntityFetcherAnnotation.name] = dgsComponent to method
+
+                    val type = registry.getType(dgsEntityFetcherAnnotation.name)
+
+                    type.ifPresent {
+                        val typeDefinition = it as? ObjectTypeDefinition
```
srinivasankavitha, line 525: If the type that has a key is an interface, this doesn't work. e.g Movie is an interface with key movieId

srinivasankavitha (changes requested):
Overall PR looks great. Tested out the changes with an example using Long and that works as expected as well.  In addition to ObjectTypeDefinition, we also need to handle keys on InterfaceTypeDefinition.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsRolloutFlags.kt
Emily: Unfortunately I don't think so, since DGSes can (and do) subclass `DefaultDgsFederationResolver` and then rely on the `@Autowired` annotations instead of the constructor. However, I think Paul's suggestion is a nice sidestep to this problem.

Emily: This is a great idea, and avoids the awkwardness of this class. :)

srinivasankavitha: Once concern with introducing a flag in the annotation might make it harder to deprecate, since it is now kind of a feature rather than a bug fix. But we could also flip the default for the flag I guess in the future in a breaking release.

Emily:
@srinivasankavitha @paulbakker pushed a revised set of changes.

srinivasankavitha (approved):
Looks great to me!

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsRolloutFlags.kt
paulbakker: Yeah my suggestion is only good if *is* like a feature. If it's really considered a bug fix I agree it shouldn't be part of the API. If it's in the annotation, it will need to stay.

Emily: Yeah, I've swapped back to the property as we're really viewing this as a bugfix, i.e. bringing it in line with input parsing.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -508,10 +518,65 @@ class DgsSchemaProvider(
                         enableInstrumentation
 
                     entityFetcherRegistry.entityFetchers[dgsEntityFetcherAnnotation.name] = dgsComponent to method
+
+                    val type = registry.getType(dgsEntityFetcherAnnotation.name)
+
```
srinivasankavitha, line 523: nit: Maybe just add a comment that we  are using the config for backwards compatibility and we intend to get rid of it

srinivasankavitha (approved):

PR #1721, opened 2023-11-21
disabling more SSE tests

paulbakker:
prevent stuck builds


Ancient-Dragon:
@paulbakker, I've noticed you've been disabling these tests but I'm not sure why they are valid tests and we don't want anything to break in them. If there is an issue, surely it's better to find out what caused them to be flakey rather than just disabling?

paulbakker:
We've been struggling with randomly hanging builds for a whil and have yet
to be able to pinpoint what's the problem. We have a hunch that the SSE
tests are having some race condition.
Once we confirm that disabling them fixes the problem we'll look into a
better solution.

PR #1735, opened 2023-12-08
Delegate to ConversionService in DefaultInputObjectMapper

kilink:
Switch to using a ConversionService from Spring to perform conversion in DefaultInputObjectMapper; this greatly simplifies the implementation, and allows for extensibility in the future.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
@@ -20,23 +20,54 @@ import com.netflix.graphql.dgs.exceptions.DgsInvalidInputArgumentException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.core.KotlinDetector
+import org.springframework.core.ResolvableType
+import org.springframework.core.convert.ConversionException
+import org.springframework.core.convert.TypeDescriptor
+import org.springframework.core.convert.converter.ConditionalGenericConverter
+import org.springframework.core.convert.converter.GenericConverter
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
```diff
@@ -105,50 +105,29 @@ class DefaultInputObjectMapper(private val customInputObjectMapper: InputObjectM
...
-                    }
-                    trySetField(declaredField, instance, mappedValue)
-                } else if (it.value is List<*>) {
-                    val newList = convertList(it.value as List<*>, targetClass, fieldClass.kotlin, fieldArgumentType)
-                    if (declaredField.type == Set::class.java) {
-                        trySetField(declaredField, instance, newList.toSet())
-                    } else {
-                        trySetField(declaredField, instance, newList)
-                    }
-                } else if (fieldClass.isEnum) {
```
srinivasankavitha, line 142: I think we don't have a test covering the enum field case. Should we add one for that as well?

kilink: There is coverage for Enums already in [InputArgumentTest](https://github.com/Netflix/dgs-framework/blob/534c4a832eeb1808018c8b9ff077a5c16cbcf217/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt#L1954).

kilink: I went ahead and added an additional test case for this.

srinivasankavitha (approved):
Thanks for the PR! The implementation does look much simpler now.

PR #1737, opened 2023-12-09
Add support for custom object mapper in GraphQLResponse

gabbigum:
The default object mapper is not enough to satisfy all deserialization scenarios.
One such scenario is where we need object mapper with custom deserializer for graphql.relay.PageInfo.

Pull request checklist
----

- [ ] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
  first

Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

Issue #

Alternatives considered
----

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
@@ -41,16 +41,18 @@ import org.slf4j.LoggerFactory
 data class GraphQLResponse(
     @Language("json") val json: String,
     val headers: Map<String, List<String>>,
-    val mapper: ObjectMapper
+    private val mapper: ObjectMapper
 ) {
 
     /**
      * A JsonPath DocumentContext. Typically, only used internally.
      */
-    val parsed: DocumentContext = JsonPath.using(Configuration.builder()
-        .jsonProvider(JacksonJsonProvider(mapper))
-        .mappingProvider(JacksonMappingProvider(mapper)).build()
-        .addOptions(Option.DEFAULT_PATH_LEAF_TO_NULL)).parse(json)
+    val parsed: DocumentContext = JsonPath.using(
+        Configuration.builder()
+            .jsonProvider(JacksonJsonProvider(mapper))
+            .mappingProvider(JacksonMappingProvider(mapper)).build()
+            .addOptions(Option.DEFAULT_PATH_LEAF_TO_NULL)
+    ).parse(json)
 
     /**
      * Map representation of data
@@ -60,14 +62,12 @@ data class GraphQLResponse(
     val errors: List<GraphQLError> = parsed.read("errors", jsonTypeRef<List<GraphQLError>>()) ?: emptyList()
 
     constructor(@Language("json") json: String) : this(json, emptyMap())
-    constructor(@Language("json") json: String, headers: Map<String, List<String>>) : this(json,
+    constructor(@Language("json") json: String, headers: Map<String, List<String>>) : this(
+        json,
         headers,
         // default object mapper instead no instance is passed in the constructor
-        jacksonObjectMapper()
-        .registerModule(JavaTimeModule())
-        .registerModule(ParameterNamesModule())
-        .registerModule(Jdk8Module())
-        .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE))
+        DEFAULT_MAPPER
+    )
 
     /**
      * Deserialize data into the given class.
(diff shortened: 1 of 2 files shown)


gabbigum:
Hi Team! I want to help add this generic functionallity inside the GraphQLResponse. This is rough draft of just adding the dependency to GraphQLResponse. 

I will need to investigate how I can inject the objectmapper directly from the spring context. I would like to discuss what you think of such support and can provide use-cases for it! Any feedback is welcome :)

srinivasankavitha:
Thanks for the PR. The changes look good, there is a linter error that is failing the build btw. Once that is fixed, we should be good to merge,

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
```diff
@@ -53,6 +60,14 @@ data class GraphQLResponse(@Language("json") val json: String, val headers: Map<
     val errors: List<GraphQLError> = parsed.read("errors", jsonTypeRef<List<GraphQLError>>()) ?: emptyList()
 
     constructor(@Language("json") json: String) : this(json, emptyMap())
+    constructor(@Language("json") json: String, headers: Map<String, List<String>>) : this(json,
```
kilink, line 63: This creates a new mapper every time a `GraphQLResponse` is instantiated. I think we should leave the default mapper construction in the companion object.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
```diff
@@ -38,12 +38,19 @@ import org.slf4j.LoggerFactory
  * Representation of a GraphQL response, which may contain GraphQL errors.
  * This class gives convenient JSON parsing methods to get data out of the response.
  */
-data class GraphQLResponse(@Language("json") val json: String, val headers: Map<String, List<String>>) {
+data class GraphQLResponse(
+    @Language("json") val json: String,
+    val headers: Map<String, List<String>>,
+    val mapper: ObjectMapper
```
kilink, line 44: Maybe we should make the property private?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
srinivasankavitha: Good catch. +1

gabbigum: If we leave the default mapper in the companion object, how can we achieve the dependency injection? I'm not aware of the kotlin semantics, but I suspect that this companion object is some sort of a static instance.

kilink: I was suggesting leaving the default ObjectMapper in the companion object so that it's only instantiated once, but can be used as an argument to the constructor. e.g.:

```diff
@@ -38,7 +38,9 @@ import org.slf4j.LoggerFactory
  * Representation of a GraphQL response, which may contain GraphQL errors.
  * This class gives convenient JSON parsing methods to get data out of the response.
  */
-data class GraphQLResponse(@Language("json") val json: String, val headers: Map<String, List<String>>) {
+data class GraphQLResponse(@Language("json") val json: String,
+                           val headers: Map<String, List<String>>,
+                           private val mapper: ObjectMapper = DEFAULT_MAPPER) {
 
     /**
      * A JsonPath DocumentContext. Typically, only used internally.
@@ -120,17 +122,12 @@ data class GraphQLResponse(@Language("json") val json: String, val headers: Map<
     companion object {
         private val logger: Logger = LoggerFactory.getLogger(GraphQLResponse::class.java)
 
-        private val mapper: ObjectMapper = jacksonObjectMapper()
+        private val DEFAULT_MAPPER: ObjectMapper = jacksonObjectMapper()
             .registerModule(JavaTimeModule())
             .registerModule(ParameterNamesModule())
             .registerModule(Jdk8Module())
             .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)
```

gabbigum: This definitely looks better!

gabbigum:
Hi guys, could I get another review/+2? Thank you :)

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLResponse.kt
```diff
@@ -118,19 +133,14 @@ data class GraphQLResponse(@Language("json") val json: String, val headers: Map<
     fun hasErrors(): Boolean = errors.isNotEmpty()
 
     companion object {
-        private val logger: Logger = LoggerFactory.getLogger(GraphQLResponse::class.java)
+        val logger: Logger = LoggerFactory.getLogger(GraphQLResponse::class.java)
```
paulbakker, line 136: Why did this change? Seems unrelated?

paulbakker:
@gabbigum I think you introduced a compilation issue while rebasing (default mapper is there twice now). Could you fix that and squash the commits into a single commit? Changes look good, so I'll merge when that's done!

gabbigum:
Hi Paul, I'm having some issues squashing the commits, as I've already rebased from master and if I squash them I'll override  some of the commits I've rebased. Could you assist me with this?

paulbakker:
Thanks @gabbigum!

PR #1760, opened 2024-01-02
Fix handling of UnknownOperationException

### LINKED ISSUE #1751: bug: UnknownOperationException should return 400 instead of 500
## Expected behavior
`{ "operationName": "books", "variables": {}, "query": "{ books {\nid\nname\n} }" }`
When this type of request/query is submitted (note, the operationName "books" without a matching named query), an UnknownOperationException is thrown from graphql.language.NodeUtil. When an UnknownOperationException is thrown, the result should be ResponseEntity.badRequest... since it is a malformed request.

## Actual behavior
When this type of query is submitted and an UnknownOperationException is thrown, the current behavior is that an Internal Server Error is returned to the caller.

## Steps to reproduce
Submit a named query with non-matching operationName. Or submit a non-named query with any operationName specified. Or submit two named queries without an operationName specified.

kilink:
When GraphQL.executeAsync throws UnknownOperationException, or any exception implementing GraphQLError, set the status to BAD_REQUEST / 400 in BaseDgsQueryExecutor. Fixes #1751.

### DIFF
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutorTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutorTest.kt
@@ -33,11 +33,14 @@ import com.netflix.graphql.dgs.internal.method.MethodDataFetcherFactory
 import graphql.InvalidSyntaxError
 import graphql.execution.AsyncExecutionStrategy
 import graphql.execution.AsyncSerialExecutionStrategy
+import graphql.execution.UnknownOperationException
 import graphql.execution.instrumentation.SimplePerformantInstrumentation
 import io.mockk.every
 import io.mockk.impl.annotations.MockK
 import io.mockk.junit5.MockKExtension
+import org.assertj.core.api.Assertions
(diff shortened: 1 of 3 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutorTest.kt
```diff
@@ -204,6 +205,17 @@ internal class DefaultDgsQueryExecutorTest {
             .isInstanceOf(InvalidSyntaxError::class.java)
     }
 
+    @Test
+    fun `Invalid operation returns a GraphQL Error with type UnknownOperationException`() {
```
srinivasankavitha, line 209: Should we also check  that the status is set to BAD_REQUEST?

Emily (approved):

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/DefaultDgsQueryExecutorTest.kt
kilink: I've added an assertion for that.

srinivasankavitha (approved):

PR #1766, opened 2024-01-04
Integrate Spring Graphql with DGS Framework

paulbakker:
The DGS and Spring-GraphQL teams are super excited to introduce deep integration between the DGS framework and Spring-GraphQL. This will bring the community together, and we can continue building the best possible GraphQL framework for Spring Boot in the future.

## Getting Started with DGS/Spring-GraphQL
Continue reading to learn how we got here, and why it's so exciting, but if you want to get started, here is how.

You can opt-in to use DGS/Spring-GraphQL by replacing the starter dependency.

Replace `implementation "com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter"` with `implementation "com.netflix.graphql.dgs:graphql-dgs-spring-graphql-starter"`.
You also need to add either Spring WebMVC or Spring WebFlux explicitly.

With this integration, it is technically possible to mix and match the DGS/Spring-GraphQL programming models. However, to maintain consistency in your codebase and to take full advantage of DGS features, we recommend sticking with the DGS programming model. Additional features from spring-graphql will be available via existing spring-graphql extensions, such as [multipart-spring-graphql](https://github.com/nkonev/multipart-spring-graphql) and the [Subscription Callback mechanism](https://github.com/apollographql/federation-jvm/pull/354) in the JVM Federation library.

## Background - Two competing frameworks
The DGS Framework provides Java developers with a programming model on top of Spring Boot to create GraphQL services. Netflix open-sourced the DGS framework in 2021, and has been the widely adopted GraphQL Java framework by many companies.  

While we continued adoption and development on the DGS Framework, the team at Spring were also exploring GraphQL support for Spring Boot. This effort was already underway prior to open-sourcing the DGS framework, while the Spring GraphQL efforts were also yet to be announced. After open sourcing the framework, both teams convened to discuss the current state and future direction. At the time Spring GraphQL was not as feature rich in comparison to the DGS Framework and we both  decided to continue what we were doing. While both the frameworks started with different goals and approaches, over the past year, Spring-GraphQL has matured and reached feature parity with many aspects of the DGS framework. This resulted in two "competing" frameworks in the community that largely solved the same problem. 

Today, new users must choose between one or the other, raising questions about which framework to adopt. It might also mean missing out on features available in one framework but not the other. This is not an ideal situation for the Java/GraphQL community.

For the maintainers of both frameworks, it would be much more efficient to collaborate on features and improvements instead of having to solve the same problem twice. Notably, bringing the communities together is a highly desirable goal!

### Why not just EOL one of the frameworks?
The DGS framework is widely used and plays a vital role in the architecture of many companies, including Netflix. Moving away from the framework in favor of Spring-GraphQL would be a costly migration without any real benefits. 

Although Spring-GraphQL doesn't have the user base of DGS, it makes sense from a Spring Framework perspective to have an out-of-the-box GraphQL offering, just like Spring supports REST. 

## The way forward
@srinivasankavitha and @paulbakker from the DGS team have worked together with Rossen Stoyanchev and Brian Clozel from the Spring-GraphQL team on integration between the frameworks. Special thanks to @kilink for being our first early adopter and contributing many bug fixes and performance improvements as part of this effort. 

With this integration, you can pull in additional features from Spring GraphQL. We also eliminate the need for part of the DGS code that integrates the framework with Spring MVC/WebFlux, since there isn't much benefit to duplicating low level functionality.

For the near term, the DGS/Spring-Graphql integration will be available as an opt-in feature via a different spring-graphql flavor of the DGS starter. We plan to make this the default mode in the long term after we see some level of successful adoption.

## Technical implementation
Both DGS and Spring-GraphQL are designed with modularity and extensibility in mind. This makes it feasible to integrate the two frameworks. The following diagrams show how the frameworks are integrated at a high level.

Today, the DGS framework looks as follows.
A user uses the DGS programming model, such as `@DgsComponent` and `@DgsQuery` to define data fetchers etc.
A GraphQL query comes in through either WebMVC or WebFlux, and is executed by the `DgsQueryExecutor`, which builds on the graphql-java library to execute the query.
The `DgsQueryExecutor` is also used directly when writing tests.
image

With the Spring-GraphQL integration, a user can write code with both the DGS programming model and/or the Spring-GraphQL programming model. A GraphQL query comes in through WebMVC/WebFlux/RSocket, which Spring-GraphQL now handles. The representation of the schema (a `GraphQLSchema` from graphql-java) is created by the DGS framework, and used by both the DGS and Spring-GraphQL components. Spring-GraphQL's `ExecutionGraphQLService` now handles the actual query execution, while the DGS `QueryExecutor` becomes a proxy on top of `ExecutionGraphQLService` so that existing test code continues to work.
image

## Required Changes
The good news is that the new integration has been mostly a drop-in replacement, not requiring any major code changes for the user.  If you use file uploads, you will need to explicitly add  a new dependency [multipart-spring-graphql](https://github.com/nkonev/multipart-spring-graphql).

## Known Gaps
At this time, we are lacking support for SSE based subscriptions. This is on the roadmap and will be made available depending on support in spring-graphql. 

## Performance
At Netflix, we tested the DGS/Spring-GraphQL integration on some of our largest services. Surprisingly, we uncovered a few performance issues in Spring WebMVC/Spring GraphQL with this integration. The Spring team has quickly addressed these issues, and the performance is now even better compared to baseline performance of Netflix applications with just the regular DGS Framework.

## Configuration

There is some overlap between configuration properties for DGS and Spring-GraphQL. Where properties overlap, we use the DGS property for the best backward compatibility. The following list is the overlapping properties. 

| *DGS property* | *Spring-GraphQL property* | *What to use* |
|----|----| ----- |
| `dgs.graphql.schema-locations` | `spring.graphql.schema.locations` |  Use `dgs.graphql.schema-locations` |
| N/A | `spring.graphql.schema.fileExtensions` | Not applicable, because `dgs.graphql.schema-locations` includes the path |
| `dgs.graphql.graphiql.enabled` | `spring.graphql.graphiql.enabled` | Use `dgs.graphql.graphiql.enabled` |
| `dgs.graphql.graphiql.path` | `spring.graphql.graphiql.path` | Use `dgs.graphql.graphiql.path` |
| `dgs.graphql.websocket.connection-init-timeout` | `spring.graphql.websocket.connection-init-timeout` | DGS property sets the Spring-GraphQL property |


graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/DgsGraphQLSourceBuilder.kt
```diff
@@ -0,0 +1,164 @@
...
+import graphql.schema.idl.TypeDefinitionRegistry
+import org.springframework.core.io.Resource
+import org.springframework.graphql.execution.*
+import org.springframework.graphql.execution.GraphQlSource.SchemaResourceBuilder
+import org.springframework.lang.Nullable
+import java.util.*
+import java.util.function.BiFunction
+import java.util.function.Consumer
+
+class DgsGraphQLSourceBuilder(private val dgsSchemaProvider: DgsSchemaProvider, private val reloadSchemaIndicator: DefaultDgsQueryExecutor.ReloadSchemaIndicator) : SchemaResourceBuilder {
```
rstoyanchev, line 36: The `GraphQlSource.Builder` hierarchy has 2 alternatives. `GraphQlSource.Builder` itself exposes options independent of how the `GraphQLSchema` is created. `SchemaResourceBuilder` on the other hand is for builders that create the `GraphQLSchema` by parsing schema files, and building a `RuntimeWiring`. On the implementation side, `AbstractGraphQlSourceBuilder` has the common parts, while `DefaultSchemaResourceGraphQlSourceBuilder` implements `SchemaResouceBuilder` and prepares `GraphQLSchema` from those inputs, while 

This was originally created for SDL vs programmatic API in GraphQL Java, so if anyone wants to build `GraphQLSchema` programmatically they can extend `AbstractGraphQlSourceBuilder`. The same applies to DGS as well I think, and even though it also parses schema files and does things the SDL way, it has its own way of building the `GraphQLSchema` so it makes to extend `AbstractGraphQlSourceBuilder` and override the `initGraphQlSchema` method to call `dgsSchemaProvider.schema()`. So that will become a very trivial class and it won't have to repeat anything from `AbstractGraphQlSourceBuilder` as it currently does.

Likewise, I don't think `DgsGraphQLSourceBuilder` should have to be an implementation of `SchemaResourceBuilder`, but rather it only needs to be an `AbstractGraphQlSourceBuilder` which implements the base `GraphQlSource.Builder.` As it is currently, the builder doesn't have anything meaningful it can do with the extra options it gets from implementing the `SchemaResourceBuilder` subtype.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/DgsGraphQLSourceBuilder.kt
```diff
@@ -0,0 +1,164 @@
...
+        return this
+    }
+
+    override fun configureGraphQl(configurer: Consumer<GraphQL.Builder>): SchemaResourceBuilder {
+        graphQlConfigurers = this.graphQlConfigurers.andThen(configurer)
+        return this
+    }
+
+    override fun build(): GraphQlSource {
+        return ReloadableGraphQLSource(reload(), this, reloadSchemaIndicator)
```
rstoyanchev, line 85: One specific thing, if you are to extend `AbstractGraphQlSourceBuilder` instead, obviously that creates the very simple `FixedGraphQlSource` while what you want is one that resolves at runtime, possibly by reloading. This can be achieved by changing `ReloadableGraphQLSource` a little to be created with a `GraphQlSource.Builder` rather than with a `GraphQlSource`. That way it can call build again whenever it wants. 

This should be a fairly small refactoring that will keep the builder focused on building a single instance, and `ReloadableGraphQlSource` responsible for all that has to do with reloading. We could also consider adding a protected method to create the `GraphQlSource` instance if needed.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/DgsGraphQLSourceBuilder.kt
```diff
@@ -0,0 +1,164 @@
...
+        return this
+    }
+
+    override fun defaultTypeResolver(typeResolver: TypeResolver): SchemaResourceBuilder {
+        this.typeResolver = typeResolver
+        return this
+    }
+
+    override fun inspectSchemaMappings(reportConsumer: Consumer<SchemaReport>): SchemaResourceBuilder {
+        this.schemaReportConsumer = reportConsumer
```
rstoyanchev, line 157: The SchemaReport is also worth a mention. What's required to run it is a `GraphQLSchema` and a `RuntimeWiring` and it's currently run from `DefaultSchemaResourceGraphQlSourceBuilder`, and since you are replacing that, you'll need to invoke it from this builder implementation. 

As part of https://github.com/spring-projects/spring-graphql/issues/878, I've added a protected method on `AbstractGraphQlSourceBuilder` that lets you hook in at the point where `GraphQL.Builder` is customized and that is where we invoke the schema report if the application has explicitly provided a `SchemaReport` consumer.  

If you look at the commit attached to that issue, you can see how it's done. DGS can do something similar but would need to obtain the resulting `RuntimeWiring` from `DgsSchemaProvider` somehow. Mainly what we need from the `RuntimeWiring` is the map of DataFetchers's if it helps to know that.

rstoyanchev:
Thanks for all the work so far! 

In addition to the excellent background, I would like to add what I am most excited about is unifying the community. Not only for contributions in the way of code, but even more importantly in terms of reporting issues, providing feedback and ideas. It's what drives an open source project.

rstoyanchev (commented):
Some initial feedback below, by no means an exhaustive review, but there should be plenty for discussion and an iteration.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/DgsGraphQLSourceBuilder.kt
srinivasankavitha: Thanks for the comments. What you suggest makes sense to us. Looking at the latest sources, AbstractGraphQlSourceBuilder is not public. Could you update that so we can use it?

rstoyanchev: You're right. I think we started out with just one package private builder before the hierarchy evolved. It does make sense for the base class to be public as it's more broadly useful, and I've made that change https://github.com/spring-projects/spring-graphql/commit/5fcf803680494f642b04f864c3179a34fb45ab25.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/DgsGraphQLSourceBuilder.kt
```diff
@@ -0,0 +1,89 @@
...
+import graphql.schema.idl.RuntimeWiring
+import graphql.schema.idl.TypeDefinitionRegistry
+import org.springframework.core.io.Resource
+import org.springframework.graphql.execution.*
+import org.springframework.graphql.execution.GraphQlSource.SchemaResourceBuilder
+import org.springframework.lang.Nullable
+import java.util.function.BiFunction
+import java.util.function.Consumer
+
+class DgsGraphQLSourceBuilder(private val dgsSchemaProvider: DgsSchemaProvider, private val reloadSchemaIndicator: DefaultDgsQueryExecutor.ReloadSchemaIndicator) : AbstractGraphQlSourceBuilder<SchemaResourceBuilder>(), SchemaResourceBuilder {
```
rstoyanchev, line 32:
```suggestion
class DgsGraphQLSourceBuilder(private val dgsSchemaProvider: DgsSchemaProvider, private val reloadSchemaIndicator: DefaultDgsQueryExecutor.ReloadSchemaIndicator) : AbstractGraphQlSourceBuilder<DgsGraphQLSourceBuilder>() {
```

Looks good extending `AbstractGraphQlSourceBuilder`! You can now also drop the `SchemaResourceBuilder` sub-interface with its methods related to building `GraphQLSchema`. In this implementation that's done through `DgsSchemaProvider`, and they are not relevant.

srinivasankavitha: Thanks for taking a look! Actually we realized we still need the SchemaResourceBuilder because third party libraries (multipart file uploads that you recommend as well)set up a schema resource on the builder. So we do need to inject schema resources from the builder into the DgsSchemaProvider prior to calling it.

rstoyanchev: Right, I see now. I was also going to bring up something related to this. 

`DgsSpringGraphQLAutoConfiguration` is declared to run before `DgsAutoConfiguration` (I presume so it can declare `DgsComponent`s for the latter to detect). At the same time, the `graphQlSource` bean  needs `DgsSchemaProvider` that comes from the latter, but that's supposed to run after. There is a sort of circular relationship there, and while it doesn't seem to cause a problem currently, I think it would be good to split the config in two distinct parts.

The first part focuses on `GraphQLSchema` creation, declares `@DgsComponent`s, and is ordered before `DgsAutoConfiguration`. The second part creates the `graphQlSource` bean and is ordered after  `DgsAutoConfiguration` with the `DgsSchemaProvider`. The first config can detect `GraphQlSourceBuilderCustomizer`s such as the one from FileUpload library, and bridge any `SchemaResourceBuilder` registrations to `DgsSchemaProvider` along the lines of what `DgsRuntimeWiringConfigurerBridge` and `DgsTypeDefinitionConfigurerBridge` currently do. For the second config, the `DgsGraphQLSourceBuilder` uses `DgsSchemaProvider` and implements only the base `GraphQlSource.Builder`.

srinivasankavitha (approved):

PR #1770, opened 2024-01-08
Provide consistent support for dynamically registered DataFetchers

ivanasen:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

Implemented a new way of using `@DgsCodeRegistry` in which a `DgsCodeRegistryBuilder` is injected instead. This way, the registered DataFetchers are processed by DataFetcherResultProcessors and the behavior is more consistent with the annotation-based approach.

[Link to discussion](https://github.com/Netflix/dgs-framework/discussions/1768)

Alternatives considered
----

Haven't considered alternatives.


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
```diff
@@ -477,29 +481,66 @@ internal class DgsSchemaProviderTest {
                 val coordinates = FieldCoordinates.coordinates("Query", "myField")
                 return codeRegistryBuilder.dataFetcher(coordinates, df)
             }
+
+            // Result should be processed by DataFetcherResultProcessors
+            @DgsCodeRegistry
```
srinivasankavitha, line 486: Would you be able to also update our docs to use the DgsCodeRegistryBuilder? Much appreciated.

srinivasankavitha (changes requested):
Thanks for the PR contribution. Changes look good, just had a few comments.

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsCodeRegistryBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+import graphql.schema.GraphQLFieldDefinition;
+
+import java.util.List;
+
+/**
+ * Utility wrapper for {@link GraphQLCodeRegistry.Builder} which provides
+ * a consistent registration mechanism of DataFetchers similar to the annotation-based approach.
+ * Can be used as a first parameter of a {@link DgsCodeRegistry} annotated method.
+ */
+public class DgsCodeRegistryBuilder {
```
kilink, line 33: Is there a reason this class is written in Java instead of Kotlin?

ivanasen: Because the other classes in this package are also in Java. I suppose this package is different from the "internal" package where there is Kotlin

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
ivanasen: Ack

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -261,18 +261,24 @@ class DgsSchemaProvider(
...
-                    throw InvalidDgsConfigurationException("Method annotated with @DgsCodeRegistry must have return type GraphQLCodeRegistry.Builder")
+                if (method.returnType != GraphQLCodeRegistry.Builder::class.java && method.returnType != DgsCodeRegistryBuilder::class.java) {
+                    throw InvalidDgsConfigurationException("Method annotated with @DgsCodeRegistry must have return type GraphQLCodeRegistry.Builder or DgsCodeRegistryBuilder")
                 }
 
-                if (method.parameterCount != 2 || method.parameterTypes[0] != GraphQLCodeRegistry.Builder::class.java || method.parameterTypes[1] != TypeDefinitionRegistry::class.java) {
-                    throw InvalidDgsConfigurationException("Method annotated with @DgsCodeRegistry must accept the following arguments: GraphQLCodeRegistry.Builder, TypeDefinitionRegistry. ${dgsComponent.javaClass.name}.${method.name} has the following arguments: ${method.parameterTypes.joinToString()}")
+                if (method.parameterCount != 2 || method.parameterTypes[0] != method.returnType || method.parameterTypes[1] != TypeDefinitionRegistry::class.java) {
```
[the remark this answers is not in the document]
ivanasen: Do you mean `GraphQLCodeRegistry.Builder` ? If yes, it does

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsCodeRegistryBuilder.java
kilink: That's not really the case, the things in Java are mostly interfaces and annotations, and are there due to compatibility issues with older versions of Kotlin.

ivanasen:
I was looking at the code today in `DgsSchemaProvider.kt` and I actually think that in order to really be consistent we need to not only account for the `DataFetcherResultProcessors` in the new `DgsCodeRegistryBuilder` . In order for it to be really consistent, shouldn't we account for most of the logic in `DgsSchemaProvider::registerDataFetcher`? Also, I shouldn't be duplicating logic here. Registering DataFetchers through annotations and dynamically should use common logic. So in that case, we should refactor out the `DgsSchemaProvider::registerDataFetcher` (or most parts of it) in some common place which can be invoked from both `DgsSchemaProvider` and `DgsCodeRegistryBuilder`.

I won't have time to make this change in the next 2-3 days, but will most likely get to it in the weekend

graphql-dgs/src/main/java/com/netflix/graphql/dgs/DgsCodeRegistryBuilder.java
ivanasen: Interesting. I will look at how to rewrite in Kotlin then. Thanks

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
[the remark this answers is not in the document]
ivanasen: In practice, the check `method.parameterTypes[0] != method.returnType` , but that's kinda quirky, so I'll refactor it to make an explicit check for the 2 classes

srinivasankavitha:
> I was looking at the code today in `DgsSchemaProvider.kt` and I actually think that in order to really be consistent we need to not account for only the `DataFetcherResultProcessors` in the new `DgsCodeRegistryBuilder` . In order for it to be really consistent, shouldn't we account for most of the logic in `DgsSchemaProvider::registerDataFetcher`? Also, I shouldn't be duplicating logic here. Registering DataFetchers through annotations and dynamically should use common logic. So in that case, we should refactor out the `DgsSchemaProvider::registerDataFetcher` (or most parts of it) in some common place which can be invoked from both `DgsSchemaProvider` and `DgsCodeRegistryBuilder`.
> 
> I won't have time to make this change in the next 2-3 days, but will most likely get to it in the weekend

Ok. I don't mind the current solution in any case. It's really not much duplication.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
ivanasen: Actually, where are the docs :smile: ?

ivanasen:
> > I was looking at the code today in `DgsSchemaProvider.kt` and I actually think that in order to really be consistent we need to not account for only the `DataFetcherResultProcessors` in the new `DgsCodeRegistryBuilder` . In order for it to be really consistent, shouldn't we account for most of the logic in `DgsSchemaProvider::registerDataFetcher`? Also, I shouldn't be duplicating logic here. Registering DataFetchers through annotations and dynamically should use common logic. So in that case, we should refactor out the `DgsSchemaProvider::registerDataFetcher` (or most parts of it) in some common place which can be invoked from both `DgsSchemaProvider` and `DgsCodeRegistryBuilder`.
> > I won't have time to make this change in the next 2-3 days, but will most likely get to it in the weekend
> 
> Ok. I don't mind the current solution in any case. It's really not much duplication.

Ok, if I'm gonna make more changes better make them in pieces. For this commit, I propose we just handle this.

srinivasankavitha:
> > > I was looking at the code today in `DgsSchemaProvider.kt` and I actually think that in order to really be consistent we need to not account for only the `DataFetcherResultProcessors` in the new `DgsCodeRegistryBuilder` . In order for it to be really consistent, shouldn't we account for most of the logic in `DgsSchemaProvider::registerDataFetcher`? Also, I shouldn't be duplicating logic here. Registering DataFetchers through annotations and dynamically should use common logic. So in that case, we should refactor out the `DgsSchemaProvider::registerDataFetcher` (or most parts of it) in some common place which can be invoked from both `DgsSchemaProvider` and `DgsCodeRegistryBuilder`.
> > > I won't have time to make this change in the next 2-3 days, but will most likely get to it in the weekend
> > 
> > 
> > Ok. I don't mind the current solution in any case. It's really not much duplication.
> 
> Ok, if I'm gonna make more changes better make them in pieces. For this commit, I propose we just handle this.

Sounds good. @kilink - any objections to merging this PR?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsCodeRegistryBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+                val env = DgsDataFetchingEnvironment(dfe)
+                dataFetcherResultProcessors.find { it.supportsType(result) }?.process(result, env) ?: result
+            }
+        }
+
+        graphQLCodeRegistry.dataFetcher(coordinates, wrapped)
+        return this
+    }
+
+    fun hasDataFetcher(coordinates: FieldCoordinates?): Boolean {
```
kilink, line 47: I'd make the parameter non-nullable.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsCodeRegistryBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+        return graphQLCodeRegistry.hasDataFetcher(coordinates)
+    }
+
+    fun getDataFetcher(coordinates: FieldCoordinates?, fieldDefinition: GraphQLFieldDefinition?): DataFetcher<*> {
```
kilink, line 51: I'd make both parameters non-nullable, I am not sure passing in a null makes sense at all, so may as well disallow it.

kilink:
> > > > I was looking at the code today in `DgsSchemaProvider.kt` and I actually think that in order to really be consistent we need to not account for only the `DataFetcherResultProcessors` in the new `DgsCodeRegistryBuilder` . In order for it to be really consistent, shouldn't we account for most of the logic in `DgsSchemaProvider::registerDataFetcher`? Also, I shouldn't be duplicating logic here. Registering DataFetchers through annotations and dynamically should use common logic. So in that case, we should refactor out the `DgsSchemaProvider::registerDataFetcher` (or most parts of it) in some common place which can be invoked from both `DgsSchemaProvider` and `DgsCodeRegistryBuilder`.
> > > > I won't have time to make this change in the next 2-3 days, but will most likely get to it in the weekend
> > > 
> > > 
> > > Ok. I don't mind the current solution in any case. It's really not much duplication.
> > 
> > 
> > Ok, if I'm gonna make more changes better make them in pieces. For this commit, I propose we just handle this.
> 
> Sounds good. @kilink - any objections to merging this PR?

LGTM, just had 2 minor comments about nullability of parameters.

ivanasen:
Thanks, I updated the PL with the suggested changes. I haven't received an answer to [this](https://github.com/Netflix/dgs-framework/pull/1770#discussion_r1452737561) question though

srinivasankavitha:
> Thanks, I updated the PR with the suggested changes. I haven't received an answer to [this](https://github.com/Netflix/dgs-framework/pull/1770#discussion_r1452737561) question though

Think we decided to keep things as is for now. We can do a refactor in a separate PR in the future, but I'm also fine leaving things the way they are for now.

PR #1779, opened 2024-01-15
Inject schema resources from spring-graphql into DgsSchemaProvider.

srinivasankavitha:
Changes in this PR
----
Inject schema resources from spring-graphql into DgsSChemaResourceProvider so those can be parsed as well.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -143,17 +143,18 @@ class DgsSchemaProvider(
 
     fun schema(
         @Language("GraphQL") schema: String? = null,
+        schemaResources: Set<Resource>? = null,
         fieldVisibility: GraphqlFieldVisibility = DefaultGraphqlFieldVisibility.DEFAULT_FIELD_VISIBILITY
     ): GraphQLSchema {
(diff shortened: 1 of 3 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -143,17 +143,18 @@ class DgsSchemaProvider(
...
             dataFetchers.clear()
             dataFetcherTracingInstrumentationEnabled.clear()
             dataFetcherMetricsInstrumentationEnabled.clear()
-            return computeSchema(schema, fieldVisibility)
+            return computeSchema(schema, schemaResources, fieldVisibility)
         }
     }
 
-    private fun computeSchema(schema: String? = null, fieldVisibility: GraphqlFieldVisibility): GraphQLSchema {
+    private fun computeSchema(schema: String? = null, schemaResources: Set<Resource>? = emptySet(), fieldVisibility: GraphqlFieldVisibility): GraphQLSchema {
```
paulbakker, line 157: minor:
1. Don't make it nullable - it can just be an empty set by default, so we don't have to null check.
2. Make it the third argument. The existing fieldVisibility argument is non-optional. If this new one is the last argument and it has a default value, you don't *have* to specify it, so for example the DgsAutoConfiguration can stay unchanged.

paulbakker (approved):

PR #1786, opened 2024-01-19
Consistently handle proxied beans in DgsSchemaProvider

### LINKED ISSUE #1163: bug: @DgsCodeRegistry not applied if bean is wrapped by Spring proxy
## Expected behavior
A method annotated with `@DgsCodeRegistry` in a Spring bean that is wrapped by a Spring CGLIB proxy should be applied.

## Actual behavior
The method is ignored an no application of the `@DgsCodeRegistry` takes place

## Root Cause
This issue is the used way of iterating over the methods in the bean at https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L220
This is not supporting beans wrapped by a proxy which in turn makes the code to miss the annotation.
The code should use the same procedure as done in other places of the class like https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L390-L392

Additionally be aware that the same issue is likely also found at
* https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L130
* https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L201
* https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L236
* https://github.com/Netflix/dgs-framework/blob/fe6ca59c04d074a4c3f5dffc2aeceb3e8aeaf4df/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L413

kilink:
Ensure that we are consistently handling proxied beans in DgsSchemaProvider by refactoring some of the logic to retrieve annotated methods; after fetching DgsComponent-annotated beans, use a new internal wrapper class, DgsBean, that correctly handles proxies, and includes a helper method to retrieve all methods decorated with a particular annotation. In addition to making the handling consistent, moving the logic to the DgsBean wrapper class avoids repeated unwrapping of proxies and allocations.

Fixes #1163

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
@@ -124,7 +124,7 @@ class DgsSchemaProvider(
      */
     fun isFieldTracingInstrumentationEnabled(field: String): Boolean {
         return schemaReadWriteLock.read {
-            dataFetcherTracingInstrumentationEnabled.getOrDefault(field, true)
+            dataFetcherTracingInstrumentationEnabled[field] ?: true
         }
     }
 
@@ -137,7 +137,7 @@ class DgsSchemaProvider(
      */
     fun isFieldMetricsInstrumentationEnabled(field: String): Boolean {
         return schemaReadWriteLock.read {
-            dataFetcherMetricsInstrumentationEnabled.getOrDefault(field, true)
+            dataFetcherMetricsInstrumentationEnabled[field] ?: true
         }
     }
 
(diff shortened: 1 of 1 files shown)


srinivasankavitha (approved):
Was there an issue that this PR fixes or just general improvement?

kilink:
> Was there an issue that this PR fixes or just general improvement?

There was an issue reported with proxies here: #1163

PR #1788, opened 2024-01-20
Avoid wrapping checked Exceptions in DataFetcherInvoker

### LINKED ISSUE #1098: bug: Exceptions thrown from dataFetchers are being thrown as `java.lang.reflect.UndeclaredThrowableException`.
## Expected behavior
`DataFetcherExceptionHandler#onException`handlerParameters Will contain the thrown exception.

## Actual behavior
`DataFetcherExceptionHandler#onException`handlerParameters contains `java.lang.reflect.UndeclaredThrowableException`, it's cause is the right exception. 

## Steps to reproduce
1. Create a simple query.
2. Create a test class for it
3. mocked class will throw a custom Exception (inherited from Exception)
4. set a breakPoint `DataFetcherExceptionHandler#onException`, see that the exception is `java.lang.reflect.UndeclaredThrowableException`

Note:
DataFetcher does throw the right exception and declares about it.

kilink:
DataFetcherInvoker implements the DataFetcher interface, and its "get" method declares that it throws Exception, which means propagating checked exceptions is not a problem. Howver, we were leveraging the Spring helper method, ReflectionUtils.handleReflectionException, which is designed to be called from methods that do not throw any checked exceptions, so we were ending up wrapping them unnecessarily in UndeclaredThrowableException. This caused an unnecessary layer of indirection that would need to be dealt with for example in a DataFetcherExceptionHandler.

To avoid the problem, stop using the Spring utility to invoke methods reflectively in DataFetcherInvoker, and leverage our own helper method that never throws UndeclaredThrowableException.

Fixes #1098

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
@@ -31,6 +31,7 @@ import org.springframework.util.CollectionUtils
 import org.springframework.util.ReflectionUtils
 import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
+import java.lang.reflect.UndeclaredThrowableException
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.callSuspendBy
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DataFetcherInvoker.kt
```diff
@@ -113,19 +123,40 @@ class DataFetcherInvoker internal constructor(
...
+    private fun handleReflectionException(exc: Exception): Nothing {
+        if (exc is NoSuchMethodException) {
+            throw IllegalStateException("Method not found: ${exc.message}")
+        }
+        if (exc is IllegalAccessException) {
+            throw IllegalStateException("Could not access method or field: ${exc.message}")
+        }
+        if (exc is InvocationTargetException) {
+            throw exc.targetException
+        }
```
srinivasankavitha, line 159: Don't we need to handle RunTimeException case as they do in ReflectionUtils?

kilink: Runtime exceptions will just get rethrown, which is the same behavior as ReflectionUtils; the reason they have special handling of RuntimeException is that they need to filter out RuntimeException so they can wrap checked Exceptions in UndeclaredThrowableException. Since we don't want that behavior at all we don't need the extra logic.

PR #1793, opened 2024-01-22
Use QueryComplexityCalculator from graphql-java

kilink:
Switch to using QueryComplexityCalculator from graphql-java in DgsGraphQLMetricsInstrumentation. Additionally, eliminate some unnecessary usage of Optional.

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
@@ -11,9 +11,9 @@ import graphql.ExecutionInput
 import graphql.ExecutionResult
 import graphql.GraphQLError
 import graphql.InvalidSyntaxError
-import graphql.analysis.QueryTraverser
-import graphql.analysis.QueryVisitorFieldEnvironment
-import graphql.analysis.QueryVisitorStub
+import graphql.analysis.FieldComplexityCalculator
+import graphql.analysis.QueryComplexityCalculator
+import graphql.execution.ExecutionContext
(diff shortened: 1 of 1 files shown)


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -276,49 +279,31 @@ class DgsGraphQLMetricsInstrumentation(
 
     internal object ComplexityUtils {
 
+        private val complexityCalculator: FieldComplexityCalculator = FieldComplexityCalculator { _, childComplexity -> childComplexity + 1 }
+
         private val queryComplexityBuckets = listOf(5, 10, 25, 50, 100, 200, 500, 1000, 2000, 5000, 10000)
```
srinivasankavitha, line 285: We had to use our own implementation of query complexity to avoid running into a bug with the calculator in graphql-java at that time (it did not dealt with fragments properly or something like that, but was a corner case). Hoping that is fixed now.

kilink: It looks like `QueryComplexityCalculator` was only added recently (v21.3), and used to be part of `MaxQueryComplexityInstrumentation` (see [this commit](https://github.com/graphql-java/graphql-java/commit/dba84dbd852bd4ec514c2ce76e7874243ef0549f#diff-94ace4aff6bcfcb501a55d5dee0a9e201303c208eda07c5288fd7f3f0bd654f3)). The actual implementation of `QueryComplexityCalculator` looks pretty much the same as the logic we had in `DgsGraphQLMetricsInstrumentation`. Is there a test case covering the fragment issue, or do you have any more details about the problem?

kilink:
I opened a new PR #1797 since the commit author was getting messed up on this one.

PR #1796, opened 2024-01-23
Use Jackson2ObjectMapperBuilder to construct Jackson mapper

### LINKED ISSUE #1461: feature: Support to inject ObjectMapper into CustomGraphQLClient
## Describe the Feature Request
If you put the time value in the input as shown below.

```kotlin
val client = GraphQLClient.createCustom("http://localhost:8080/graphql") { url, headers, body ->
    val httpHeaders = HttpHeaders()
    headers.forEach { httpHeaders.addAll(it.key, it.value) }
    httpHeaders.set(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE)

    val exchange = RestTemplate().exchange(
        url,
        HttpMethod.POST,
        HttpEntity(body, httpHeaders),
        String::class.java,
    )

    HttpResponse(exchange.statusCodeValue, exchange.body)
}

val mutation = "some mutation"

val input = FooInput(
    name = "test",
    date = LocalDate.of(2023, 1, 1),
)
val variables = mapOf("input" to input)


client.executeQuery(mutation, variables)
    .extractValueAsObject("foo", typeRef<FooField>())
```

An error will occur.

```
Java 8 date/time type `java.time.LocalDate` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.netflix.graphql.dgs.client.Request["variables"]->java.util.Collections$SingletonMap["input"]->com.spoqa.cart.generated.types.CreateOrderSheetInput["date"])
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.LocalDate` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.netflix.graphql.dgs.client.Request["variables"]->java.util.Collections$SingletonMap["input"]-
... 
```

So I had to take action as below.

```kotlin
val input = FooInput(
    name = "test",
    date = LocalDate.of(2023, 1, 1),
)

val objectMapper = jacksonObjectMapper().apply { registerModule(JavaTimeModule()) }
val variables = mapOf("input" to objectMapper.readValue(objectMapper.writeValueAsString(input)))
```

I found the following settings in the [DGS document](https://netflix.github.io/dgs/advanced/custom-object-mapper/).

```java
@Bean
@Qualifier("dgsObjectMapper")
public ObjectMapper dgsObjectMapper() {
    ObjectMapper customMapper = new ObjectMapper()
    customMapper.registerModule(JavaTimeModule());
    return customMapper;
}
```

However, this setting does not work. Because of the code below.

```java
internal object GraphQLClients {

    internal val objectMapper: ObjectMapper =
        if (ClassUtils.isPresent("com.fasterxml.jackson.module.kotlin.KotlinModule\$Builder", this::class.java.classLoader)) {
            ObjectMapper().registerModule(KotlinModule.Builder().nullIsSameAsDefault(true).build())
        } else ObjectMapper().registerKotlinModule()
    // ...
}
```

```java
class CustomGraphQLClient(private val url: String, private val requestExecutor: RequestExecutor) : GraphQLClient {
    //...
    override fun executeQuery(query: String, variables: Map<String, Any>, operationName: String?): GraphQLResponse {
        val serializedRequest = GraphQLClients.objectMapper.writeValueAsString(
            Request(
                query,
                variables,
                operationName
            )
        )

        val response = requestExecutor.execute(url, GraphQLClients.defaultHeaders, serializedRequest)
        return GraphQLClients.handleResponse(response, serializedRequest, url)
    }
}
```

The `objectMapper` cannot be replaced by redefining bean. Because it is generated by GraphQLClients.
In my opinion, it seems that the problem can be solved if the object mapper can be injected dependently.

kilink:
Switch to using Spring's Jackson2ObjectMapperBuilder to construct the Jackson ObjectMapper instance in GraphQLClients and DgsWebMvcAutoConfiguration, since the builder will automatically install well-known modules (such as JavaTimeModule) if they are present on the class path.

Fixes #1461

### DIFF
--- a/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt
+++ b/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt
@@ -18,7 +18,7 @@ package com.netflix.graphql.dgs.webmvc.autoconfigure
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
-import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.KotlinModule
 import com.netflix.graphql.dgs.DgsQueryExecutor
 import com.netflix.graphql.dgs.internal.DgsSchemaProvider
 import com.netflix.graphql.dgs.internal.method.ArgumentResolver
@@ -42,6 +42,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.context.annotation.Import
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder
 import org.springframework.web.bind.support.WebDataBinderFactory
 import org.springframework.web.method.annotation.RequestHeaderMapMethodArgumentResolver
 import org.springframework.web.method.annotation.RequestHeaderMethodArgumentResolver
@@ -60,7 +61,9 @@ open class DgsWebMvcAutoConfiguration {
     @Qualifier("dgsObjectMapper")
     @ConditionalOnMissingBean(name = ["dgsObjectMapper"])
     open fun dgsObjectMapper(): ObjectMapper {
-        return jacksonObjectMapper().registerModule(JavaTimeModule())
+        return Jackson2ObjectMapperBuilder.json()
+            .modulesToInstall(KotlinModule.Builder().build(), JavaTimeModule())
+            .build()
     }
 
     @Bean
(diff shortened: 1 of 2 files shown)


graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt
```diff
@@ -60,7 +61,9 @@ open class DgsWebMvcAutoConfiguration {
     @Qualifier("dgsObjectMapper")
     @ConditionalOnMissingBean(name = ["dgsObjectMapper"])
     open fun dgsObjectMapper(): ObjectMapper {
-        return jacksonObjectMapper().registerModule(JavaTimeModule())
+        return Jackson2ObjectMapperBuilder.json()
```
srinivasankavitha, line 64: So it seems to install kotlin modules, but not JavaTimeModule, correct? https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.html

kilink: It does install `JavaTimeModule`, see [here](https://github.com/spring-projects/spring-framework/blob/6bd7f0231dcc47e8099000dea301157788fd5b15/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java#L880). I think the Javadoc is just out of date.

kilink: Oh, the Javadoc refers to `jackson-datatype-jsr310`, which is the project the time module used to live in.

PR #1798, opened 2024-01-23
Allow connection directive to be applied to Unions and Interfaces

### LINKED ISSUE #1317: bug: Using @connection on interface fails
When I use pagination to create a relay connection on an interface, my spring application won't start. 

## Expected behavior
As far as I understand the class `DgsPaginationTypeDefinitionRegistry` it should be supported to add pagination for a type, interface and union:
```
types.filter { it is ObjectTypeDefinition || it is InterfaceTypeDefinition || it is UnionTypeDefinition }
            .filter { it.hasDirective("connection") }
            .forEach {
                definitions.add(createConnection(it.name))
                definitions.add(createEdge(it.name))
            }
```

## Actual behavior

```
***************************
APPLICATION FAILED TO START
***************************

Description:

There are problems with the GraphQL Schema:
	 * 'Animal' [@14:1] tried to use a directive 'connection' in the 'INTERFACE' location but that is illegal
```

## Steps to reproduce`
I created a simple schema and tried to start a spring application.

```
type Query {
    animals: AnimalConnection
}

interface Animal @connection {
    type: String
}

type Cat implements Animal @connection  {
    type: String
    name: String
}

type Dog implements Animal @connection {
    type: String
    size: Int
}
```

kilink:
DgsPaginationTypeDefinitionRegistry generated a connection directive which could only be applied to Object types; update it to allow Union and Interfaces as well.

There were tests for this scenario but they didn't catch the problem because they never built a GraphQLSchema. To avoid such issues, switch all of the tests to fully build a GraphQLSchema.

Fixes #1317.

### DIFF
--- a/graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
+++ b/graphql-dgs-pagination/src/main/kotlin/com/netflix/graphql/dgs/pagination/DgsPaginationTypeDefinitionRegistry.kt
@@ -40,7 +40,10 @@ class DgsPaginationTypeDefinitionRegistry {
             val directive = DirectiveDefinition.newDirectiveDefinition()
                 .name(CONNECTION_DIRECTIVE_NAME)
                 .description(createDescription("Connection"))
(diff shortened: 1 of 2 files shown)


srinivasankavitha (approved):
Good catch! Thanks for the fixes.

PR #1799, opened 2024-01-23
Fix DGS Micrometer timer configuration

### LINKED ISSUE #1249: bug: DGS metrics stop reporting Spring Boot 2.6.10, micrometer-core 1.88
Hi team,

We have projects that have upgrade to Spring Boot 2.6.10 and have seemingly lost all of the DGS loader, fetcher and query metrics.

It looks like we're at micrometer-core 1.8.8 now. Should this version of spring-boot/micrometer be compatible with `graphql-dgs-spring-boot-micrometer` ?

Thanks in advance, let me know if you need any more information.

kilink:
A recent update broke the documented autotime-related properties that live under "management.metrics.dgs-graphql.autotime", since it renamed the property on the DgsGraphQLMetricsProperties class from "autoTime" to "autoTimeProperties". Fix the naming to restore the old config prefix, and add a test case that covers it.

Additionally, move the AutoTimer / PropertiesAutoTimer out of DgsGraphQLMetricsProperties, and make it a dependency of DgsGraphQLMetricsInstrumentation and the other classes that depend on it.

Bug was introduced in 21e625a; this PR fixes #1249.


srinivasankavitha (approved):
Good catch! Thanks for fixing.

PR #1801, opened 2024-01-24
Add back getters returning Optional to MetricsInstrumentationState

kilink:
A previous change inadvertently broke API compatibility, since MetricsInstrumentationState is exposed externally; to restore compat, add getters that return Optional; move the nullable properties to properties suffixed with "Value" and with an internal visibility.


srinivasankavitha (approved):
This looks good to me! Thanks for the fix.

PR #1811, opened 2024-02-05
Make all timer threads daemonic in subscription handlers

kilink:
Make sure all the Timer instances use daemonic threads so they don't block JVM shutdown, and add a PreDestroy method to the implementations to proactively clean up the session watchdog Timer. Additionally, add a descriptive name to each Timer instance.


srinivasankavitha:
Should we uncomment the subscription tests as well to see if this fixes the hanging tests?

srinivasankavitha (approved):

kilink:
> Should we uncomment the subscription tests as well to see if this fixes the hanging tests?

Actually, the only disabled tests I can find are from `graphql-dgs-subscriptions-sse`, which I didn't touch in this PR.

PR #1818, opened 2024-02-15
add new hooks for instrumenting dataloaders

Emily:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----

add new hooks for instrumenting dataloaders
- DgsDataLoaderScanningInterceptor: allows for wrapping a dataloader as it is initially being registered, including changing the type of the data loader
- DgsSimpleDataloaderInstrumentation: an even easier way to add instrumentation like metrics, tracing, etc...

also note: it is likely that in the future this will be able to simplify the current micrometer metrics instrumentation that uses ByteBuddy.


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsWrapWithContextDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+        return MappedBatchLoaderWithContextWrapper(original)
+    }
+
+    override fun provide(original: MappedBatchLoaderWithContext<*, *>, name: String): Any {
+        return original
+    }
+}
+
+internal class BatchLoaderWithContextWrapper<K, V>(private val original: BatchLoader<K, V>) : BatchLoaderWithContext<K, V> {
+    override fun load(keys: MutableList<K>?, environment: BatchLoaderEnvironment?): CompletionStage<MutableList<V>> {
```
kilink, line 46: The parameters don't need to be nullable, and I don't think `keys` or the return type need to be mutable either.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,42 @@
...
+ * While this hook appears very similar to [com.netflix.graphql.dgs.DataLoaderInstrumentationExtensionProvider]
+ * there are two important differences:
+ * - DgsDataLoaderScanningInterceptors are called when scanning for @DgsDataLoaders, whereas the
+ *   DataLoaderInstrumentationExtensionProviders are called once per request
+ * - DgsDataLoaderScanningInterceptors are afforded the opportunity to change the type of each DataLoader
+ *   as it is being initially registered. Most notably, this allows for converting BatchLoader and MappedBatchLoader
+ *   into their "WithContext" versions, which can simplify certain types of DataLoader instrumentation.
+ */
+interface DgsDataLoaderScanningInterceptor {
+    fun provide(original: BatchLoader<*, *>, name: String): Any
```
kilink, line 38: Would it make sense for all of these to be default methods that do nothing? It would make it easier to implement and override just a single method.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+        return MappedBatchLoaderWithContextSimpleInstrumentationDriver(original, name, instrumentations)
+    }
+}
+
+internal class BatchLoaderWithContextSimpleInstrumentationDriver<K : Any, V>(
+    private val original: BatchLoaderWithContext<K, V>,
+    private val name: String,
+    private val instrumentations: List<DgsSimpleDataLoaderInstrumentation<*>>
+) : BatchLoaderWithContext<K, V> {
+    override fun load(keys: MutableList<K>, environment: BatchLoaderEnvironment): CompletionStage<MutableList<V>> {
```
kilink, line 54: I think it goes for all of these, but `keys` doesn't need to be mutable, nor does the type parameter of the return type.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+internal class MappedBatchLoaderWithContextSimpleInstrumentationDriver<K : Any, V>(
+    private val original: MappedBatchLoaderWithContext<K, V>,
+    private val name: String,
+    private val instrumentations: List<DgsSimpleDataLoaderInstrumentation<*>>
+) : MappedBatchLoaderWithContext<K, V> {
+    override fun load(keys: MutableSet<K>, environment: BatchLoaderEnvironment): CompletionStage<MutableMap<K, V>> {
+        val keysList = ArrayList(keys)
```
kilink, line 73: You can just call `keys.toList()`

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+internal class BatchLoaderWithContextSimpleInstrumentationDriver<K : Any, V>(
+    private val original: BatchLoaderWithContext<K, V>,
+    private val name: String,
+    private val instrumentations: List<DgsSimpleDataLoaderInstrumentation<*>>
+) : BatchLoaderWithContext<K, V> {
+    override fun load(keys: MutableList<K>, environment: BatchLoaderEnvironment): CompletionStage<MutableList<V>> {
+        val contexts = instrumentations.map { it.beforeLoad(name, keys, environment) }
```
kilink, line 55: Couldn't we just map this to a `Pair<DgsSimpleDataLoaderInstrumentation<*>, Context>` to avoid needing to zip it again below?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+
+        val future = original.load(keys, environment)
+
+        return instrumentations.zip(contexts).reversed().fold(future) { f, p ->
```
kilink, line 59: The chaining of operations here is producing a bunch of intermediate Lists that just get thrown away. Can get rid of the zip, and switching to `asReversed` gives you a reversed read-only list without making a copy. The other option is using a sequence at the start.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+            f.whenComplete { result, exception ->
```
kilink, line 60: I am pretty sure this doesn't behave correctly if the `afterLoad` methods happens to throw an Exception; it will make the returned future actually fail and the original result will be lost.

It might make more sense to just attach a single call back that loops through the instrumentations and calls them with a try / catch.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsSimpleDataLoaderInstrumentation.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import org.dataloader.BatchLoaderEnvironment
+
+/**
+ * Provides a simple interface to define instrumentation around the dispatch of data loaders.
+ *
+ * Requires [com.netflix.graphql.dgs.internal.DgsWrapWithContextDataLoaderScanningInterceptor] to be enabled,
+ * as DgsSimpleDataLoaderInstrumentation only provides hooks for the WithContext versions of batch loaders.
+ */
+interface DgsSimpleDataLoaderInstrumentation<C> {
```
kilink, line 27: Does the generic type actually add any value here? I think it would practically be cleaner to just return / accept a context parameter as `Any`, and then you wouldn't need the `afterLoadWithCast` method. It seems like it's always a star-projected type so it's always `Any?` in practice.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsDataLoaderScanningInterceptor.kt
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * While this hook appears very similar to [com.netflix.graphql.dgs.DataLoaderInstrumentationExtensionProvider]
+ * there are two important differences:
+ * - DgsDataLoaderScanningInterceptors are called when scanning for @DgsDataLoaders, whereas the
+ *   DataLoaderInstrumentationExtensionProviders are called once per request
+ * - DgsDataLoaderScanningInterceptors are afforded the opportunity to change the type of each DataLoader
+ *   as it is being initially registered. Most notably, this allows for converting BatchLoader and MappedBatchLoader
+ *   into their "WithContext" versions, which can simplify certain types of DataLoader instrumentation.
+ */
+interface DgsDataLoaderScanningInterceptor {
```
kilink, line 37: IMO, a better name might be something like `DataLoaderCustomizer` or `DgsDataLoaderCustomizer`. It is sort of the prevailing pattern already to prefix everything with Dgs, which is a bit unfortunate.

Emily: Ah good suggestion! And yeah, it's a bummer about the `Dgs` prefix, every time I add a new class I dream of cutting a major version where we just remove it from everything...

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsSimpleDataLoaderInstrumentation.kt
Emily: It allows someone using the interface to have type safety in their implementation. I was (and am) pretty torn on whether or not it's worthwhile though.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -166,20 +168,42 @@ open class DgsAutoConfiguration(
...
     }
 
+    @Bean
+    @ConditionalOnProperty(
+        prefix = "$AUTO_CONF_PREFIX.convertAllDataLoadersToWithContext",
+        name = ["enabled"],
+        havingValue = "true",
+        matchIfMissing = true
+    )
+    @Order(0)
```
srinivasankavitha, line 178: Consider using something like ` @Order(PriorityOrdered.HIGHEST_PRECEDENCE)`

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsDataLoaderScanningInterceptor.kt
srinivasankavitha: I'm actually wondering if this can just be moved to under internal.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
```diff
@@ -50,6 +50,7 @@ import kotlin.system.measureTimeMillis
 class DgsDataLoaderProvider(
     private val applicationContext: ApplicationContext,
     private val extensionProviders: List<DataLoaderInstrumentationExtensionProvider> = listOf(),
+    private val scanningInterceptors: List<DgsDataLoaderScanningInterceptor> = listOf(),
```
srinivasankavitha, line 53: Do we want to support a list of interceptors/customizers? I think it makes sense for instrumentations but not so convinced about opening up to multiple interceptors that can wrap. Especially since it's going to be used internally not something we expose to users?

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
paulbakker: Related to what @srinivasankavitha mentions, 0 is neither highest or lowest, because priority is INT.max/INT.min.

paulbakker:
I'm curious if you've tried to replace the bytebuddy stuff with this. At the very least it would be a good way to test the new API, see if it's possible or what you run into.

I generally like the approach with the two different interfaces that happen at different times in the lifecycle.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/DgsSimpleDataLoaderInstrumentation.kt
kilink: The `InstruementationState` example is analogous to what I am suggesting though. `Instrumentation` doesn't have an generic type, all of the methods take an `InstrumentationState` instance and the implementation needs to cast to the specific type.

I guess another approach is to change the flow a bit and have the returned context have the callback method for the "onComplete" event; it would be more analogous to the `Instrumentation` stuff in graphql-java already.

```
interface BatchLoaderInstrumentation {
    fun beginLoad(context: LoaderContext): BatchLoaderInstrumentationContext
}

interface BatchLoaderInstrumentationContext {
    fun onDispatched(result: CompletableFuture<*>)
    fun onCompleted(result: Any?, t: Throwable?)
}
```

Then we don't really have to deal with unchecked casts and the like.

kilink: Actually based on [this](https://github.com/graphql-java/graphql-java/pull/3457), the interface could be simplified.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSimpleDataLoaderInstrumentationDataLoaderScanningInterceptor.kt
Emily: Yeah, was torn on which way to do this. I think it's probably best to do what you've suggested and make sure instrumentation can't mess things up.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -166,20 +168,42 @@ open class DgsAutoConfiguration(
...
+    open fun dgsWrapWithContextDataLoaderScanningInterceptor(): DgsWrapWithContextDataLoaderScanningInterceptor {
+        return DgsWrapWithContextDataLoaderScanningInterceptor()
+    }
+
+    @Bean
+    @Order(100)
```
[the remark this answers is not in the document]
Emily: Mostly arbitrary. It should happen after the one that has Order(0), and the other one has Order(0) because while it should in most cases probably happen first, I don't want to unnecessarily restrict someone from putting something before it.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
Emily: see: https://github.com/Netflix/dgs-framework/pull/1818#discussion_r1499974931

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderProvider.kt
Emily: I think it is still useful to expose the ability to add customizers to end users. I see them as a useful building block (or escape hatch?) if our particular instrumentation implementation isn't quite what they want.

Emily:
@kilink @srinivasankavitha @paulbakker PTAL

@paulbakker I haven't yet, as the tests for the data loader metrics are currently disabled. Re-enabling them looks to be moderately involved so I was going to do that and the new style of instrumentation together as a follow up.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -166,20 +168,42 @@ open class DgsAutoConfiguration(
...
 
+    @Bean
+    @ConditionalOnProperty(
+        prefix = "$AUTO_CONF_PREFIX.convertAllDataLoadersToWithContext",
+        name = ["enabled"],
+        havingValue = "true",
+        matchIfMissing = true
+    )
+    @Order(0)
+    open fun dgsWrapWithContextDataLoaderScanningInterceptor(): DgsWrapWithContextDataLoaderCustomizer {
```
Emily, line 179: Missed renaming this.

srinivasankavitha (approved):
Looks good to me!

PR #1842, opened 2024-03-15
Rework DgsDataLoaderRegistry to avoid some overhead

kilink:
DgsDataLoaderRegistry extends DataLoaderRegistry, which exposes its underlying ConcurrentHashMap of dataloaders in a protected field. Switch to accessing this map directly to avoid extra copying incurred in methods such as getKeys.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
@@ -22,6 +22,7 @@ import org.dataloader.registries.DispatchPredicate
 import org.dataloader.registries.ScheduledDataLoaderRegistry
 import org.dataloader.stats.Statistics
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
 import java.util.function.Function
 
 /**
@@ -33,8 +34,7 @@ import java.util.function.Function
  * https://github.com/graphql-java/java-dataloader#scheduled-dispatching
  */
 open class DgsDataLoaderRegistry : DataLoaderRegistry() {
-    private val scheduledDataLoaderRegistries: MutableMap<String, DataLoaderRegistry> = ConcurrentHashMap()
-    private val dataLoaderRegistry = DataLoaderRegistry()
+    private val scheduledDataLoaderRegistries: ConcurrentMap<String, ScheduledDataLoaderRegistry> = ConcurrentHashMap()
 
     /**
      * This will register a new dataloader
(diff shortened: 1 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
```diff
@@ -165,34 +165,23 @@ open class DgsDataLoaderRegistry : DataLoaderRegistry() {
...
      */
     override fun dispatchAllWithCount(): Int {
-        var sum = 0
-        scheduledDataLoaderRegistries.forEach {
-            sum += it.value.dispatchAllWithCount()
-        }
-        sum += dataLoaderRegistry.dispatchAllWithCount()
-        return sum
+        return scheduledDataLoaderRegistries.values.sumOf { registry -> registry.dispatchAllWithCount() } +
+            dataLoaders.values.sumOf { dataLoader -> dataLoader.dispatchWithCounts().keysCount }
```
srinivasankavitha, line 169: Can't we use `super.dispatchAllWithCount` and add that to the sum?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
```diff
@@ -165,34 +165,23 @@ open class DgsDataLoaderRegistry : DataLoaderRegistry() {
...
     override fun dispatchDepth(): Int {
-        var totalDispatchDepth = 0
-        scheduledDataLoaderRegistries.forEach {
-            totalDispatchDepth += it.value.dispatchDepth()
-        }
-        totalDispatchDepth += dataLoaderRegistry.dispatchDepth()
-
-        return totalDispatchDepth
+        return scheduledDataLoaderRegistries.values.sumOf { registry -> registry.dispatchDepth() } +
+            dataLoaders.values.sumOf { dataLoader -> dataLoader.dispatchDepth() }
```
srinivasankavitha, line 178: Same as above, can we use `super.dispatchDepth` which will give us the same result?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsDataLoaderRegistry.kt
kilink: I think it would double count, since it calls `getDataLoaders`, which we've overridden to include the dataloaders from `scheduledDataLoaderRegistries`.

srinivasankavitha: Oh I see..right. Thanks for that.

srinivasankavitha (approved):

PR #1859, opened 2024-03-27
Fix DgsDataLoaderRegistryConsumer when wrapped with Micrometer

prokop7:
Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix

Changes in this PR
----

When DgsDataLoader is declared as DgsDataLoaderRegistryConsumer and micrometer wrapper is on. Then `batchLoader is DgsDataLoaderRegistryConsumer` condition willl not work beacuse MappedBatchLoaderWithContextInstrumentationDriver is not DgsDataLoaderRegistryConsumer

Alternatives considered:
----

I see two viable solutions, first - the one that is in this pull request. This solution isn't the most beautiful but it works
Second - make interface DataLoaderProxy with one method `fun getOriginal(): DataLoader`. With that we can always get the original dataloader within DgsDataLoaderProvider and setup Registry there

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
@@ -243,6 +243,82 @@ class MicrometerServletSmokeTest {
     }
 
     @Test
+    fun `Metrics for a successful request with data loaders with registry consumer`() {
+        mvc.perform(
+            MockMvcRequestBuilders
+                .post("/graphql")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(
+                    """{"query": 
+                    |   "{ registryVerifier }" }
+                    """.trimMargin()
+                )
+        ).andExpect(status().isOk)
+            .andExpect(
+                content().json(
+                    """
+                    |{
+                    |   "data":{
+                    |       "registryVerifier": "present"
+                    |   }
+                    |}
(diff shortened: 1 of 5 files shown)


graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
```diff
@@ -242,6 +242,32 @@ class MicrometerServletSmokeTest {
             )
     }
 
+    @Test
```
srinivasankavitha, line 245: I don't see any metrics being verified here? I understand the change on its own, but how is it related to Micrometer other than being used in this test?

prokop7: I could verify the metrics here, however there is a similar test with literally the same metrics above.
This is not directly related to Micrometer; it would be better to move such a test to the main module.
I put it here just to cut the corner of initialization of a wrapped DataLoader

prokop7: Yup, sure. I also found out that tests within DgsDataLoaderInstrumentationTest don't verify anything, because verifications made within `whenComplete` and run aside main thread
Would you like me to fix it?

srinivasankavitha (approved):
Thanks for the PR

srinivasankavitha (commented):
Would you be able to rebase?

prokop7:
Sure, done

PR #1874, opened 2024-04-06
Disable problematic autoconfigs from spring-graphql by default

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [ x] Feature
- [ ] Build related changes

Changes in this PR
----
* GraphQlObservationAutoConfiguration: Unneeded since DGS already takes care of metrics
* GraphQlWebMvcSecurityAutoConfiguration: This changes how security errors are handled, which would break the expectations from the errors spec.

This uses the `spring.autoconfigure.exclude` property to exclude the autoconfigs, while not overriding what the user might have excluded already.

The PR also introduces properties to explicitly enable the autoconfigs if wanted.


kilink (approved):

srinivasankavitha (commented):
Looks good to me!

PR #1875, opened 2024-04-06
Use micrometer context propagation for virtual thread executor

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes

Changes in this PR
----

Set up a Micrometer `ContextSnapshotFactory` and use this to wrap tasks in `VirtualThreadTaskExecutor`. This enables context propagation using the micrometer context-propagation library.
By default it registers the SLF4j MDC propagation that comes out of the box with micrometer context. A user can provide more accessors by either registering them in `META-INF/services`, which is the SPI Micrometer uses, or you can override the bean definition that's now in the DGS autoconfiguration:

```
@Bean
    @ConditionalOnMissingBean
    open fun dgsMicrometerContextRegistry(): ContextRegistry {
        return ContextRegistry.getInstance()
            .registerThreadLocalAccessor(Slf4jThreadLocalAccessor())
    }
```

WIth this PR, a datafetcher that runs on a virtual thread, has context propagated, so you can for example rely on MDC.

### DIFF
--- a/graphql-dgs-client/dependencies.lock
+++ b/graphql-dgs-client/dependencies.lock
@@ -848,6 +848,14 @@
                 "com.github.mifmif:generex"
             ]
         },
+        "io.micrometer:context-propagation": {
+            "locked": "1.1.1",
+            "transitive": [
+                "com.netflix.graphql.dgs:graphql-dgs",
+                "com.netflix.graphql.dgs:graphql-dgs-platform",
+                "com.netflix.graphql.dgs:graphql-dgs-spring-boot-oss-autoconfigure"
+            ]
+        },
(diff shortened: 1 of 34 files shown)


srinivasankavitha:
Should the micrometer dependency be pinned here? Should we add it to the BOM instead?

srinivasankavitha (approved):

paulbakker:
> Should the micrometer dependency be pinned here? Should we add it to the BOM instead?

Yes, good idea. Done. 
There's one place I couldn't get rid of the version because the multi-release `java21Implementation` syntax didn't work with a BOM version.

kilink (approved):

PR #1884, opened 2024-04-16
Set InputArgumentResolver to highest priority so it is matched first

srinivasankavitha:
This fixes issues with resolving MultipartFile with @InputArgument.

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Without the ordering on InputArgumentResolver, the HandlerMethodArgumentResolverAdapter for RequestParamMethodArgumentresolver was match due to this logic: https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/method/annotation/RequestParamMethodArgumentResolver.java#L142

This breaks the use of `@InputArgument MultipartFile inputFile` in DGS data fetchers. The InputArgumentResolver handles this correctly and we should match that resolver. It was working previously prior to spring-graphql intgeration since we were getting lucky with the order.

### DIFF
--- a/graphql-dgs-example-java/src/main/java/com/netflix/graphql/dgs/example/datafetcher/FileUploadMutation.java
+++ b/graphql-dgs-example-java/src/main/java/com/netflix/graphql/dgs/example/datafetcher/FileUploadMutation.java
@@ -66,6 +66,13 @@ public class FileUploadMutation {
         return !parts.isEmpty();
     }
 
+    @DgsData(parentType = "Mutation", field = "uploadTest")
+    public boolean uploadTest(@InputArgument MultipartFile input, DataFetchingEnvironment dfe) {
+        MultipartFile file = input;
+        if (file == null || file.isEmpty()) return false;
+        return true;
+    }
+
     static class FileUploadInput {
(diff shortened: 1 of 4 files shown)


paulbakker (approved):

kilink:
I don't think it was random before, it was just that `DgsInputArgumentConfiguration` was getting imported before [this configuration](https://github.com/Netflix/dgs-framework/blob/74542dfda020fc41606cbd9e440ec727c5a68958/graphql-dgs-spring-webmvc-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/webmvc/autoconfigure/DgsWebMvcAutoConfiguration.kt#L133). It looks like for the Spring GraphQL configuration we import this stuff before `DgsAutoConfiguration`, causing the breakage.

srinivasankavitha:
Yeah it wasn't random in that it was non-deterministic. It's more the fact that it happened to work because of the autoconfig order, but once we changed that we introduced new behavior which was unexpected because it's not explicit that the InputArgument resolver should always run first.

PR #1886, opened 2024-04-18
Add request size metric

kailyak:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Added micrometer DistributionSummmary metric for graphQL request size. 

```
Metric Name: gql.request.size
Tags: [
	- tag(gql.operation={operationTypeValue})
	- tag(gql.operation.name={operationNameValue})
	- tag(gql.query.complexity={qComplexityValue})
	- tag(gql.query.sig.hash={sigHashValue})
	]

```

This metric is useful in measuring the request size for a graphql operation.
The request size measurement is the size in bytes of the request body, obtained via the Content-Length header in the graphql request. This avoids needing to de/serialize data. A graphql request body contains: query, variables, and operationname.
Via this approach, the Content-Type header for the response is not available at this point in the request processing. This change is scoped to the request size metric, as it is a metric that is useful in correlating graphql request sizes to graphql operations.

Tested via smoke tests and manual testing on ExampleApp.

Side note: An alternative approach considered was adding a web filter which would allow for more efficient measurement of both the request and response sizes but it is not standardized for graphql clients to send operation-name as a header. Presently getting the operation-name in the filter would require deserializing the http request body to get the "operationName" field.

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
@@ -1,5 +1,7 @@
 package com.netflix.graphql.dgs.metrics.micrometer
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.kotlinModule
 import com.netflix.graphql.dgs.Internal
 import com.netflix.graphql.dgs.internal.DgsSchemaProvider
 import com.netflix.graphql.dgs.metrics.DgsMetrics.GqlMetric
@@ -29,16 +31,27 @@ import graphql.schema.DataFetcher
 import graphql.schema.GraphQLNamedType
 import graphql.schema.GraphQLTypeUtil
 import graphql.validation.ValidationError
+import io.micrometer.core.instrument.DistributionSummary
 import io.micrometer.core.instrument.MeterRegistry
 import io.micrometer.core.instrument.Tag
 import io.micrometer.core.instrument.Timer
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.boot.actuate.metrics.AutoTimer
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder
+import java.io.ByteArrayOutputStream
 import java.util.Optional
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 import kotlin.jvm.optionals.getOrNull
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
 
 class DgsGraphQLMetricsInstrumentation(
     private val schemaProvider: DgsSchemaProvider,
@@ -52,6 +65,8 @@ class DgsGraphQLMetricsInstrumentation(
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(DgsGraphQLMetricsInstrumentation::class.java)
+        private val applicationScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+        private val mapper: ObjectMapper = Jackson2ObjectMapperBuilder().modules(kotlinModule()).build()
     }
 
     @Deprecated("Deprecated in Java")
@@ -92,6 +107,14 @@ class DgsGraphQLMetricsInstrumentation(
         state: InstrumentationState
     ): CompletableFuture<ExecutionResult> {
         require(state is MetricsInstrumentationState)
+
+        try {
+            applicationScope.launch(CoroutineName("captureGqlQueryResponseSizeMetric-${state.operationNameValue}")) {
+                captureGqlQueryResponseSizeMetric(executionResult, parameters, state)
+            }
+        } catch (exc: Exception) {
+            log.warn("Exception thrown while attempting to serialize and measure response size of graphql data.")
+        }
 
         val errorTagValues = ErrorUtils.sanitizeErrorPaths(executionResult)
         if (errorTagValues.isNotEmpty()) {
@@ -204,6 +227,15 @@ class DgsGraphQLMetricsInstrumentation(
         if (properties.tags.complexity.enabled) {
             state.queryComplexityValue = ComplexityUtils.resolveComplexity(parameters)
         }
+
+        try {
+            applicationScope.launch(CoroutineName("captureGqlQueryRequestSizeMetric-${state.operationNameValue}")) {
+                captureGqlQueryRequestSizeMetric(parameters.executionContext.executionInput, state)
+            }
+        } catch (exc: Exception) {
+            log.warn("Exception thrown while attempting to serialize and measure request size of graphql data.")
+        }
+
         return super.beginExecuteOperation(parameters, state)
     }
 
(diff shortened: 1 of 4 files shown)


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -382,4 +446,13 @@ class DgsGraphQLMetricsInstrumentation(
 
         internal data class ErrorTagValues(val path: String, val type: String, val detail: String)
     }
+
+    internal object MeasurementUtils {
+        inline fun <reified T> serializeAndMeasureSize(obj: T): Int {
+            return ByteArrayOutputStream().use { outputStream ->
+                mapper.writeValue(outputStream, obj)
```
paulbakker, line 453: This serializes the whole thing _just_ for the metric. Since JSON serialization is quite expensive, I'm expecting this to impact performance.

paulbakker (changes requested):

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
srinivasankavitha: Perhaps capturing the size of the request and response body in a ServletFilter might save on the overhead, since it's serialized already?

paulbakker: I think a ServletFilter would be a more appropriate place. There's also some complexity because you can only read the body once in a filter. I'm not completely sure what the best way to get the size there is.

Either way, something like that needs to be done so that body is serialized/deserialized more than once. Extra serialization/deserialization won't be acceptable from a CPU perspective.

kailyak: Thank you for the feedback, I will investigate the ServletFilter approach

paulbakker: After digging for a little bit, Spring has a `ContentCachingRequestWrapper` and `ContentCachingResponseWrapper` that takes care of the "can only read/write once" issue.

A filter like this seems to work: 

```

@Component
public class SizeFilter implements Filter {
    @Override
    public void init(FilterConfig filterConfig) throws ServletException {
        Filter.super.init(filterConfig);
    }

    @Override
    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {

        if (request instanceof HttpServletRequest servletRequest) {
            var wrappedRequest = new ContentCachingRequestWrapper(servletRequest);
            var wrappedResponse = new ContentCachingResponseWrapper((HttpServletResponse) response);
            System.out.println("Request content length: " + wrappedRequest.getContentLength());
            chain.doFilter(wrappedRequest, wrappedResponse);

            System.out.println("Response content length: " + wrappedResponse.getContentSize());
            wrappedResponse.copyBodyToResponse();

        } else {
            chain.doFilter(request, response);
        }
    }

    @Override
    public void destroy() {
        Filter.super.destroy();
    }
}
```

There will still be some overhead because it needs to copy bytes around, but it will be way less expensive than JSON (de)serialization.

This sits at the request level, so it's not GraphQL specific at all. It probably doesn't need to be?

paulbakker: Another idea - just take the size of the query string as the size? It's not json at that point, but does that matter at all? Not sure how this metric is going to be used, but I expect it's to get an idea of the size of the query, not necessarily the exact JSON size

srinivasankavitha: Oh, if the query is available as a string, then that's all we need! assuming variables are also taken care of.

kailyak: Yeah thinking about it more, I don't think it's needed to serialize it, I thought serializing it would more accurately reflect the size but it might not be worth the tradeoff plus yeah, the metric is likely needed more to get an idea of the size, not the exact size and I'll double check with the user who requested the metric

For the variables, in the approach where we measure the size in the instrumentation code, the type of that and other components types: 
- variables (`Map<String, Object>`)
- result data (`Map<String, *>`)
- extensions (`Map<Object, Object>`)
- errors (`List<GraphQLError>`) 

So they are not available as strings like the input query is, I'll look into the object types as that might be okay, if not then I don't think this has to be graphQL specific necessarily and the approach with measuring at the request/response level would work better

kilink: For a request body, one can just check the `Content-Length` header. For the response, I think it should be opt-in regardless due to the overhead of performing extra serialization.

See the related micrometer issue: https://github.com/micrometer-metrics/micrometer/issues/880

srinivasankavitha: What if we just add the metric for the request using the content-length as part of instrumentation for now. The response body metric is proving to be costly so I doubt it's a useful metric to have anyway.

kailyak: @kilink , yes, to get operationName for the metric tag, it seems like the request body would need to be deserialized in the filter. It would be cool if operationName was a expected header for graphql clients to provide, then, with the filter approach, getting the response size would be more efficient. Thanks for the suggestion of content length!
@srinivasankavitha yes, for now I will add the request size as that is a valuable metric to have and with content-length we can get the size efficiently in the instrumentation code, thanks for the feedback!

graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
```diff
@@ -767,6 +767,110 @@ class MicrometerServletSmokeTest {
...
+        // Check expected percentiles: .90, .95, .99
+        assertThat(meters["gql.request.size.percentile"]).isNotNull
+
+        // Check metric name and expected tags.
+        assertThat(meters["gql.request.size"]).isNotNull
+        assertThat(meters["gql.request.size"]?.first()?.id?.tags)
+            .containsAll(
+                Tags.of("gql.operation", "QUERY")
+                    .and("gql.operation.name", "my_op_1")
+                    .and("gql.query.complexity", "5")
```
paulbakker, line 797: Do we need tags like query.complexity on this metric?

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMetricsInstrumentation.kt
```diff
@@ -225,6 +230,29 @@ class DgsGraphQLMetricsInstrumentation(
         )
     }
 
+    private fun recordGqlRequestSizeMetric(
+        executionContext: ExecutionContext,
+        state: MetricsInstrumentationState
+    ) {
+        // Capture size using Content-Length header.
+        val requestSizeBytes = DgsContext.from(executionContext.graphQLContext).requestData?.headers?.contentLength
```
paulbakker, line 238: This looks much better from a performance perspective :)

kailyak:
@paulbakker @kilink @srinivasankavitha I requested a re-review this AM, please disregard it.

kailyak:
Thank you for the feedback, an internal implementation of these metrics is being explored first, closed out this PR.

PR #1904, opened 2024-05-06
Replace ByteBuddy with Java Proxy

AyushChaubey:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Please find the details related to this PR in [this](https://github.com/Netflix/dgs-framework/discussions/1892) discussion

ByteBuddy seems to be causing issue when the native image is run. One of the solution to overcome this issue is to use Java Proxy.

**Note:**
I ran some performance tests from my side on a service. It would be nice if you also have some performance tests that covers this flow.

**Instructions on creating a native image:**
For a library to support GraalVM, it needs to provide the [reachability metadata](https://www.graalvm.org/latest/reference-manual/native-image/metadata/). GraalVM needs to know about the dynamically-accessed elements(Reflection/Resource loading etc.) at the time of build. This information is provided by the libraries in the form of `json` files. 

There is [this](https://github.com/oracle/graalvm-reachability-metadata/tree/master) repository, where community contributes this metadata information for different libraries. One example is [these](https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/com.graphql-java/graphql-java/19.2) metadata json files for `graphql-java`. You will find files like `reflect-config.json`, `resource-config.json` and others. 

There is a way to generate these metadata files which is using the [GraalVM tracing agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/). So what we do is, we run the tests in our service that uses `DGS` using gradle `test` task and provide the tracing agent parameter with the path where it should generate these metadata json files.

```
tasks.withType<Test> {
        jvmArgs("-agentlib:native-image-agent=config-output-dir=./src/main/resources/META-INF/native-image")
}
```

By doing this, those metadata json files are generated at this path `src/main/resources/META-INF/native-image`. Next when we build the native image, it uses that information and creates a native image that works perfectly. 
Luckily, other than `DGS micrometer`, we didn't face any issue with the native image in terms of the implementation in DGS.

We are using Spring Boot, which works with GraalVM to generate the metadata information but DGS would also need to provide this information to reduce the complete dependency on the tracing agent. I guess we would still need to use the tracing agent to cover all the cases which also means writing tests that cover all the scenarios.


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/DgsDataLoaderInstrumentationProvider.kt
```diff
@@ -2,94 +2,61 @@ package com.netflix.graphql.dgs.metrics.micrometer.dataloader
...
-                .getDeclaredConstructor()
-        }.newInstance()
-    }
+    private val batchLoaderClasses = mutableMapOf<String, BatchLoader<*, *>>()
+    private val batchLoaderWithContextClasses = mutableMapOf<String, BatchLoaderWithContext<*, *>>()
+    private val mappedBatchLoaderClasses = mutableMapOf<String, MappedBatchLoader<*, *>>()
+    private val mappedBatchLoaderWithContextClasses = mutableMapOf<String, MappedBatchLoaderWithContext<*, *>>()
+
+    override fun provide(original: BatchLoader<*, *>, name: String): BatchLoader<*, *> =
+        batchLoaderClasses.getOrPut(name) {
```
AyushChaubey, line 21: I am not creating a new instance at each access like it was done with ByteBuddy. Not sure if it will cause any issue. Would like to know your opinion on this.

paulbakker:
Sorry for the slow reply @AyushChaubey, I was traveling and completely missed the PR.

This looks great! I've tested it and it seems to work fine. I also pushed some more changes to your branch, because I realized that we don't need a separate handler now for each type of dataloader, since it's using reflection now.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/DgsDataLoaderInstrumentationProvider.kt
paulbakker: We were recently debating why we did that for Bytebuddy, and I couldn't remember the reason (I originally implemented it I think), so it might not have been necessary.
It should be ok like this I think.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderWithContextInterceptor.kt
```diff
@@ -3,34 +3,53 @@ package com.netflix.graphql.dgs.metrics.micrometer.dataloader
...
-import net.bytebuddy.implementation.bind.annotation.Pipe
-import org.dataloader.BatchLoaderWithContext
 import org.slf4j.LoggerFactory
+import java.lang.reflect.InvocationHandler
+import java.lang.reflect.Method
 import java.util.concurrent.CompletionStage
 
 internal class BatchLoaderWithContextInterceptor(
-    private val batchLoader: BatchLoaderWithContext<*, *>,
+    private val batchLoaderWithContext: Any,
```
paulbakker, line 14: Because it's now using reflection the specific interface doesn't actually matter anymore, and we can get away with a single InvocationHandler interceptor that handles all four types of batch loaders.

graphql-dgs-spring-boot-micrometer/build.gradle.kts
```diff
@@ -1,7 +1,6 @@
 dependencies {
     api(project(":graphql-dgs"))
 
-    implementation("net.bytebuddy:byte-buddy")
```
paulbakker, line 4: 🎉

paulbakker (approved):
@AyushChaubey  can you double check that you're ok with the changes I made to your branch?

AyushChaubey:
> Sorry for the slow reply @AyushChaubey, I was traveling and completely missed the PR.
> 
> This looks great! I've tested it and it seems to work fine. I also pushed some more changes to your branch, because I realized that we don't need a separate handler now for each type of dataloader, since it's using reflection now.

No problem @paulbakker. Thanks for looking into the PR. It makes sense, we don't need separate handler for different types of dataloaders. Thanks for pushing the changes.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/DgsDataLoaderInstrumentationProvider.kt
AyushChaubey: Great Thanks.

AyushChaubey:
> @AyushChaubey can you double check that you're ok with the changes I made to your branch?

Changes look good @paulbakker. Appreciate your help with this. Thanks 🙇‍♂️.

paulbakker:
@AyushChaubey, this is just a heads-up that we got a bit distracted by some urgent issues that needed attention. We were planning to do a bit more testing before releasing this since it's a pretty small but important change.

Will be merged and released soon, sorry for the delay.

kilink (approved):

AyushChaubey:
> @AyushChaubey, this is just a heads-up that we got a bit distracted by some urgent issues that needed attention. We were planning to do a bit more testing before releasing this since it's a pretty small but important change.
> 
> Will be merged and released soon, sorry for the delay.

No problem at all @paulbakker. Thanks for the update :). I see the PR is merged. Appreciate your time and effort for taking it through.

PR #1905, opened 2024-05-07
Convert graphql-java validation errors to TypedGraphQL error

srinivasankavitha:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
This makes the error consistent with our other error handling in the DGS framework where we emit TypedGraphQlErrors. Previously, we were emitting the validation error from graphql-java directly.


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ValidationErrorInstrumentation.kt
```diff
@@ -0,0 +1,34 @@
...
+import graphql.ExecutionResult
+import graphql.GraphQLError
+import graphql.execution.instrumentation.InstrumentationState
+import graphql.execution.instrumentation.SimplePerformantInstrumentation
+import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters
+import graphql.validation.ValidationError
+import java.util.concurrent.CompletableFuture
+
+class ValidationErrorInstrumentation : SimplePerformantInstrumentation() {
+    override fun instrumentExecutionResult(executionResult: ExecutionResult?, parameters: InstrumentationExecutionParameters?, state: InstrumentationState?): CompletableFuture<ExecutionResult> {
```
tinnou, line 16: nit: you might be able to get rid of the null checks as the parameters are likely non nullable.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ValidationErrorInstrumentation.kt
```diff
@@ -0,0 +1,34 @@
...
+        if (executionResult != null && executionResult.errors.isNotEmpty()) {
+            val newExecutionResult = ExecutionResult.newExecutionResult().from(executionResult)
+            val graphqlErrors: MutableList<GraphQLError> = mutableListOf()
+            executionResult.errors.filterIsInstance<ValidationError>().forEach { validationError ->
+                val graphqlError = TypedGraphQLError
+                    .newBuilder()
+                    .errorType(ErrorType.BAD_REQUEST)
+                    .errorDetail(ErrorDetail.Common.FIELD_NOT_FOUND)
```
tinnou, line 24: validationError can represent more than just `FIELD_NOT_FOUND` right?

tinnou: example I can think of is syntax errors

srinivasankavitha: Yeah it was what is most appropriate in the given enum. I thought about omitting the error detail to avoid confusion. Alternatively, we could set it if only we get fieldUndefined, but leave it empty for invalid argument, for e.g.?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/ValidationErrorInstrumentation.kt
```diff
@@ -0,0 +1,34 @@
...
+import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters
+import graphql.validation.ValidationError
+import java.util.concurrent.CompletableFuture
+
+class ValidationErrorInstrumentation : SimplePerformantInstrumentation() {
+    override fun instrumentExecutionResult(executionResult: ExecutionResult?, parameters: InstrumentationExecutionParameters?, state: InstrumentationState?): CompletableFuture<ExecutionResult> {
+        if (executionResult != null && executionResult.errors.isNotEmpty()) {
+            val newExecutionResult = ExecutionResult.newExecutionResult().from(executionResult)
+            val graphqlErrors: MutableList<GraphQLError> = mutableListOf()
+            executionResult.errors.filterIsInstance<ValidationError>().forEach { validationError ->
```
tinnou, line 20: are there other kind of errors we want to catch and transform into our `TypedGrahQLError` structure? For example, if this error [NonNullableFieldWasNullError](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/NonNullableFieldWasNullError.java#L44) is thrown will it be mapped correctly?

tinnou: it looks like on our v1 gateway side we used a [mapping](https://stash.corp.netflix.com/projects/ste/repos/graphlayer/browse/graphlayer-core/src/main/kotlin/com/netflix/graphlayer/core/TypedError.kt?at=c130399c3e4e599f79938752003787722f89e1f1#80) to transform errors thrown graphql-java style

srinivasankavitha: hmm..Ok, I might have to move this to the internal framework then. The framework doesn't really translate all  graphql-java errors to TypedGraphQLErrors as such.only exceptions are handled by default this way.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/GraphQLJavaErrorInstrumentation.kt
```diff
@@ -0,0 +1,72 @@
...
+                        } else if (error.validationErrorType == ValidationErrorType.FieldUndefined) {
+                            graphqlErrorBuilder.errorDetail(ErrorDetail.Common.FIELD_NOT_FOUND)
+                        }
+                    }
+
+                    graphqlErrors.add(graphqlErrorBuilder.build())
+                } else if (error.errorType == graphql.ErrorType.OperationNotSupported) {
+                    val graphqlErrorBuilder = TypedGraphQLError
+                        .newBuilder()
+                        .errorType(ErrorType.UNAVAILABLE)
```
tinnou, line 51: unavailable is like 503 right, maybe this is more of a bad request?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/GraphQLJavaErrorInstrumentation.kt
```diff
@@ -0,0 +1,72 @@
...
+                }
+
+                if (error.errorType == graphql.ErrorType.ValidationError || error.errorType == graphql.ErrorType.InvalidSyntax ||
+                    error.errorType == graphql.ErrorType.NullValueInNonNullableField
+                ) {
+                    val path = if (error is ValidationError) error.queryPath else error.path
+                    val graphqlErrorBuilder = TypedGraphQLError
+                        .newBadRequestBuilder()
+                        .locations(error.locations)
+                        .path(path)
```
tinnou, line 36: you are not setting the errorType here, shouldn't it be BAD_REQUEST?

srinivasankavitha: it's a badrequest builder, so it will be set there automatically.

tinnou (approved):

PR #1920, opened 2024-05-24
Upgrade to Spring Boot 3.3

paulbakker:
Pull Request type
----

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes

Changes in this PR
----
Upgrade to Spring Boot 3.3, which comes with graphql-java 22 and federation 5.0.0.

### DIFF
--- a/graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/ErrorsTest.kt
+++ b/graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/ErrorsTest.kt
@@ -17,6 +17,7 @@
 package com.netflix.graphql.dgs.client
 
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.springframework.http.HttpEntity
 import org.springframework.http.HttpHeaders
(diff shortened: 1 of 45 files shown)


iuliiasobolevska (approved):

Emily (approved):
looks like the main change ends up being the async preparsed document provider? I think this lgtm assuming the test is fixed

paulbakker:
For release notes:
* Mention this bug for Kotlin with `@JsonAnySetter` in Jackson 2.17: https://github.com/FasterXML/jackson-databind/issues/4508

graphql-dgs-client/src/test/kotlin/com/netflix/graphql/dgs/client/ErrorsTest.kt
```diff
@@ -178,6 +179,7 @@ class ErrorsTest {
     }
 
     @Test
+    @Disabled("Broken by Jackson 2.17 https://github.com/FasterXML/jackson-databind/issues/4508")
```
kilink, line 182: There's a workaround in that linked issue that we could apply instead:

https://github.com/FasterXML/jackson-databind/issues/4508#issuecomment-2107622837

paulbakker: @kilink From what I understand from the issue is that the workaround is for the Jackson folks to do, it's not something we can do as a user of Jackson.

Issue #1925, opened 2024-05-27
bug: Java Optional<> - Failed to convert from type LinkedHasMap<?, ?> to type


filippocozzini:
Hello everyone. I need to retrieve data from a graphql db and load it into a SetPointParametersInput object. When I try to convert it to Optional, it gives the following error:

`**Failed to convert from type [java.util.LinkedHashMap<?, ?>] to type [@com.netflix.graphql.dgs.InputArgument @jakarta.validation.Valid cloud.smartfish.writedown.inputs.SetPoint.SetPointInput] for value [{...}]**`

It works without the Optional conversion.
Bean of Jackson Object Mapper is setted. 
If i convert with ObjectMapper.convertValue function all works.

SetPointInput
```
@Getter
@Setter
@NoArgsConstructor
@AllArgsConstructor
@SuperBuilder(toBuilder = true)
public class SetPointParametersInput {

    private Optional<String> idFeedTable;

    private Optional<SetPointParameters.Rule> feeding;

    private Optional<SetPointParameters.Rule> conversion;

    private SetPointParameters.TimedTriggerRule loss;

    private Optional<SetPointParameters.TimedTriggerRule> selfGrow;

    private SetPointParameters.TimedTriggerRule starvation;

    private Optional<String> idBreedingStage;

    private Optional<String> idBreedingProcess;

    private Optional<String> idGradingStage;

    private Optional<String> idPathologicStage;

}
```

```
@Getter
    @Setter
    @NoArgsConstructor
    @AllArgsConstructor
    @SuperBuilder(toBuilder = true)
    public static class Rule implements Serializable {

        private String id;

        private RuleType ruleType;

    }

    @Getter
    @Setter
    @NoArgsConstructor
    @AllArgsConstructor
    @SuperBuilder(toBuilder = true)
    public static class TimedTriggerRule extends Rule {

        private Duration triggerAfter;

        private Duration frequency;

        public Duration getMaxPeriod() {
            return triggerAfter.getSeconds() >= frequency.getSeconds() ? triggerAfter : frequency;
        }

    }
    public Boolean isSetPointParametersEmpty() {
        return  this.getIdFeedTable() == null &&
                this.getFeeding() == null &&
                this.getConversion() == null &&
                this.getLoss() == null &&
                this.getSelfGrow() == null &&
                this.getStarvation() == null &&
                this.getIdBreedingStage() == null &&
                this.getIdBreedingProcess() == null &&
[...]

PR #1926, opened 2024-05-27
Fix handling of @InputArgument for Java objects with optionals.

srinivasankavitha:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
This PR fixes handling of @InputArgument for optionals. The current implementation of the DefaultInputObjectMapper does not handling converting from map to optionals, resulting in the following error when used: 
```
**Failed to convert from type [java.util.LinkedHashMap<?, ?>] to type [@com.netflix.graphql.dgs.InputArgument 
```
Conversion of String to Optional<String>, for e.g works fine since Spring's `ObjectToOptionalConverter` handles this fine: https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/support/ObjectToOptionalConverter.java

Adding similar logic to the `DefaultInputObjectMapper` to handle optionals based on this code: https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/support/ObjectToOptionalConverter.java#L76

Issue #1925

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
@@ -30,7 +30,9 @@ import org.springframework.core.convert.converter.GenericConverter
 import org.springframework.core.convert.support.DefaultConversionService
 import org.springframework.util.CollectionUtils
 import org.springframework.util.ReflectionUtils
+import java.lang.reflect.Array
 import java.lang.reflect.Type
+import java.util.*
 import kotlin.reflect.KClass
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.primaryConstructor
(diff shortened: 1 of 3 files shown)


Emily (changes requested):
Could you add a test that checks the behavior when one of the optional fields isn't present?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DefaultInputObjectMapper.kt
```diff
@@ -67,6 +69,18 @@ class DefaultInputObjectMapper(customInputObjectMapper: InputObjectMapper? = nul
             if (KotlinDetector.isKotlinType(targetType.type)) {
                 return mapper.mapToKotlinObject(sourceMap, targetType.type.kotlin)
             }
+
+            if (targetType.resolvableType.hasGenerics()) {
```
paulbakker, line 73: Should this check specifically for `Optional` instead of any type with generics?

filippocozzini:
Any update?

srinivasankavitha:
> Any update?

Sorry for the delay. I need to address some of the comments before I can push the fix through.

srinivasankavitha:
Declining this PR in favor of a better fix here: https://github.com/Netflix/dgs-framework/pull/1951

PR #1939, opened 2024-06-20
Fix deserialization of GraphQLErrorDebugInfo

kilink:
A recent Jackson upgrade broke deserialization of GraphQLErrorDebugInfo, and the recent fix broke the existing behavior of collecting unknown properties into the additionalInformation field. Add an explicit qualifier for the JsonAnySetter annotation so that it gets applied to the backing field and restores the old behavior, and re-enable the test case now that it passes.


paulbakker (approved):
Thanks for fixing this!

PR #1945, opened 2024-06-25
Pin to graphql-java 22.1, fix compilation errors

kilink:
Override the graphql-java version from the Spring BOM to pull in 22.1, and fix the compilation errors caused by nullability annotations.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,6 +59,10 @@ allprojects {
     // as dependencies. e.g. KOTLIN_VERSION
     extra["sb.version"] = "3.3.1"
     extra["kotlin.version"] = Versions.KOTLIN_VERSION
+    // Use graphql-java 22.1, which is the version used by spring-graphql. Once
+    // the Spring Boot BOM depends on this version, we can remove this override.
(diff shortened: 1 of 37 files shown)


build.gradle.kts
```diff
@@ -59,6 +59,10 @@ allprojects {
     // as dependencies. e.g. KOTLIN_VERSION
     extra["sb.version"] = "3.3.1"
     extra["kotlin.version"] = Versions.KOTLIN_VERSION
+    // Use graphql-java 22.1, which is the version used by spring-graphql. Once
+    // the Spring Boot BOM depends on this version, we can remove this override.
+    // See: https://github.com/spring-projects/spring-boot/issues/41219
+    extra["graphql-java.version"] = "22.1"
```
paulbakker, line 65: For future reference: I was surprised this works, since we don't reference this anywhere. However, it looks like the dependency recommender plugin from Spring Boot uses this to override versions.

paulbakker (approved):

srinivasankavitha (approved):

PR #1950, opened 2024-06-27
Register jdk8 module in DgsSSESubscriptionHandler

kailyak:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
- Add configuration for Jdk8 Module to jackson object mapper  
- Added implementation dependency for bringing in jackson's Jdk8Module

Internally the use case arose for needing to serialize java 8 Optionals in SSE subscription calls. Since the object mapper in DgsSSESubscriptionHandler is not configured for the Jdk8 module an error was thrown when that type was part of the payload [here](https://github.com/Netflix/dgs-framework/blob/57977c665c775048fc5028527e752037a8ffcba6/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt#L113).

Testing done
----

Tested on app which had the serialization error.

### DIFF
--- a/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
+++ b/graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
@@ -16,7 +16,7 @@
 
 package com.netflix.graphql.dgs.subscriptions.sse
 
-import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.graphql.dgs.DgsQueryExecutor
 import com.netflix.graphql.types.subscription.QueryPayload
 import com.netflix.graphql.types.subscription.SSEDataPayload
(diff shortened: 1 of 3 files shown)


srinivasankavitha (approved):

graphql-dgs-subscriptions-sse/src/main/kotlin/com/netflix/graphql/dgs/subscriptions/sse/DgsSSESubscriptionHandler.kt
```diff
@@ -139,7 +140,7 @@ open class DgsSSESubscriptionHandler(open val dgsQueryExecutor: DgsQueryExecutor
     }
 
     companion object {
-        private val mapper = jacksonObjectMapper()
+        private val mapper = jacksonObjectMapper().registerModule(Jdk8Module())
```
kilink, line 143: Maybe we should just use the `Jackson2ObjectMapperBuilder` from Spring, like we do elsewhere. It will automatically register these well-known modules if they're there on the classpath.

kilink: [Example](https://github.com/Netflix/dgs-framework/blob/872b0e76b8130d57766584d2ba210728695fbd69/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClients.kt#L29), except the usage here would be even simpler since the KotlinModule isn't being customized.

kailyak: Thank you! Yes, let's maintain consistency. Updated to use Jackson2ObjectMapperBuilder

PR #1954, opened 2024-07-02
Update to nebula.netflixoss 11.5.0

kilink:


srinivasankavitha (approved):

reneleonhardt:
@kilink Thank you for your contribution! Apply plugin fails with `Git user.name is not set.` in CI build, could you look into it?

PR #1955, opened 2024-07-08
Add trivial field to DgsData / DgsQuery annotations for marking trivial data fetcher methods

kilink:
Add ability to mark data fetcher methods as trivial
    
Add a "trivial" boolean field to DgsData and DgsQuery annotations which allows signaling to DGS
that the generated DataFetcher should implement TrivialDataFetcher. This can benefit data fetcher
methods that don't do any I/O, since they can be handled specially by instrumentation.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/MethodDataFetcherFactory.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/MethodDataFetcherFactory.kt
@@ -18,12 +18,17 @@ package com.netflix.graphql.dgs.internal.method
 
 import com.netflix.graphql.dgs.DgsData
 import com.netflix.graphql.dgs.internal.DataFetcherInvoker
+import graphql.TrivialDataFetcher
 import graphql.schema.DataFetcher
+import graphql.schema.DataFetchingEnvironment
+import graphql.schema.FieldCoordinates
 import org.springframework.core.DefaultParameterNameDiscoverer
 import org.springframework.core.MethodParameter
 import org.springframework.core.ParameterNameDiscoverer
+import org.springframework.core.annotation.MergedAnnotations
 import org.springframework.core.task.AsyncTaskExecutor
 import java.lang.reflect.Method
+import kotlin.jvm.optionals.getOrNull
 
 /**
  * Factory for constructing a [DataFetcher] given a [DgsData] annotated method.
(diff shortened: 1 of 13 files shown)


graphql-dgs/src/main/java/com/netflix/graphql/dgs/Trivial.java
```diff
@@ -0,0 +1,38 @@
...
+ * that the underlying {@link DataFetcher} should implement {@link TrivialDataFetcher}. Useful
+ * to connote simple data fetcher methods that don't do any I/O for instance; such methods are
+ * often handled specially by instrumentation.
+ *
+ * @see TrivialDataFetcher
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+public @interface Trivial { }
```
Emily, line 38: `DgsTrivial`?

I'm torn on this, as I never liked the prefix, but it's on _nearly all_ of the interfaces. And with the spring-graphql integration, I feel like the inconsistency might add some initial readability confusion.

Emily (approved):

graphql-dgs/src/main/java/com/netflix/graphql/dgs/Trivial.java
paulbakker, line 38: Have you considered making it a new argument for `@DgsData`? As long as it has a default value, that would be backward compatible as well I think?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/MethodDataFetcherFactory.kt
```diff
@@ -39,16 +43,31 @@ class MethodDataFetcherFactory(
     private val resolvers = ArgumentResolverComposite(argumentResolvers)
 
     fun createDataFetcher(bean: Any, method: Method): DataFetcher<Any?> {
+        if (isTrivial(method) || isTrivial(bean.javaClass)) {
```
paulbakker, line 46: Wondering if we should allow this on class level. It's probably unlikely that it's adding a lot of convenience, while it would be easy to have an unintended side-effect on datafetchers added later to an existing class.

graphql-dgs/src/main/java/com/netflix/graphql/dgs/Trivial.java
kilink: I recall considering that approach, but I couldn't think of a clean way to make `DgsQuery` work, although I guess I could add it to `DgsQuery` as well and annotate the field with `@AliasFor` to make that work. Having multiple `DgsData` annotations via `@DgsData.List` is also another thing I'd have to make sure works correctly. I can experiment with that and see if it works.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/MethodDataFetcherFactory.kt
kilink: I can just remove it for now then.

graphql-dgs/src/main/java/com/netflix/graphql/dgs/Trivial.java
kilink: Yeah I understand that concern, I ended up revisiting just adding a new field to the existing annotations based on Paul's feedback, and was able to make that work, so I'm scrapping this new annotation anyway.

kilink: @paulbakker okay, I've updated the PR to just use fields on the existing DgsData / DgsQuery annotations, and added test cases for the various scenarios.

paulbakker: Very nice! I think I prefer this approach, WDYT?

kilink: Yeah, I prefer this over adding a new annotation.

paulbakker (approved):

srinivasankavitha (approved):
Changes look good to me.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/MethodDataFetcherFactory.kt
```diff
@@ -38,17 +43,44 @@ class MethodDataFetcherFactory(
 
     private val resolvers = ArgumentResolverComposite(argumentResolvers)
 
-    fun createDataFetcher(bean: Any, method: Method): DataFetcher<Any?> {
```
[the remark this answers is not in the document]
kilink: I had written tests but forgot to check them in. I've updated the PR.

PR #1956, opened 2024-07-10
Enable toggling introspection via spring property

kailyak:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Context:  
- Disabling introspection via spring property (`spring.graphql.schema.introspection.enabled`) does not disable introspection. 
- Disabling introspection via dgs property works as expected and does disables introspection.  

This PR is focused on addressing the first point. 

In this PR:
- Adds disabling introspection for spring property  
- Adds mapping where, if dgs property is set, then the spring property gets set
- Refactors usages of graphql-java method for toggling introspection to use `Introspection.INTROSPECTION_DISABLED` instead.

Alternatives considered
----
One thing is whether to enforce the dgs property by throwing an exception if the spring property is set. This can be done [here](https://github.com/Netflix/dgs-framework/blob/master/graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLEnvironmentPostProcessor.kt#L49) but this code doesn't seem to be utilized presently so opted to not enforce by throwing `InvalidDgsConfigurationException` to allow for both properties to be able to toggled for introspection being disabled but I can update the PR to enforce it.

### DIFF
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -35,6 +35,7 @@ import com.netflix.graphql.dgs.internal.method.ArgumentResolver
 import com.netflix.graphql.dgs.internal.method.MethodDataFetcherFactory
 import com.netflix.graphql.dgs.scalars.UploadScalar
 import com.netflix.graphql.mocking.MockProvider
+import graphql.GraphQLContext
 import graphql.execution.AsyncExecutionStrategy
 import graphql.execution.AsyncSerialExecutionStrategy
 import graphql.execution.DataFetcherExceptionHandler
@@ -48,6 +49,8 @@ import graphql.schema.DataFetcherFactory
 import graphql.schema.GraphQLCodeRegistry
 import graphql.schema.GraphQLSchema
 import graphql.schema.idl.TypeDefinitionRegistry
+import graphql.schema.visibility.DefaultGraphqlFieldVisibility.DEFAULT_FIELD_VISIBILITY
+import graphql.schema.visibility.GraphqlFieldVisibility
 import io.micrometer.context.ContextRegistry
 import io.micrometer.context.ContextSnapshotFactory
 import io.micrometer.context.integration.Slf4jThreadLocalAccessor
@@ -279,8 +282,8 @@ open class DgsAutoConfiguration(
 
     @Bean
     @ConditionalOnMissingBean
-    open fun schema(dgsSchemaProvider: DgsSchemaProvider): GraphQLSchema {
-        return dgsSchemaProvider.schema(null).graphQLSchema
+    open fun schema(dgsSchemaProvider: DgsSchemaProvider, fieldVisibility: GraphqlFieldVisibility): GraphQLSchema {
+        return dgsSchemaProvider.schema(null, fieldVisibility).graphQLSchema
     }
 
     @Bean
@@ -305,8 +308,22 @@ open class DgsAutoConfiguration(
         havingValue = "false",
         matchIfMissing = false
     )
-    open fun noIntrospectionFieldVisibility() {
-        Introspection.enabledJvmWide(false)
+    open fun disableIntrospectionContextContributor(): GraphQLContextContributor {
+        return object : GraphQLContextContributor {
+            override fun contribute(
+                builder: GraphQLContext.Builder,
+                extensions: Map<String, Any>?,
+                requestData: DgsRequestData?
+            ) {
+                builder.put(Introspection.INTROSPECTION_DISABLED, true)
+            }
+        }
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    open fun defaultFieldVisibility(): GraphqlFieldVisibility {
+        return DEFAULT_FIELD_VISIBILITY
     }
 
     @Bean
--- a/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
@@ -43,13 +43,11 @@ import graphql.execution.ExecutionStrategy
 import graphql.execution.instrumentation.ChainedInstrumentation
 import graphql.execution.instrumentation.Instrumentation
 import graphql.execution.preparsed.PreparsedDocumentProvider
+import graphql.introspection.Introspection
 import graphql.schema.DataFetcherFactory
 import graphql.schema.GraphQLCodeRegistry
 import graphql.schema.GraphQLSchema
 import graphql.schema.idl.TypeDefinitionRegistry
-import graphql.schema.visibility.DefaultGraphqlFieldVisibility.DEFAULT_FIELD_VISIBILITY
-import graphql.schema.visibility.GraphqlFieldVisibility
-import graphql.schema.visibility.NoIntrospectionGraphqlFieldVisibility.NO_INTROSPECTION_FIELD_VISIBILITY
 import io.micrometer.context.ContextRegistry
 import io.micrometer.context.ContextSnapshotFactory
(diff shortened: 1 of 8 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -188,13 +184,13 @@ class DgsSchemaProvider(
...
         }
 
-        val codeRegistryBuilder = GraphQLCodeRegistry.newCodeRegistry().fieldVisibility(fieldVisibility)
+        val codeRegistryBuilder = GraphQLCodeRegistry.newCodeRegistry()
         if (defaultDataFetcherFactory.isPresent) {
             codeRegistryBuilder.defaultDataFetcher(defaultDataFetcherFactory.get())
         }
 
         val runtimeWiringBuilder =
-            RuntimeWiring.newRuntimeWiring().codeRegistry(codeRegistryBuilder).fieldVisibility(fieldVisibility)
```
Emily, line 197: I'm surprised this doesn't break disabling introspection for the non-spring-graphql version of the framework. Do we have a test for it?

kailyak: Yes, there is a test for it [here on the non-spring integration](https://github.com/Netflix/dgs-framework/blob/688a5db91bb0aabaf4f2c2c908b63acb767d6ee5/graphql-dgs-spring-boot-oss-autoconfigure/src/test/kotlin/com/netflix/graphql/dgs/autoconfig/DisabledIntrospectionTest.kt#L4) and I added one also on the [dgs-spring integration here](https://github.com/Netflix/dgs-framework/pull/1956/files#diff-11612275406ba1239d9ded889d1d58eae29ab7b27aeaf275108b5ecd36f660b0R241). 

What this part change does is refactor disabling introspection via the current deprecated way (configuring it on the executable schema via the RuntimeWiring) to the using the recommended way Instrospection.enabledJvmWide.

By default introspection is enabled so it doesn't need to be explicitly set, my reasoning was that the use case for setting this in the builder only applied when disabling introspection via the deprecated way but that's no longer the [best](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/visibility/NoIntrospectionGraphqlFieldVisibility.java#L16) way, this is why I removed it when refactoring to use Instrospection.enabledJvmWide

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -281,8 +279,8 @@ open class DgsAutoConfiguration(
 
     @Bean
     @ConditionalOnMissingBean
-    open fun schema(dgsSchemaProvider: DgsSchemaProvider, fieldVisibility: GraphqlFieldVisibility): GraphQLSchema {
```
kilink, line 284: Why do we not register a `GraphqlFieldVisibility` bean any longer? There may be use cases where people are currently registering a custom implementation, no?

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
```diff
@@ -307,14 +305,8 @@ open class DgsAutoConfiguration(
...
-    open fun noIntrospectionFieldVisibility(): GraphqlFieldVisibility {
-        return NO_INTROSPECTION_FIELD_VISIBILITY
-    }
-
-    @Bean
-    @ConditionalOnMissingBean
-    open fun defaultFieldVisibility(): GraphqlFieldVisibility {
-        return DEFAULT_FIELD_VISIBILITY
+    open fun noIntrospectionFieldVisibility() {
+        Introspection.enabledJvmWide(false)
```
kilink, line 309: I know that `NoIntrospectionGraphqlFieldVisibility` is deprecated, and suggests using this static / global setting. But affecting the state of the entire JVM as a side effect of a bean method seems a little off to me.

There's also the `Introspection.INTROSPECTION_DISABLED` String which can be put in the `GraphQlContext` to control this at a per-request level. Have we considered instead registering a `GraphQLContextContributor` that sets this context instead?

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
kilink: I mentioned in another comment, but `GraphqlFieldVisibility` itself is not deprecated. I think we're taking away functionality here, since someone could be registering a custom implementation with their own logic.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
kailyak: Re: affecting the state of the entire JVM as a side effect of a bean, that bean is for configuring introspection and, to my understanding, there one executable schema which queries are executed against in the dgs application. However, I do see that affecting the entire JVM could be too restrictive for more advanced usecases where there might be multiple executable schema running in the application or other cases (like the per request) where introspection is set dynamically. 

I have not looked into `Introspection.INTROSPECTION_DISABLED`, I'll take a look, thanks! In this case though the purpose of using Instrospection.disabledJvmWide was to be able to disable introspection via the spring property without adding in a deprecated method (for the case where users use that flag to disable introspection). I'm thinking the approach to take here might be the alternative approach considered, which is throwing an exception to prefer the dgs config as disabling at a per request level might be a bigger change/feature add for this case

kilink: Doing what I suggested would not use any deprecated classes / methods.

What I'm suggesting is doing something like this:

```kotlin
    @Bean
    @ConditionalOnProperty(
        prefix = "$AUTO_CONF_PREFIX.introspection",
        name = ["enabled"],
        havingValue = "false",
        matchIfMissing = false
    )
    open fun disableIntrospectionContextContributor(): GraphQLContextContributor {
        return GraphQLContextContributor { builder, _, _ ->
            builder.put(Introspection.INTROSPECTION_DISABLED, true)
        }
    }
```

paulbakker: I think this approach from @kilink is quite nice. The global JVM setting is a bit weird.

Emily: +1 to Patrick's approach.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
kailyak: Good point, updated to not remove this functionality.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
kailyak: Updated.

graphql-dgs-spring-boot-oss-autoconfigure/src/main/kotlin/com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.kt
kailyak: +1 very nice approach. Tried it out and updated it for both spring and dgs introspection enabled properties.

srinivasankavitha (approved):
The latest changes using the context contributor looks good to me.

PR #1957, opened 2024-07-10
GraphQLResponse.getRequestDetails should return nullable

Emily:
It can return null, so ought to have `?` so that folks using this from Kotlin don't end up with NPEs unexpectedly.


paulbakker:
While you're correct that it should be nullable, this is a painful change to make. It will effectively fail compilation for every Kotlin app using this API. Simple fix, but quite annoying, specially for a non-major release.

Emily:
@paulbakker true, though we've had at least one internal use get bitten by this. I think the minor annoyance of having to either add `!!` in a bunch of places, or actually properly null check is worth it.

paulbakker (approved):

PR #1972, opened 2024-08-01
Remove custom VirtualThreadTaskExecutor

kilink:
Switch to using the VirtualThreadTaskExecutor implementation from Spring, and existing conditional annotations from Spring Boot.


paulbakker (approved):
Yes, we can do this now. The previous code was there prior to the virtual thread support was released in Spring Boot, and that's no longer needed.

srinivasankavitha (approved):

PR #1973, opened 2024-08-01
Add ability to supply custom ObjectMapper to clients

kilink:
Add constructors / static factory methods to allow supplying a custom ObjectMapper to the GraphQL clients, which can be useful when the user needs precise control over serialization.

Additionally, remove deprecated DefaultGraphQLClient, fix various tests, and simplify serialization of the request body to avoid needing the KotlinModule installed at all.


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DefaultGraphQLClient.kt
paulbakker, line 1: This has been deprecated for a very long time, but I do see a bunch of internal usage (Netflix) still.
Not sure if it's worth removing it right now.

kilink: Okay, I brought back the deprecated client, but didn't implement this feature there.

paulbakker (approved):

srinivasankavitha (approved):

PR #1979, opened 2024-08-05
Support Kotlin value class parameters to data fetchers

### LINKED ISSUE #355: bug: Kotlin inline value classes don't work as @InputArgument
Hello guys,
New to graphql so may or may not be a configuration issue somewhere, or it might be a complicated multiparty problem.

## Expected behavior
To be able to use kotlin inline value classes as custom scalar inputs and pass them as method inputs with @InputArgument.

## Actual behavior
Various degrees of success, not able to use method definition as contract but `dfe.getArgument` works.

## Steps to reproduce

```kotlin
// Fetcher

@DgsComponent
class SomethingFetcher {
    // Errors due to value class jvm mangling
    // Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'schema' defined in class path resource [com/netflix/graphql/dgs/autoconfig/DgsAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [graphql.schema.GraphQLSchema]: Factory method 'schema' threw exception; nested exception is graphql.AssertException: Name must be non-null, non-empty and match [_A-Za-z][_0-9A-Za-z]* - was 'something-ViFSl3M'
    @DgsQuery
    fun something(@InputArgument("color") color: Color?): Int = 1
    
    /**
     * {
     *   "errors": [
     *     {
     *       "message": "com.netflix.graphql.dgs.exceptions.DgsInvalidInputArgumentException: Specified type 'class java.lang.String' is invalid. Found com.example.valuescalar.backend.graphql.model.Color instead.",
     *       "locations": [],
     *       "path": [
     *         "something"
     *       ],
     *       "extensions": {
     *         "errorType": "INTERNAL"
     *       }
     *     }
     *   ],
     *   "data": {
     *     "teams": null
     *   }
     * }
     */
    @DgsQuery
    @Suppress("INAPPLICABLE_JVM_NAME") // Needed as @Component from spring stops @JvmName from being usable, not sure why this is happening in the first place
    @JvmName("something")
    fun something(@InputArgument("color") color: Color): Int = 1
    
    // WORKS but less nice
    @DgsQuery
    fun something(dataFetchingEnvironment: DataFetchingEnvironment): Int {
        val color: Color = dataFetchingEnvironment.getArgument("color")
        return 1
    }
}
```

```kotlin
// Color
@JvmInline
value class Color(val value: String) {
    init {
        require(value.matches("^#[0-9A-F]{6}$".toRegex()))
    }

    override fun toString(): String = value
}
```

```kotlin
// Scalar coercing

@DgsScalar(name = "Color")
internal class ColorScalar : Coercing<Color, String> {
    @Throws(CoercingParseValueException::class)
    override fun parseValue(input: Any): Color =
        runCatching { Color(serialize(input)) }
            .getOrElse { throw CoercingParseValueException("Invalid color.", it) }

    @Throws(CoercingParseLiteralException::class)
    override fun parseLiteral(input: Any): Color {
        val colorString = (input as? StringValue)?.value ?: throw CoercingParseLiteralException("Invalid color.")
        return runCatching { Color(colorString) }
            .getOrElse { throw CoercingParseLiteralException("Invalid color.", it) }
    }

    override fun serialize(dataFetcherResult: Any): String = dataFetcherResult.toString()
}
```

```graphql
# Schema
type Query {
    something(color: Color): Int
}
scalar Color
```

```kotlin
// build.gradle.kts
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

plugins {
    id("org.springframework.boot") version "2.4.5"
    id("io.spring.dependency-management") version "1.0.11.RELEASE"
    kotlin("jvm") version "1.5.0"
    kotlin("plugin.spring") version "1.5.0"
    id("com.netflix.dgs.codegen") version "4.6.4"
}

group = "com.example.valuescalar"
version = "0.0.1-SNAPSHOT"
java.sourceCompatibility = JavaVersion.VERSION_11

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.jetbrains.kotlin:kotlin-reflect")
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")

    implementation("org.springframework.boot:spring-boot-starter-web")
    implementation("com.fasterxml.jackson.module:jackson-module-kotlin")

    implementation(platform("com.netflix.graphql.dgs:graphql-dgs-platform-dependencies:3.12.1"))
    implementation("com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter")

    testImplementation("org.springframework.boot:spring-boot-starter-test")
}

tasks {
    withType<com.netflix.graphql.dgs.codegen.gradle.GenerateJavaTask> {
        schemaPaths = mutableListOf("${projectDir}/src/main/resources/schema")
        packageName = "com.example.valuescalar.backend.graphql.generated"
        typeMapping = mutableMapOf(
            "Color" to "com.example.valuescalar.backend.graphql.model.Color"
        )
    }

    withType<KotlinCompile> {
        kotlinOptions {
            freeCompilerArgs = listOf("-Xjsr305=strict")
            jvmTarget = "11"
        }
    }

    withType<Test> {
        useJUnitPlatform()
    }
}
```

kilink:
Support Kotlin value class parameters as top-level arguments to data fetcher methods.

Fixes #355

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/AbstractInputArgumentResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/AbstractInputArgumentResolver.kt
@@ -26,6 +26,9 @@ import org.springframework.core.convert.TypeDescriptor
 import org.springframework.core.convert.support.DefaultConversionService
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
+import kotlin.reflect.KParameter
+import kotlin.reflect.jvm.jvmErasure
+import kotlin.reflect.jvm.kotlinFunction
 
 abstract class AbstractInputArgumentResolver(inputObjectMapper: InputObjectMapper) : ArgumentResolver {
 
@@ -43,6 +46,19 @@ abstract class AbstractInputArgumentResolver(inputObjectMapper: InputObjectMappe
     override fun resolveArgument(parameter: MethodParameter, dfe: DataFetchingEnvironment): Any? {
         val argumentName = getArgumentName(parameter)
         val value = dfe.getArgument<Any?>(argumentName)
+
+        val kfunc = parameter.method?.kotlinFunction
+        if (kfunc != null) {
+            val parameterIdx = if (kfunc.parameters.first().kind == KParameter.Kind.INSTANCE) {
+                parameter.parameterIndex + 1
+            } else {
+                parameter.parameterIndex
+            }
+            val param = kfunc.parameters[parameterIdx]
+            if (param.type.arguments.isEmpty() && param.type.jvmErasure.isInstance(value)) {
+                return value
+            }
+        }
 
         val typeDescriptor = TypeDescriptor(parameter)
         val convertedValue = convertValue(value, typeDescriptor)
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/InputArgumentTest.kt
@@ -20,6 +20,7 @@ import com.netflix.graphql.dgs.DgsComponent
 import com.netflix.graphql.dgs.DgsData
 import com.netflix.graphql.dgs.DgsQuery
 import com.netflix.graphql.dgs.DgsRuntimeWiring
+import com.netflix.graphql.dgs.DgsScalar
 import com.netflix.graphql.dgs.InputArgument
 import com.netflix.graphql.dgs.LocalDateTimeScalar
 import com.netflix.graphql.dgs.exceptions.DataFetcherInputArgumentSchemaMismatchException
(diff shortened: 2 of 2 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/AbstractInputArgumentResolver.kt
```diff
@@ -44,6 +47,19 @@ abstract class AbstractInputArgumentResolver(inputObjectMapper: InputObjectMappe
         val argumentName = getArgumentName(parameter)
         val value = dfe.getArgument<Any?>(argumentName)
 
+        val kfunc = parameter.method?.kotlinFunction
+        if (kfunc != null) {
+            val parameterIdx = if (kfunc.parameters.first().kind == KParameter.Kind.INSTANCE) {
```
srinivasankavitha, line 52: What does this check mean?

kilink: The first parameter to the `KFunction` is the instance parameter, e.g. the instance the method is being called on, so we need to adjust the parameter index based on that (essentially start the indexing after this instance parameter).

srinivasankavitha (approved):

PR #1981, opened 2024-08-09
Use SimpleAsyncTaskExecutor for virtual threads

kilink:
Switch to using SimpleAsyncTaskExecutor in conjunction with ContextPropagatingTaskDecorator when virtual threads is enabled.


kilink:
Just a minor clean up so we don't need to return an `AsyncTaskExecutor` implementation that handles the context propagation.

srinivasankavitha (approved):

PR #2002, opened 2024-09-09
added codeowners file

paulbakker:
Added codeowners file with core maintainers

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @paulbakker @kilink @srinivasankavitha @iuliiasobolevska @kzwang @asibross @iparadiso @jjacobs44


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @pbakker @kilink @srinivasankavitha @iuliiasobolevska @kzwang @asibross @iparadiso @jjacobs44
```
setchy, line 1:
```suggestion
* @paulbakker @kilink @srinivasankavitha @iuliiasobolevska @kzwang @asibross @iparadiso @jjacobs44
```

paulbakker: Thanks for the catch @setchy. You would think I at least get my own username right, but no 🥹

setchy: 🤣 all good mate

Issue #2041, opened 2024-10-17
bug: 2 separate functions annotated with DgsEntityFetcher for same entity works


stejas:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue.

## Expected behavior
should result in some kind of compiler or bootup error

## Actual behavior
no error, probably registers the function based on some custom order

paulbakker:
Fixed in #2053

PR #2053, opened 2024-11-05
Throw exception during startup when multiple entity fetchers for the same type are registered

paulbakker:
Pull Request type
----

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
Fix for issue #2041: Throw an exception during startup when multiple entity fetchers for the same type are registered.

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DuplicateEntityFetcherException.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DuplicateEntityFetcherException.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.exceptions
+
+import java.lang.reflect.Method
+
+class DuplicateEntityFetcherException(
+    val entityType: String,
+    val firstEntityFetcherClass: Class<out Any>,
+    val firstEntityFetcherMethod: Method,
+    val secondEntityFetcherClass: Class<out Any>,
+    val secondEntityFetcherMethod: Method,
+) : RuntimeException(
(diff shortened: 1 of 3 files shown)


srinivasankavitha (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DuplicateEntityFetcherException.kt
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.graphql.dgs.exceptions
+
+import java.lang.reflect.Method
+
+class DuplicateEntityFetcherException(
+    val entityType: String,
+    val firstEntityFetcherClass: Class<out Any>,
```
iuliiasobolevska, line 23: what if there are multiple (>=3) entity fetechers? it would be more efficient to have a set or smth and fail when there are more than one so that the consumer doesn't have to restart the app to find all violations

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DuplicateEntityFetcherTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+package com.netflix.graphql.dgs
+
+import com.netflix.graphql.dgs.DefaultDgsFederationResolverTest.Movie
+import com.netflix.graphql.dgs.exceptions.DuplicateEntityFetcherException
+import com.netflix.graphql.dgs.internal.DgsSchemaProvider
+import com.netflix.graphql.dgs.internal.method.MethodDataFetcherFactory
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
```
iuliiasobolevska, line 25: extra nit: why mix assertj and junit assertions in one test and not just to stick to either?

iuliiasobolevska (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DuplicateEntityFetcherException.kt
paulbakker: That would require extra temporary storage, because there will be only one value per key in the map. This is possible but adds complexity, which I don't think it's worth it because it's already quite an uncommon case.

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DuplicateEntityFetcherTest.kt
paulbakker: fixed, there was no reason for it.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/exceptions/DuplicateEntityFetcherException.kt
iuliiasobolevska: fair

Issue #2028, opened 2024-10-07
feature: Make DgsDataLoader usable on method


stengvac:
Hello,

could be possible to add on `DgsDataLoader` target `java.lang.annotation.ElementType.METHOD` ? To allow creation via `@Bean` functions. For `DgsComponent` such behavior is already possible.

paulbakker:
Yes, this probably makes sense for the `@Bean` use case.

PR #2055, opened 2024-11-05
support bean method-level @DgsDataLoader annotations

iparadiso:
Pull request checklist
----

- [ X] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
  first
- [ X] Make sure the PR doesn't introduce backward compatibility issues
- [ X] Make sure to have sufficient test cases

Pull Request type
----

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

Changes in this PR
----

This PR adds support for method-level annotations of `@DgsDataLoader` so their creation can be made via a bean definition of `@DgsComponent`.  The specific change needed for this is to correctly create the DataLoader holder from both the class and the bean method annotation metadata.  The other is in the DFE which retrieves the correct data loader.  Previous code assumed class/field anotations would always be present via reflection, so a special case was added for looking it up from the bean definition using the ApplicationContext, which must now be wired through to the DFE.

Issue # [2028](https://github.com/Netflix/dgs-framework/issues/2028)

Alternatives considered
----


paulbakker (approved):
Nice work @iparadiso!

PR #2086, opened 2024-12-13
Remove mocking module

paulbakker:
The mocking module was the first functionality of the framework even before the framework really existed. While it seemed like an interesting idea, it never really panned out and is mostly unused.

Since we're removing a lot of legacy in the upcoming release, I want to remove this module as well.


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt
```diff
@@ -71,7 +69,7 @@ import org.springframework.core.io.support.ResourcePatternUtils
 import org.springframework.util.ReflectionUtils
 import java.io.IOException
 import java.lang.reflect.Method
-import java.util.Optional
+import java.util.*
```
iuliiasobolevska, line 72: extra nit: revert back

iparadiso (approved):

srinivasankavitha (approved):

PR #2090, opened 2024-12-17
Add new `dgs-starter` module for an easier/shorter starter name

paulbakker:


srinivasankavitha (approved):

graphql-dgs-spring-graphql-starter/build.gradle.kts
```diff
@@ -16,9 +16,5 @@
 
 
 dependencies {
```
asibross, line 18: Consider adding a comment that this will be removed in the future and users should use the dgs-starter instead.

asibross (approved):

iparadiso (approved):

graphql-dgs-spring-graphql-starter/build.gradle.kts
paulbakker: done

Issue #2032, opened 2024-10-11
bug: Changed behaviour of spring.autoconfigure.exclude since com.netflix.graphql.dgs:graphql-dgs-spring-graphql-starter:8.5.4


edee111:
I wanted to upgrade DGS to the latest version `9.1.2`. So I changed the dependency in Gradle:
- from `com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter:8.5.8`
- to `com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter:9.1.2`

But then when starting an application there is a deprecation warning in the log
```
DEPRECATION WARNING - This project is using the deprecated 'graphql-dgs-spring-boot-starter'. Please switch to 'graphql-dgs-spring-graphql-starter'. For more context: https://netflix.github.io/dgs/spring-graphql-integration
```
But when changing the dependency to `com.netflix.graphql.dgs:graphql-dgs-spring-graphql-starter:9.1.2`, some unit tests stopped working. See the details bellow.

## Expected behavior
When overriding the Spring Boot (SB) property `spring.autoconfigure.exclude` e.g. in a unit test by the `@TestProperties` annotation, the value set in the annotation should be present in the application context.

## Actual behavior
Setting the `spring.autoconfigure.exclude` e.g. in unit test by the `@TestProperties` appends the value to the configuration property already defined e.g. in `application.yaml` file. Does not override it.

## Steps to reproduce
In [this repository](https://github.com/edee111/dgsautoconfigurationdemo/tree/main), the issue is reproduced.
Experiments in the repository show that:
- with `com.netflix.graphql.dgs:graphql-dgs-spring-boot-starter`
  - the autoconfiguration exclusion behaves as expected 
- with `com.netflix.graphql.dgs:graphql-dgs-spring-graphql-starter`
  - until DGS version 8.5.3 autoconfiguration exclusion behaves as expected
  - since DGS version 8.5.4 autoconfiguration exclusion behavior changed

## More info
[...]

PR #2102, opened 2025-01-07
Account for test property source in ExcludeAutoConfigurationsEnvironmentPostProcessor

paulbakker:
If a `spring.autoconfigure.exclude` property is defined in a test property source, don't undo the override in the `ExcludeAutoConfigurationsEnvironmentPostProcessor`

This fixes issue #2032 .


graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/ExcludeAutoConfigurationsEnvironmentPostProcessor.kt
```diff
@@ -59,8 +59,15 @@ class ExcludeAutoConfigurationsEnvironmentPostProcessor : EnvironmentPostProcess
...
 
-    private fun extractAllExcludes(propertySources: MutablePropertySources): String =
-        propertySources
+    private fun extractAllExcludes(propertySources: MutablePropertySources): String {
+        if (propertySources.any { it.name == INLINED_TEST_PROPERTIES }) {
+            val testExclude = propertySources.find { it.name == INLINED_TEST_PROPERTIES }?.getProperty(EXCLUDE)
+            if (testExclude is String && testExclude.isNotBlank()) {
+                return testExclude
+            }
+        }
```
iuliiasobolevska, line 68: to avoid double-iteration over `propertySources`:
```suggestion
    private fun extractAllExcludes(propertySources: MutablePropertySources): String {
        val testExclude = propertySources.find { it.name == INLINED_TEST_PROPERTIES }?.getProperty(EXCLUDE)
        if (testExclude != null && testExclude is String && testExclude.isNotBlank()) {
            return testExclude
        }
```

iparadiso (approved):

Issue #2101, opened 2025-01-03
After upgrading to DGS10.0.0 the response headers doesn't have custom headers set


dokkaraman:
### Discussed in https://github.com/Netflix/dgs-framework/discussions/2100

<div type='discussions-op-text'>

<sup>Originally posted by **dokkaraman** January  3, 2025</sup>
After upgrading to DGS10.0.0 the headers (Cache-Control) set via SimplePerformantInstrumentation no longer get set in the final response of a subgraph
This works till version 9.2.1</div>

paulbakker:
Can you provide some details about what you're doing in the `SimplePerformantInstrumentation` implementation?

dokkaraman:
We are adding Cache Control header as part of custom instrumentation which extends `SimplePerformantInstrumentation`
The value of cache control is variable which gets determined in this class.
This used to work till version 9.1.2 of DGS and after upgrade it started failing

paulbakker:
`SimplePerformantInstrumentation` doesn't have access to request/response without using some other context mechanism, so can you share exactly how you're doing this? That might explain why this changed hopefully.

dokkaraman:
If I change the above code to below it works

```
import graphql.GraphQLContext;
import lombok.extern.slf4j.Slf4j;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.graphql.server.WebGraphQlInterceptor;
import org.springframework.graphql.server.WebGraphQlRequest;
import org.springframework.graphql.server.WebGraphQlResponse;
import org.springframework.http.HttpHeaders;
import org.springframework.stereotype.Component;
import reactor.core.publisher.Mono;

import java.util.ArrayList;
import java.util.Collections;
import java.util.List;
import java.util.stream.Collectors;


@Component
@Slf4j
class ResponseHeaderInstrument implements WebGraphQlInterceptor {
    @Value("${entity.cacheMaxAge:max-age=60}")
    private String cacheMaxAge;


    @Override
    public Mono<WebGraphQlResponse> intercept(WebGraphQlRequest request, Chain chain) {
        return chain.next(request).doOnNext(response -> {
            response.getExecutionInput().getGraphQLContext();
            GraphQLContext
                gqlContext = (response != null && response.getExecutionInput() != null) ? response.getExecutionInput().getGraphQLContext() : null;
            HttpHeaders responseHeaders = getResponseHeadersFromContext(gqlContext, cacheMaxAge);
[...]

PR #2104, opened 2025-01-08
Dgs headers backward compatibility

paulbakker:
Fix for #2101. Prior to Spring GraphQL integration we could set response headers in two ways:

1. Set a field `dgs-response-headers` in the `extensions` of an `ExecutionResult`.
2. Return a `DgsExecutionResult` which has a `headers` field.

This functionality stopped working, because `DgsExecutionResult` (which was responsible for both mechanisms) was no longer referenced in the framework, so the special handling stopped working.

While the now recommended mechanism is to implement a `WebGraphQlInterceptor` to set headers as described in the [docs](https://netflix.github.io/dgs/advanced/intercepting-http-request-response/), existing code would have to be updated.

This PR adds support for the previous mechanisms. The implementation is now based on a `WebGraphQlInterceptor`.


graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
```diff
@@ -16,17 +16,7 @@
...
-import com.netflix.graphql.dgs.DgsDataLoaderCustomizer
-import com.netflix.graphql.dgs.DgsDataLoaderInstrumentation
-import com.netflix.graphql.dgs.DgsDataLoaderOptionsProvider
-import com.netflix.graphql.dgs.DgsDefaultPreparsedDocumentProvider
-import com.netflix.graphql.dgs.DgsFederationResolver
-import com.netflix.graphql.dgs.DgsQueryExecutor
-import com.netflix.graphql.dgs.DgsRuntimeWiring
-import com.netflix.graphql.dgs.DgsTypeDefinitionRegistry
-import com.netflix.graphql.dgs.ReloadSchemaIndicator
+import com.netflix.graphql.dgs.*
```
dokkaraman, line 19: Instead of using .* can you expand the imports like the deleted lines. This gives info on what explicit classes are pulled in the class.

graphql-dgs-spring-graphql/src/test/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsHeadersSmokeTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+@SpringBootTest(
+    classes = [
+        DgsHeadersSmokeTest.TestApp::class,
+        DgsSpringGraphQLAutoConfiguration::class,
+        GraphQlAutoConfiguration::class,
+        GraphQlWebMvcAutoConfiguration::class,
+        WebMvcAutoConfiguration::class,
+    ],
+    properties = [
```
dokkaraman, line 50: I believe formatting of this class is needed to fix the lint checks
https://github.com/Netflix/dgs-framework/actions/runs/12662158067/job/35286599471?pr=2104#step:10:6092

srinivasankavitha (approved):

iuliiasobolevska (approved):

iparadiso (approved):

dokkaraman:
May I know in which version this would be available?

PR #2107, opened 2025-01-13
update federation-jvm to 5.3.0

dariuszkuc:
`federation-jvm` v5 included a bump to a protobuf v4 version which is not ABI compatible with protobuf v3. Using protobuf v4 generates messages with a protobuf v3 service results in a runtime exception. A lot of the libraries in the `grpc` ecosystem are still stuck at v3 (including `grpc-java`) and since `federation-jvm` only uses protobuf to generate messages, the latest `federation-jvm` version ([5.3.0](https://github.com/apollographql/federation-jvm/releases/tag/v5.3.0)) also downgrades its dependency back to a v3 version (which generates messages that are compatible with both v3 and v4 versions of protobuf). 

see:
* https://github.com/apollographql/federation-jvm/issues/421
* https://github.com/grpc/grpc-java/issues/11015

### DIFF
--- a/dgs-starter-test/dependencies.lock
+++ b/dgs-starter-test/dependencies.lock
@@ -771,35 +771,35 @@
             ]
         },
         "io.mockk:mockk": {
-            "locked": "1.13.13"
+            "locked": "1.13.16"
         },
         "io.mockk:mockk-agent": {
-            "locked": "1.13.13",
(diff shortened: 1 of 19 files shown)


paulbakker (approved):
Thanks @dariuszkuc!

Issue #2122, opened 2025-02-19
Provide spring-configuration-metadata.json correctly


paulbakker:
Previously we had this file generated by Kapt, but it seems this stopped working when moving to K2.
Also, Kapt is deprecated and replaced by KSP, but this is not supported by Spring.
Also, the annotation processor always missed properties not defined in `@ConfigurationProperties`, such as properties in auto configuration.

Providing these files by hand is probably the best way forward for now.

paulbakker:
Fixed by #2123

PR #2123, opened 2025-02-19
Add spring-configuration-metadata.json files manually and remove kapt.

paulbakker:
Fix for #2122

Previously we had this file generated by Kapt, but it seems this stopped working when moving to K2.
Also, Kapt is deprecated and replaced by KSP, but this is not supported by Spring.
Also, the annotation processor always missed properties not defined in @ConfigurationProperties, such as properties in auto configuration.

Providing these files by hand is probably the best way forward for now.


graphql-dgs-spring-graphql/src/main/resources/META-INF/spring-configuration-metadata.json
```diff
@@ -0,0 +1,80 @@
+{
+  "properties": [
+    {
+      "name": "dgs.graphql.convert-all-data-loaders-to-with-context.enabled",
```
iparadiso, line 4: general nit:  We generally add "type" and "description" if we want to be extra friendly for type completion and IDE hints.

iparadiso (approved):

iuliiasobolevska (approved):

Issue #2126, opened 2025-02-19
feature: Enable DGS virtual threads when spring virtual threads are enabled


paulbakker:
Default virtual threads to enabled when VT are enabled in Spring with `spring.threads.virtual.enabled`.
It should still be possible to explicitly disable DGS virtual threads with `dgs.graphql.virtualthreads.enabled=false` however.

paulbakker:
Fixed by #2127

PR #2127, opened 2025-02-19
feature/gh-2126

paulbakker:
1. Allow enabling DGS virtual threads without Spring virtual threads enabled (different conditional on the autoconfig)
2. Automatically enable DGS virtual threads when Spring virtual threads are enabled (but allow to explicitly disable)


graphql-dgs-spring-graphql/src/test/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLEnvironmentPostProcessorTest.kt
```diff
@@ -104,4 +104,23 @@ class DgsSpringGraphQLEnvironmentPostProcessorTest {
...
+    fun `DGS virtual threads should by default be enabled when Spring virtual threads are enabled`() {
+        env.setProperty("spring.threads.virtual.enabled", "true")
+
+        DgsSpringGraphQLEnvironmentPostProcessor().postProcessEnvironment(env, application)
+
+        Assertions.assertThat(env.getProperty("dgs.graphql.virtualthreads.enabled")).isEqualTo("true")
+    }
+
+    @Test
+    fun `DGS virtual threads should by disabled when explicitly set to false`() {
```
iuliiasobolevska, line 118:
```suggestion
    fun `DGS virtual threads should be disabled when explicitly set to false`() {
```

iuliiasobolevska (approved):

PR #2128, opened 2025-02-21
User configured PreParsedDocumentProvider should be used for regular queries when APQ feature is enabled.

srinivasankavitha:
Pull Request type
----

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Changes in this PR
----
Support for Automated Persisted Queries are implemented in graphql-java using a PreParsedDocumentProvider : https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/preparsed/persisted/PersistedQuerySupport.java

The current setup allows us to configure only one PreParsedDocumentProvider. Therefore, if we enable APQ, we can only use a PreParsedDocumentProvider for APQ queries. Non-APQ or regular queries without a query hash will not be able to take advantage of a PreParsedDocumentProvider. Typically, users are able set up their own PreParsedDocumentProvider as a bean to speed up the request processing, especially critical for high scale applications.

This PR provides a wrapper to delegate to the appropriate PreParsedDocumentProvider based on whether the incoming query has an APQ hash or not. Also added some tests for the same.

Long term, we can work with spring or graphql-java to make this configurable.

Testing this feature:
- Enable  `dgs.graphql.apq.enabled` for the spring-graphql exampl app and start it
- Run the following curl commands to test APQ functionality:

```
curl -X POST http://localhost:8080/graphql --header 'content-type: application/json' -d '{"query":"{__typename}","extensions":{"persistedQuery":{"version":1, "sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}}'
```

```
curl -X POST http://localhost:8080/graphql --header 'content-type: application/json' -d '{"extensions":{"persistedQuery":{"version":1, "sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}}'
```

Example of regular query:
```
curl -X POST http://localhost:8080/graphql --header 'content-type: application/json' -d '{"query":"{__typename}"}'
```

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
@@ -14,7 +14,6 @@ import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegi
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration
 import org.springframework.boot.actuate.autoconfigure.metrics.PropertiesAutoTimer
 import org.springframework.boot.autoconfigure.AutoConfiguration
-import org.springframework.boot.autoconfigure.AutoConfigureAfter
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
@@ -22,6 +21,8 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cache.CacheManager
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.core.Ordered
+import org.springframework.core.annotation.Order
 import java.util.Optional
 
 /**
@@ -39,6 +40,7 @@ open class DgsGraphQLMicrometerAutoConfiguration {
     }
 
     @Bean
+    @Order(Ordered.LOWEST_PRECEDENCE)
     @ConditionalOnProperty(
         prefix = "$AUTO_CONF_PREFIX.instrumentation",
         name = ["enabled"],
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
@@ -14,6 +14,7 @@ import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegi
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration
 import org.springframework.boot.actuate.autoconfigure.metrics.PropertiesAutoTimer
 import org.springframework.boot.autoconfigure.AutoConfiguration
+import org.springframework.boot.autoconfigure.AutoConfigureAfter
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
@@ -21,8 +22,6 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cache.CacheManager
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
-import org.springframework.core.PriorityOrdered
-import org.springframework.core.annotation.Order
 import java.util.Optional
 
 /**
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
@@ -40,7 +40,6 @@ open class DgsGraphQLMicrometerAutoConfiguration {
     }
 
     @Bean
-    @Order(PriorityOrdered.LOWEST_PRECEDENCE)
     @ConditionalOnProperty(
         prefix = "$AUTO_CONF_PREFIX.instrumentation",
         name = ["enabled"],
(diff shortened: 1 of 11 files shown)


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
```diff
@@ -40,7 +40,6 @@ open class DgsGraphQLMicrometerAutoConfiguration {
     }
 
     @Bean
-    @Order(PriorityOrdered.LOWEST_PRECEDENCE)
```
paulbakker: What does this fix? The `@Order` is just bean creation order, so I want to make sure what the intent is.

graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
```diff
@@ -387,7 +387,7 @@ class MicrometerServletSmokeTest {
                     .and("gql.operation.name", "anonymous")
                     .and("gql.query.complexity", "none")
                     .and("gql.query.sig.hash", "none")
-                    .and("gql.errorDetail", "none")
+                    .and("gql.errorDetail", "INVALID_ARGUMENT")
```
paulbakker, line 390: Why did this change?

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQPreParsedDocumentProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+import graphql.ExecutionInput
+import graphql.execution.preparsed.PreparsedDocumentEntry
+import graphql.execution.preparsed.PreparsedDocumentProvider
+import graphql.execution.preparsed.persisted.ApolloPersistedQuerySupport
+import graphql.execution.preparsed.persisted.PersistedQueryCache
+import java.util.*
+import java.util.concurrent.CompletableFuture
+import java.util.function.Function
+
+class DgsAPQPreParsedDocumentProvider(
```
paulbakker, line 28: Could be a more descriptive name, e.g. it should communicate that this is really a wrapper

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
```diff
@@ -491,12 +491,14 @@ open class DgsSpringGraphQLAutoConfiguration(
...
         dataFetcherExceptionHandler: DataFetcherExceptionHandler,
+        environment: Environment,
     ): GraphQlSourceBuilderCustomizer =
         GraphQlSourceBuilderCustomizer { builder ->
             builder.configureGraphQl { graphQlBuilder ->
-                if (preparsedDocumentProvider.isPresent) {
-                    graphQlBuilder
-                        .preparsedDocumentProvider(preparsedDocumentProvider.get())
+                val apqEnabled = environment.getProperty("dgs.graphql.apq.enabled", Boolean::class.java, false)
+                // if apq is enabled use the APQPreparsedDocumentProvider instead
```
paulbakker, line 499: Comment and if statement don't seem to match? The if statement checks for apq NOT being enabled?

graphql-dgs-spring-graphql/src/test/kotlin/com/netflix/graphql/dgs/springgraphql/apq/DgsAPQPreParsedDocumentProviderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+                .query(PERSISTED_QUERY_MARKER)
+                .extensions(extensions)
+                .build()
+
+        every {
+            persistedQueryCache.getPersistedQueryDocumentAsync(any(), any(), any())
+        }.returns(CompletableFuture.completedFuture(computeFunction(executionInput)))
+
+        dgsAPQPreParsedDocumentProvider.getDocumentAsync(executionInput, computeFunction)
+        assertThat(count == 1)
```
paulbakker, line 77: `assertThat(count).isEqualTo(1)` for better error reporting

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
srinivasankavitha: I added this is a previous commit, which I now removed because it doesn't do anything. The micrometer test started failing because the order in which the instrumentations run seems to have changed in my branch. If the DgsGraphQlJavaErrorInstrumentation runs before the MetricsInstrumentation, the tag value is set properly (which is in this branch). However, if it's the other way around, the metrics add the tag for `gql.error` before the DgsGraphQlJavaErrorInstrumentation runs and that's why it passes in master branch.

Basically need a way to ensure the MetricsInstrumentation runs last, but not sure what the best way to do that is, so looking into it.

graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
srinivasankavitha: explained in response to earlier comment

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
srinivasankavitha: Will clean up, it's from an older commit.

graphql-dgs-spring-boot-micrometer/src/test/kotlin/com/netflix/graphql/dgs/metrics/micrometer/MicrometerServletSmokeTest.kt
srinivasankavitha: The micrometer test started failing because the order in which the instrumentations run seems to have changed in my branch. If the `DgsGraphQlJavaErrorInstrumentation` runs before the `MetricsInstrumentation`, the tag value is set properly (which is in this branch). However, if it's the other way around, the metrics add the tag for gql.error before the `DgsGraphQlJavaErrorInstrumentation` runs and that's why it passes in master branch.
The correct order should be `DgsGraphQlJavaErrorInstrumentation` first and then `DgsMetricsInstrumentation` so the metrics get the correct tag value as fixed in this PR.

Basically need a way to ensure the MetricsInstrumentation runs last, but not sure what the best way to do that is, so looking into it.

graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/DgsGraphQLMicrometerAutoConfiguration.kt
srinivasankavitha: Fixed by specifying order on both instrumentation classes. That ensures the correct relative ordering.

paulbakker (approved):

Issue #1618, opened 2023-08-30
bug: @DgsEntityFetcher annotation does not prevent compilation when method arguments are incorrect


ghost:
## Expected behavior

WHEN an entity fetcher is defined as such:
```java
@DgsEntityFetcher(name = DgsConstants.STORE.TYPE_NAME)
public Store storeFetcher(final DgsDataFetchingEnvironment dfe, final Map<String, Object> values) {
    return Store.newBuilder().name("myStore").build();
}
```
THEN compilation should fail because the parameters are in the wrong order

## Actual behavior

Compilation is successful and the service starts. When making this query:

```graphql
query {
  _entities(representations: [{__typename: "Store", id: "123"}]) {
    ... on Store {
      name
    }
  }
}
```
the returned error states `Exception while fetching data (/_entities) : argument type mismatch`. This is a misleading error that appears to come from GraphQL, but actually comes from Java.

## Steps to reproduce

```graphql
# schema
type Store @key(fields: "id") {
  id: ID!
  name: String
}
```

1. Create an entityfetcher with the signature described above.
2. Attempt to invoke it with the query described above.

kailyak:
Hi @aogilvie-indeed  thanks for the well documented issue. What version of dgs-framework are you using? On latest (version 7.5.1) with no custom exception handler I get this error message:
```
{
  "errors": [
    {
      "message": "java.lang.IllegalArgumentException: argument type mismatch",
      "locations": [],
      "path": [
        "_entities"
      ],
      "extensions": {
        "errorType": "INTERNAL"
      }
    }
  ],
  "data": {
    "_entities": [
      null
    ]
  }
}
```

You may be using a custom exception handler. The message on the latest version 7.5.1 with the default handler is not misleading and includes the java lang exception. Added a ticket internally for the compilation behavior.

aogilvie-indeed:
Hi @kailyak !

I'm using 5.4.1. However, my error message is identical to yours.

My suggestion is that we catch this function signature problem at compile time and not at run time. Why build something that cannot be executed?

Also, it's not clear that "argument type mismatch" is a Java problem with the entityfetcher and not a GraphQL problem with the type of one of the arguments sent in the query.

bcole:
I just want to +1 this issue, as I also faced the same error output and couldn't figure out why.  Thankfully I found this post to resolve my issue and swap argument order, but this issue is very valid and should be addressed in two was as @aogilvie-indeed is mentioning:

[...]

PR #2143, opened 2025-04-02
Allow DgsDataFetchingEnvironment as first argument for DgsEntityFetcher

kzwang:
Allow DgsDataFetchingEnvironment and Map<String, Object> values to be passed in any order for `@DgsEntityFetcher`

fixes https://github.com/Netflix/dgs-framework/issues/1618


paulbakker (approved):
We discussed offline to additonally also check for the correct argument during startup of the application. That can be done in a separate PR tho.

paulbakker (approved):

Issue #1192, opened 2022-08-18
feature: support JPMS


xenoterracide:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue. Also consider discussing your idea on
the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first.

## Describe the Feature Request

I'd like to be able to import dgs into an app using java modules. I searched the repo for a `module-info.java` or an `Automatic-Module-Name` Given the absence of the former it makes sense to do the latter.

## Describe Preferred Solution

You can add an `Automatic-Module-Name` to the manifest https://stackoverflow.com/a/46742802/206466 this is not backwards incompatible. A `module-info` would be nice, but it will  force people to java 9+

## Describe Alternatives

Without any support applications have to remove there's You simply can't import a module that doesn't have one of these if you're using it.

paulbakker:
I agree, let's start with adding a module name, so that at least it becomes usable. 

Once we baseline on Spring Boot 3.0, which requires Java 17, it makes sense to add module-info as well. There isn't a concrete plan for that yet, but will likely be early 2023 (since Spring Boot 3 will be release end of this year).

xenoterracide:
btw, any chance what's the potential itinerary on `Automatic-Module-Name`?

xenoterracide:
2 year bump... ?

paulbakker:
Implemented (finally 😅) in https://github.com/Netflix/dgs-framework/releases/tag/v10.1.0

PR #2144, opened 2025-04-03
Add support for JPMS

paulbakker:
Fix for: #1192 

This PR adds support for the Java Platform Module System (JPMS) by adding `module-info.java` to key modules. 
Unfortunately, Kotlin doesn't work well with JPMS, and it doesn't check any module enforcement during compilation, which means this was more of a trial-and-error process than it ideally should been.

I've tested the modules as follows:
1. With a newly created Maven application (Maven works nicer with modules) 
2. A Maven application where I "explode" the generated application jar file and run it manually from the command line with the `java` command and a `--module-path`.
3. Exploded as above but with Gradle

An application can use modules by creating a `module-info.java` in it's sources.

```java
module demo {
    requires com.netflix.graphql.dgs.springgraphql;
    exports dgs.modularity.demo to spring.beans, com.netflix.graphql.dgs.core;
}
```

An application should have a module require on `com.netflix.graphql.dgs.springgraphql` which transitively requires `com.netflix.graphql.dgs.core` which contains most exported types.
Note that you need to export your packages annotated with DGS annotations to both `spring.beans` and `com.netflix.graphql.dgs.core` for reflection access.

If you put all JAR dependencies and the application module  in a `lib` directory you can run an application as follows:

```
java --add-modules ALL-MODULE-PATH --module-path lib -m demo/dgs.modularity.demo.DemoApplication
```

The `--add-modules ALL-MODULE-PATH` is important because many of the transitive dependencies are not modules, and are used as automatic modules. You could instead list these modules explicitly in `requires` in your `module-info.java`, or use a mixed module/classpath.


graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLConfigurationProperties.kt
```diff
@@ -16,18 +16,13 @@
 
 package com.netflix.graphql.dgs.springgraphql.autoconfig
 
-import com.netflix.graphql.dgs.springgraphql.autoconfig.DgsSpringGraphQLConfigurationProperties.Companion.PREFIX
 import org.springframework.boot.context.properties.ConfigurationProperties
 import org.springframework.boot.context.properties.bind.DefaultValue
 
-@ConfigurationProperties(prefix = PREFIX)
+@ConfigurationProperties(prefix = "dgs.graphql.spring")
```
srinivasankavitha, line 22: Why do these need to change?

graphql-dgs-client/build.gradle.kts
```diff
@@ -40,3 +40,7 @@ dependencies {
     testImplementation(project(":graphql-dgs-spring-graphql-starter"))
     testImplementation(project(":graphql-dgs-spring-graphql-starter-test"))
 }
+
```
srinivasankavitha, line 43: Why is this needed for only some but not all modules? For e.g. why don;t we need this for `graphql-dgs-pagination` or `graphql-dgs-spring-boot-micrometer`

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/ReloadSchemaIndicator.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Netflix, Inc.
+ * Copyright 2025 Netflix, Inc.
```
iuliiasobolevska, line 2: why change it here? there are 50 other places with 2024 and hundreds with <=2023, I thought it's based on file introduction year

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLConfigurationProperties.kt
paulbakker: These top level constants don't work when compiling with modules. I think this is a Kotlin compiler bug, but you get an error that the package where the constant is defined isn't exported to the module it's being referenced, however, it's the same module.

build.gradle.kts
```diff
@@ -98,6 +98,9 @@ configure(subprojects.filterNot { it in internalBomModules }) {
         testImplementation("org.springframework.boot:spring-boot-starter-test") {
             exclude(group = "org.junit.vintage", module = "junit-vintage-engine")
         }
+
+
+        implementation("org.jetbrains:annotations:26.0.2")
```
iuliiasobolevska, line 103: nit: `graphql-dgs-client/build.gradle.kts` uses `org.jetbrains:annotations:26.0.1`, not a big deal but would be better to have all aligned (ideally some BOM should just recommend it for the repo)

graphql-dgs-client/build.gradle.kts
paulbakker: I added module-info to micrometer and pagination as well now. I tried extended validation as well, but there's a split package between `graphql-java` and `graphql-java-extended-validation`, so that's not working.

For Jars that are not real modules you can load them as automatic-modules on the module path, or just load them from the classpath.

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/ReloadSchemaIndicator.kt
paulbakker: oh actually that makes sense. I relocated this file to the core module, I think that's where it belongs. But that probably made it a "new" file

srinivasankavitha (approved):

Issue #1985, opened 2024-08-13
feature: Global Type Resolver


esfomeado:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue. Also consider discussing your idea on
the [discussion forum](https://github.com/Netflix/dgs-framework/discussions) first.

## Describe the Feature Request

As of now we have to add a `@DGSTypeResolver` for each type that we want to resolve.

If we have many types and they all behave the same way it becomes cumbersome to add a new one everytime we add a new interface/union.

## Describe Preferred Solution

Allow somehow to costumize the default type resolver that uses the class name or provide one that is used as fallback.

mrvaruntandon:
@esfomeado could you please share example code block that would explain the use of `@DGSTypeResolver` becoming cumbersome?

esfomeado:
@mrvaruntandon Creating a single `@DGSTypeResolver` is not cumbersome.

The issue is that you have to create one per interface that you have. So if you have thousands of interfaces you have to create thousands of type resolvers.

esfomeado:
Now that DGS is integrated with Spring GraphQL is possible to create a `GraphQlSourceBuilderCustomizer` and set a default type resolver like for instance a `ClassNameTypeResolver`.

The issue is that DGS on `DgsGraphQLSourceBuilder` doesn´t do anything with this type resolver. Check for instance the `DefaultSchemaResourceGraphQLSourceBuilder`, there they set the default type resolver on the `runtimeWiring` as expected.

The same should be possible on DGS. This way I could set a default type resolver and get rid of the `@DGSTypeResolver`

jjacobs44:
Closing as this should be addressed by https://github.com/Netflix/dgs-framework/pull/2145. Feel free to reopen or file a new issue if not.

PR #2145, opened 2025-04-04
Support fallback/global type resolver

jjacobs44:
Potential solution to https://github.com/Netflix/dgs-framework/issues/1985

Supports supplying a bean of type `TypeResolver` to act as the fallback type resolver when an explicit `@DgsTypeResolver` is not already registered for a given InterfaceType.

### DIFF
--- a/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
+++ b/graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
@@ -99,7 +99,7 @@ internal class DgsSchemaProviderTest {
             componentFilter = componentFilter,
             schemaWiringValidationEnabled = schemaWiringValidationEnabled,
             dataFetcherResultProcessors = dataFetcherResultProcessors,
-            fallbackTypeResolver = fallbackTypeResolver
+            fallbackTypeResolver = fallbackTypeResolver,
         )
 
     @DgsComponent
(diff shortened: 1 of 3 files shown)


graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/DgsSchemaProviderTest.kt
```diff
@@ -427,7 +430,90 @@ internal class DgsSchemaProviderTest {
...
+
+        class MyTypeResolverConfig {
+            fun myTypeResolver(): TypeResolver = TypeResolver { env -> env.schema.getObjectType("FakeType") }
+        }
+
+        @DgsComponent
+        class FetcherWithDefaultResolver {
+            @DgsTypeResolver(name = "Video")
+            @DgsDefaultTypeResolver
+            fun resolveType(
```
paulbakker, line 501: The test would maybe be better if these are two separate methods, which would be the realistic use case I think.

jjacobs44: Actually let me just remove the `@DgsDefaultTypeResolver` annotation from this test because I think it's confusing. That's not the fallback that I'm adding support for with this PR, that's an existing annotation that supports a default resolver for a **single** given type. It's functionality is best described by the `addOverrideTypeResolvers` test.

What I'm actually testing is whether or not `myTypeResolver` kicks in

paulbakker (approved):
Nice work!

PR #2147, opened 2025-04-08
Add @Source annotation to get easy access to the Source object in a @DgsData fetcher.

paulbakker:
Previously you would have to do: 

```java
 @DgsData(parentType = "Show")
public String artworkUrl(DgsDataFetchingEnvironment dfe) {
    Show show = dfe.getSourceOrThrow();
    // Do something with Show
}
```

```java
@DgsData(parentType = "Show")
    public String artworkUrl(@Source Show show) {
    //Do something with Show
}
```

### DIFF
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
@@ -28,7 +28,11 @@ class SourceArgumentResolver : ArgumentResolver {
         dfe: DataFetchingEnvironment,
     ): Any {
         val source = dfe.getSource<Any>()
-        if (source != null && parameter.parameterType == source.javaClass) {
+        if (source == null) {
+            throw IllegalArgumentException("Source is null. Are you trying to use @Source on a root field (e.g. @DgsQuery)?")
+        }
+
+        if (parameter.parameterType == source.javaClass) {
             return source
         } else {
             throw IllegalArgumentException(
--- a/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
+++ b/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2025 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun resolveArgument(
+        parameter: MethodParameter,
+        dfe: DataFetchingEnvironment,
+    ): Any {
+        val source = dfe.getSource<Any>()
+        if (source != null && parameter.parameterType == source.javaClass) {
+            return source
+        } else {
+            throw IllegalArgumentException(
+                "Invalid source type '${source?.javaClass?.name}'. Expected type '${parameter.parameterType.name}'",
```
srinivasankavitha, line 35: If this is null, what would the message be? Will it indicate that source is null?

srinivasankavitha: Can source even be null? I'm guessing graphql-java will throw an exception before it gets to this?

graphql-dgs/src/test/kotlin/com/netflix/graphql/dgs/internal/SourceArgumentTest.kt
```diff
@@ -0,0 +1,134 @@
...
+                        FallbackEnvironmentArgumentResolver(DefaultInputObjectMapper()),
+                    ),
+                ),
+        )
+
+    data class Show(
+        val title: String,
+    )
+
+    @Test
```
srinivasankavitha, line 59: Depending on whether @Source can be null, consider adding a test case for that scenario.

iuliiasobolevska (approved):

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/method/SourceArgumentResolver.kt
paulbakker: I don't think it can be null because the datafetcher wouldn't even be called

iuliiasobolevska: based on the interface definition:
```
@PublicApi
public interface DataFetchingEnvironment extends IntrospectionDataFetchingEnvironment {
        @Nullable
        <T> T getSource();
...
}
```
I guess it can

paulbakker: Actually it could be null when using `@Source` when there is no source, e.g. on a `@DgsQuery`. Added better handling and test.

srinivasankavitha (approved):

paulbakker:
No it was added in 10.0.0 and won't be backported. Time to upgrade :-)

PR #2151, opened 2025-04-16
Support scalar deserialize in GraphQLResponse

iparadiso:
Pull request checklist
----

  first

Pull Request type
----

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

Changes in this PR
----

This change supports scalar serialization and deserialize in GraphQLRequest/Response to support scalar options as variable parameters. We wrap each scalar in a Jackson serializer/deserializer to be used in the existing ObjectMapper.

This change also aligns all copyright comments, as well as corrects wildcard import formatting and linting to auto-fix and auto-add unambiguous imports to make development easier in the IDE and in the kotlin formatter.  

Issue #1702


paulbakker (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomMonoGraphQLClient.kt
```diff
@@ -29,8 +29,16 @@ class CustomMonoGraphQLClient(
...
     private val mapper: ObjectMapper,
+    private val options: GraphQLRequestOptions? = null,
 ) : MonoGraphQLClient {
-    constructor(url: String, monoRequestExecutor: MonoRequestExecutor) : this (url, monoRequestExecutor, GraphQLClients.objectMapper)
+    constructor(url: String, monoRequestExecutor: MonoRequestExecutor) : this (url, monoRequestExecutor, null)
+
+    constructor(url: String, monoRequestExecutor: MonoRequestExecutor, options: GraphQLRequestOptions? = null) : this(
+        url,
+        monoRequestExecutor,
+        GraphQLRequestOptions.createCustomObjectMapper(options),
```
srinivasankavitha, line 39: So does this mean the mapper that the user passes in during creation won't be used?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLRequestOptions.kt
```diff
@@ -0,0 +1,111 @@
...
+            serializers: SerializerProvider,
+        ) {
+            val serializedValue = coercing.serialize(value as Any, graphQLContext, Locale.getDefault())
+            gen.writeString(serializedValue.toString())
+        }
+    }
+
+    companion object {
+        fun createCustomObjectMapper(options: GraphQLRequestOptions? = null): ObjectMapper {
+            val mapper = ObjectMapper()
```
srinivasankavitha, line 85: Should we use the mapper that is passed in during construction of the GraphQlClient? See earlier comment.

srinivasankavitha (commented):
Changes looks good to me. A few comments.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomMonoGraphQLClient.kt
iparadiso: If the user uses the default constructor, they can supply their own ObjectMapper.

srinivasankavitha: that won't have the custom module you install though right?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLRequestOptions.kt
iparadiso: If they pass in their own mapper, then this code shouldn't be called.  We'll trust they know what they're doing.  :)

srinivasankavitha (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomMonoGraphQLClient.kt
iparadiso: That's right.  We could pass in an ObjectMapper that they provide here and extend it, but we risk undefined behavior or clobbering their configuration.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLRequestOptions.kt
regulskimichal: Hi, this code is actually used, even though someone wants to use a custom ObjectMapper. A custom mapper is never passed to GraphQLResponse—it is always created by this method.

PR #2167, opened 2025-05-27
Prefer caffeine cache with micrometer metrics in APQ autoconfig.

srinivasankavitha:
Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----
There are 2 configuration for the caffeine cache used for APQ - one with caffeine cache metrics set up in micrometer registry and one to default to if micromter's Meter registry is not available. We want to prefer the one with caffeine cache metrics when using APQ. This PR fixes the behavior where we end up using the fallback option without metrics due to the way the  conditionals were set up. With the change,  we will now always use the caffeine cache with micrometer metrics set up as long as the Meter Registry is available in the class path.

### DIFF
--- a/graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQSupportAutoConfiguration.kt
+++ b/graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQSupportAutoConfiguration.kt
@@ -31,6 +31,7 @@ import org.springframework.boot.autoconfigure.AutoConfigureAfter
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.autoconfigure.graphql.GraphQlSourceBuilderCustomizer
 import org.springframework.boot.context.properties.EnableConfigurationProperties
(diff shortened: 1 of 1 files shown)


graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQSupportAutoConfiguration.kt
```diff
@@ -93,7 +93,6 @@ open class DgsAPQSupportAutoConfiguration {
     )
     open class APQMicrometerMeteredCaffeineCacheConfiguration {
         @Bean
-        @ConditionalOnBean(io.micrometer.core.instrument.MeterRegistry::class)
```
paulbakker, line 96: This conditional should still be there, no?

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQSupportAutoConfiguration.kt
```diff
@@ -105,8 +109,13 @@ open class DgsAPQSupportAutoConfiguration {
         }
     }
 
+    // We want this version only if there is no micrometer meter registry
     @Configuration
-    @ConditionalOnMissingBean(APQMicrometerMeteredCaffeineCacheConfiguration::class)
```
paulbakker, line 109: Still want this one as well I think, so it could be user provided

srinivasankavitha: It's there, it's just a list now https://github.com/Netflix/dgs-framework/pull/2167/files#diff-81c22179fe1aa8884cdd1f6e458ea9be32c11bff73c59367c6d02bb256af8b47R115

paulbakker (approved):

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/apq/DgsAPQSupportAutoConfiguration.kt
srinivasankavitha: Fixed.

PR #2169, opened 2025-06-04
Plumb through user-defined ObjectMapper to be used in both request/response

iparadiso:
A previous change to merge in scalar types into serialization and deserialization missed an edge case where user-provided object mappers to the GQL request, and is clobbered in the GQL response by our generated mapper.  

This change passes the mapper used in the request to the response for Client wrappers.  The Mapper used in the request is either user-defined (which we pass through), or it is created and merged with any provided Scalar implementations, which was the feature added in [2151](https://github.com/Netflix/dgs-framework/pull/2151)

Pull request checklist
----

- [ X] Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md)
  first
- [X ] Make sure to have sufficient test cases

Pull Request type
----

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

This is a breaking change to 12.2.0 internal client constructors, but since it was only released yesterday there shouldn't be any usage of it.  I can make it non-breaking but it leaves an unused parameter, so I feel it's better to rip off the bandaid and have a cleaner API.

### DIFF
--- a/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
+++ b/graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
@@ -28,20 +28,16 @@ class CustomGraphQLClient(
     private val url: String,
     private val requestExecutor: RequestExecutor,
     private val mapper: ObjectMapper,
-    private val options: GraphQLRequestOptions? = null,
 ) : GraphQLClient {
     constructor(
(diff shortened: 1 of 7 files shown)


regulskimichal:
Hi @iparadiso. I can see that the solution to the problem I solved 3 weeks ago has already been merged. It's a little bit disappointing that my [pull request](https://github.com/Netflix/dgs-framework/pull/2159) has not been reviewed.

PR #2186, opened 2025-06-26
Correctly rethrow target exception in BatchLoaderWithContextInterceptor when dataloader fails

paulbakker:
Pull Request type
----

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

The `BatchLoaderWithContextInterceptor` tracks how much time is spent in a dataloader. The error handling of this interceptor would also catch exceptions happening in the dataloader, which is incorrect and leads to the exception being wrapped in an `InvocationTargetException`.
With this PR it will correctly rethrow the `targetException`,

### DIFF
--- a/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderWithContextInterceptor.kt
+++ b/graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderWithContextInterceptor.kt
@@ -7,6 +7,7 @@ import io.micrometer.core.instrument.Tag
 import io.micrometer.core.instrument.Timer
 import org.slf4j.LoggerFactory
 import java.lang.reflect.InvocationHandler
+import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
 import java.util.concurrent.CompletionStage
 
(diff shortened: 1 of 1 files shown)


graphql-dgs-spring-boot-micrometer/src/main/kotlin/com/netflix/graphql/dgs/metrics/micrometer/dataloader/BatchLoaderWithContextInterceptor.kt
```diff
@@ -51,9 +52,15 @@ internal class BatchLoaderWithContextInterceptor(
...
+                //If the dataloader itself fails just propagate the exception
+                if(exception is InvocationTargetException) {
+                    throw exception.targetException
+                } else {
+                    logger.warn("Error creating timer interceptor '{}' for {} with exception {}", ID, javaClass.simpleName, exception.message)
+                    @Suppress("UNCHECKED_CAST")
+                    method.invoke(batchLoaderWithContext, *(args)) as CompletionStage<List<*>>
+                }
+
             }
```
iuliiasobolevska, line 64: imho catching each exception type is cleaner:
```
            } catch (exception: InvocationTargetException) {
                //If the dataloader itself fails just propagate the exception
                throw exception.targetException
            } catch (exception: Exception) {
                logger.warn("Error creating timer interceptor '{}' for {} with exception {}", ID, javaClass.simpleName, exception.message)
                @Suppress("UNCHECKED_CAST")
                method.invoke(batchLoaderWithContext, *(args)) as CompletionStage<List<*>>
            }
```

paulbakker: Of course, I don't know why I didn't do that to begin with 🤦‍♂️

PR #2200, opened 2025-08-05
Feature: Add data loader reload API for development environments

berngp:
Pull Request type
----
- [X] Feature

Changes in this PR
----

## DGS Data Loader Reload Support

Implements programmatic mechanism to force reloading of _Data Loaders_, to be used during development without application restart. The implementation provides no overhead when _DGS Reloading_ is disabled. When enabled, it wrapps the current implementation, no called `DefaultDgsDataLoaderProvider` in such a way that a new one is created when the reload is invoked via `DgsDataLoaderReloadController`. It supports either reloading by the current application context or a new one provided through the _force reload_ api.

Features:
- DgsDataLoaderReloadController public API with statistics tracking
- ReloadableDgsDataLoaderProvider wrapping DefaultDgsDataLoaderProvider
- @ConditionalOnDgsReload annotation for development-only activation
- Zero overhead when disabled (default for non-laptop profiles)
- Symmetric functionality with existing schema reload while maintaining safety and not affecting performance characteristics.

### Spring Cloud RefreshScope

As an alternative we could leverage Spring Cloud's `@RefreshScope` mechanism versus implementing a custom reload solution. This will require a re-architecture of the current supported for the schema, broader changes to the DGS Spring Boot integration, and a direct dependency on Spring Cloud.

#### Spring Cloud RefreshScope Overview

**Core Characteristics:**
- Creates **lazy proxies** that initialize when methods are called
- Acts as a **cache of initialized values** that can be invalidated
- Enables **dynamic reconfiguration** via configuration property changes
- Supports **re-binding of `@ConfigurationProperties` beans**
- Triggered via `/refresh` endpoint or external configuration changes

**Lifecycle Management:**
- Beans are destroyed and recreated when refresh occurs
- Proxy wrapper remains but target bean instance changes
- Works primarily with configuration-driven changes

#### Architectural Decision

| Aspect                 | Spring Cloud RefreshScope    | DGS Custom Implementation   | Decision                              |
|------------------------|------------------------------|-----------------------------|---------------------------------------|
| **Scope**              | Any Spring bean              | Data loaders only           | **Custom** - Precise control needed |
| **Trigger Mechanism**  | Configuration changes        | Programmatic control        | **Custom** - Development-focused    |
| **Performance**        | Proxy overhead always        | Zero overhead when disabled | **Custom** - Better performance     |
| **Thread Safety**      | Framework-managed complexity | Explicit synchronization    | **Custom** - Predictable behavior   |
| **Use Case Alignment** | External config management   | Development hot-reload      | **Custom** - Purpose-built          |
| **Dependencies**       | Spring Cloud ecosystem       | Core DGS only               | **Custom** - Minimal dependencies   |

## Leveraging `DgsDataLoaderReloadController` with Spring Cloud

The current implementation provides an easy way to leverage Spring Cloud's reload mechanism. This could be done as follows...

```java
import org.springframework.cloud.context.scope.refresh.RefreshScopeRefreshedEvent;
import org.springframework.context.ApplicationListener;
import org.springframework.stereotype.Component;

@Component
public class DgsDataLoaderRefreshLifecycleListener implements ApplicationListener<RefreshScopeRefreshedEvent> {

    private DgsDataLoaderReloadController controller;

    @Override
    public void onApplicationEvent(RefreshScopeRefreshedEvent event) {
        controller.reloadDataLoaders(event.getApplicationContext());
    }
}
```

or

```java
import org.springframework.cloud.context.scope.refresh.RefreshScopeRefreshedEvent;
import org.springframework.context.event.EventListener;
import org.springframework.stereotype.Component;

@Component
public class RefreshEventHandler {

    private DgsDataLoaderReloadController controller;

    @EventListener
    public void handleRefresh(RefreshScopeRefreshedEvent event) {
        controller.reloadDataLoaders(event.getApplicationContext());
    }
}
```


paulbakker:
I'm in favor of the current implementation in the PR, I would like to stay away from Spring Cloud for this.

berngp:
Thanks Paul for the review, will address the feedback.

PR #2299, opened 2026-04-15
Jackson3 support with opt-in jackson2 support

jjacobs44:
## Summary                                                                                                                                                                                                    
                                         
  Following the precedent that OSS Spring Boot 4 set, these changes make Jackson 3 the default that's autoconfigured for the framework to use and Jackson 2 remains available as an opt-in via the `graphql-dgs-jackson2` module.
 
 ### Core changes                                                                                                                                                                                              
  - `DgsJsonMapper` interface abstracts Jackson version from DGS internals
  - `Jackson3DgsJsonMapper`: default implementation using Jackson 3                                                                                                                                             
  - `graphql-dgs-jackson2`: new module with `Jackson2DgsJsonMapper` and `DgsJackson2AutoConfiguration`
  - `BaseDgsQueryExecutor.objectMapper`/`.parseContext`: kept as deprecated lazy vals for backward compatibility (`HIDDEN` from new code, helpful error if Jackson 2 absent at runtime). I could also just remove them, it would be breaking but I don't think there's many usages                          
  - json-path upgraded from 2.x to 3.x for Jackson 3 provider support
                                                                                                                                                                                                                
  ### Client changes                  
  - `DgsGraphQLResponse` interface extracted from `GraphQLResponse`                                                                                                                                          
  - New jackson agnostic client classes: `DgsCustomGraphQLClient`, `DgsRestClientGraphQLClient`, `DgsWebClientGraphQLClient`, `DgsCustomMonoGraphQLClient`, etc..
  - Existing classes that tie their impl to Jackson 2 are deprecated                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
   
  ### Example projects                                                                                                                                                                                                                                                                                                                                                     
 `graphql-dgs-example-jackson3-only` | OSS default works, Jackson 2 clients fail to load 
  `graphql-dgs-example-jackson2-only` | Jackson 2 opt-in works, Jackson 3 clients fail to load                                                                                         
  `graphql-dgs-example-jackson-both` | Mixed classpath works, Jackson 2 wins autoconfig unless configuration specifies otherwise, both client sets usable                                                                                
                                                                                                                                                                                                                
  ---                                                                                                                                                                                                           
                                                                                                                                                                                                                
  ## Why New Classes/Interfaces for the Client Module?   

In the client module we have multiple cases where ObjectMapper is exposed in the public api (constructor). I evaluated a couple approaches for how to support jackson3 in this case. After discussing internally option 3 was picked for the best overall migration story for libraries 
                                                                                                                                                                                                                
  ### Approach 1: Constructor Overloads                     
  Add `JsonMapper` (Jackson 3) overloads alongside existing `ObjectMapper` constructors on the same class.

 This is very cleanly backwards compatible giving people a clean migration path, but requires both jackson2 and 3 to be on the consumers compile classpath. The whole point of this effort is to unblock people from having to have jackson2 on any part of their classpath if they'd like to embrace jackson3.                                                                                                                                                                                    
                                                            
  ### Approach 2: Change Primary Constructor to Abstraction                                                                                                                                                     
  Change the clients' primary constructor from `ObjectMapper` to a version-agnostic `ClientMapper` sealed interface.
                                                                                                                                                                                                                
This would move us away from exposing jackson directly in the api which has it's advantages long term. It's ultimately a breaking change for everyone constructing clients today with `ObjectMapper` which we'd like to avoid.
                                                                                                                                                                                                                
  ### Approach 3: New Classes w/ Abstraction in Primary Constructors (what this pr implements)                                                                                                                                                 
  Similar to Approach 2 but putting these into new classes avoids the breaking aspect of the change. We will backport the abstractions to give libraries on older versions the opportunity to migrate before fully removing the old versions of the classes.
                                                                                                                                                                                                                
We admit the naming is confusing in the interim period, but we deem it worth it and hope the deprecation warnings/kdoc will be informative enough.


graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsJsonMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+
+package com.netflix.graphql.dgs.internal
+
+import com.jayway.jsonpath.Configuration
+
+/**
+ * Abstraction over Jackson ObjectMapper/JsonMapper to allow DGS to work with
+ * either Jackson 3 (default) or Jackson 2 (opt-in via graphql-dgs-jackson2 module).
+ */
+interface DgsJsonMapper {
```
sjohnr: Would there be value in making this `internal` or is there an intent to allow users to use/implement as well? If it's just an internal implementation detail, would be nice to hide from the public API.

graphql-dgs-jackson2/src/main/kotlin/com/netflix/graphql/dgs/jackson2/DgsJackson2AutoConfiguration.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * When this module is on the classpath, Jackson 2 is used by default.
+ * Runs BEFORE DgsSpringGraphQLAutoConfiguration so that the Jackson 2 bean
+ * takes precedence over the Jackson 3 default.
+ *
+ * To override and use Jackson 3 even when this module is present, set:
+ *   dgs.graphql.preferred-json-mapper=jackson3
+ */
+@AutoConfiguration(
+    beforeName = ["com.netflix.graphql.dgs.springgraphql.autoconfig.DgsSpringGraphQLAutoConfiguration"],
```
sjohnr, line 38: Could a `compileOnly` dep on `graphql-dgs-spring-graphql` allow the non-string reference? I am thinking it's not at risk of cycles since the jackson2 module is an opt-in?

graphql-dgs-jackson2/src/main/kotlin/com/netflix/graphql/dgs/jackson2/Jackson2DgsJsonMapper.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.netflix.graphql.dgs.internal.DgsJsonMapper
+
+/**
+ * Jackson 2 implementation of [DgsJsonMapper].
+ * Used when consumers opt back into Jackson 2 via the `graphql-dgs-jackson2` module`.
+ */
+class Jackson2DgsJsonMapper : DgsJsonMapper {
+    val objectMapper: ObjectMapper =
+        ObjectMapper()
+            .registerModule(kotlinModule { enable(KotlinFeature.NullIsSameAsDefault) })
```
sjohnr, line 39: Intentional difference in config from Jackson3? (This is not present in `Jackson3DgsJsonMapper`)

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/Jackson3DgsJsonMapper.kt
```diff
@@ -0,0 +1,62 @@
...
+ * Note: Jackson 3 has built-in Java time support, so no separate JavaTimeModule is needed.
+ */
+class Jackson3DgsJsonMapper : DgsJsonMapper {
+    private val mapper: JsonMapper =
+        JsonMapper
+            .builder()
+            .addModule(KotlinModule.Builder().build())
+            .enable(EnumFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)
+            .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+            .disable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
```
sjohnr, line 41: Intentional difference in config from Jackson2? (This is not present in `Jackson2DgsJsonMapper`) Perhaps a comment to mention the defaults changed, or explicit configuration to make them match?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/Jackson3RestClientGraphQLClient.kt
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * A RestClient implementation of the DGS Client for blocking use, using Jackson 3 for serialization.
+ * A RestClient instance configured for the graphql endpoint (at least an url) must be provided.
+ */
+class Jackson3RestClientGraphQLClient(
+    private val restClient: RestClient,
+    private val headersConsumer: Consumer<HttpHeaders>,
+    private val mapper: JsonMapper,
+) {
```
sjohnr: This doesn't implement `GraphQLClient`, presumably to avoid inheriting the deprecated methods? (I'll talk about types below.)

Would it help to do a cleanup PR/feature/release to remove deprecated methods first?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/Jackson3RestClientGraphQLClient.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    constructor(restClient: RestClient, options: Jackson3RequestOptions? = null) : this(
+        restClient,
+        Consumer { },
+        Jackson3RequestOptions.createJsonMapper(options),
+    )
+
+    fun executeQuery(
+        @Language("graphql") query: String,
+    ): Jackson3GraphQLResponse = executeQuery(query, emptyMap(), null)
```
sjohnr: Why not return `GraphQLClientResponse` here? It seems like it would be possible to do a cleanup of the `GraphQLClient` interface to return this type. That would be a breaking change, but an easy one to adopt and easily justifiable.

If that change was made, perhaps you could unify this class (and other impls) under that interface? I think having a client like this not implement an interface of any kind is not ideal, and without the `GraphQLClient` interface to tie things together, there's a gap in the migration story here, which will be just as impactful as a breaking change (and in fact is one).

But there might be still other challenges you encountered that make this infeasible?

dgs-starter/build.gradle.kts
```diff
@@ -20,4 +20,5 @@ dependencies {
     api(project(":graphql-dgs-reactive"))
     api(project(":graphql-error-types"))
     api("org.springframework.boot:spring-boot-starter-graphql")
+    api("tools.jackson.module:jackson-module-kotlin")
```
sjohnr, line 23: Since this is defaulting to and providing Jackson3, this is already a breaking change (in that users would need to opt-in to Jackson2 if needed), correct? If that's the case, I think it might be worth considering a few other strategic breaking changes (see other comments).

graphql-dgs-example-jackson2-only/src/test/kotlin/com/netflix/graphql/dgs/example/jackson2/Jackson2OnlyTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        // The Jackson 3 client classes should NOT be usable since
+        // Jackson 3 is excluded from the classpath
+        val result =
+            runCatching {
+                Class
+                    .forName("com.netflix.graphql.dgs.client.Jackson3RestClientGraphQLClient")
+                    .getDeclaredConstructor(RestClient::class.java)
+                    .newInstance(RestClient.builder().baseUrl("http://localhost:$port/graphql").build())
+            }
+        assertThat(result.isFailure)
```
sjohnr: Should this also assert something specific about the thrown `Exception` so it's clear it's failing for the right reason?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
```diff
@@ -24,6 +24,7 @@ import org.intellij.lang.annotations.Language
  * The user is responsible for doing the actual HTTP request, making this pluggable with any HTTP client.
  * For a more convenient option, use [WebClientGraphQLClient] instead.
  */
+@Deprecated("Use Jackson3CustomGraphQLClient for Jackson 3 support", ReplaceWith("Jackson3CustomGraphQLClient"))
```
sjohnr: Should `GraphQLClients.handleResponse(HttpResponse, String, String, ObjectMapper)` be `@Deprecated`?

graphql-dgs-example-jackson2-only/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+            mapOf("group" to "tools.jackson.module"),
+            mapOf("group" to "tools.jackson.datatype"),
+            mapOf("module" to "spring-boot-starter-jackson"),
+        )
+    configurations.filter { it.isCanBeResolved }.forEach { config ->
+        jackson3Excludes.forEach { exclude ->
+            config.exclude(exclude)
+        }
+    }
+}
```
sjohnr, line 35: I'm guessing this is just to ultra-ban the deps for testing, and not the pattern we would be recommending for [excluding](https://docs.gradle.org/current/userguide/how_to_exclude_transitive_dependencies.html) Jackson3 from the classpath? (I don't know that `afterEvaluate` is generally recommended?)

sjohnr (changes requested):
@jjacobs44, this is great work to get to this point! I have some comments below, but the most important thing (worth discussing if it's not clear from my comments) is around `GraphQLClient`. See comments on `Jackson3RestClientGraphQLClient`.

To expand more generally, I think a breaking change of some kind is needed to actually smooth out migration, as well as keeping the framework healthy. I would be in favor of a minor (or major) release to separately introduce the new `GraphQLClientResponse` interface. You could then also introduce a small breaking change to return that type from non-deprecated `GraphQLClient` methods, and also remove the deprecated methods (or increase deprecation level to error). Maybe some other cleanup and deprecations could be done as needed to provide visibility for other coming changes.

Afterwards, this PR could be part of a follow-up major release to switch over the default to Jackson 3 and introduce this support with the ability to fall-back to Jackson 2. Any additional necessary breaking changes would then be on the table, but would be much smoother for users since the groundwork is laid in the previous release.

graphql-dgs-example-jackson2-only/build.gradle.kts
jjacobs44: Yeah this wouldn't necessarily be the recommended pattern, was more a quick and easy way to get a test going. I can iterate on cleaning this up into something we can recommend

graphql-dgs-jackson2/src/main/kotlin/com/netflix/graphql/dgs/jackson2/Jackson2DgsJsonMapper.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.jayway.jsonpath.spi.json.JacksonJsonProvider
+import com.jayway.jsonpath.spi.mapper.JacksonMappingProvider
+import com.netflix.graphql.dgs.internal.DgsJsonMapper
+
+/**
+ * Jackson 2 implementation of [DgsJsonMapper].
+ * Used when consumers opt back into Jackson 2 via the `graphql-dgs-jackson2` module`.
+ */
+class Jackson2DgsJsonMapper : DgsJsonMapper {
+    val objectMapper: ObjectMapper =
```
iparadiso: nit: should make this private otherwise it can leak Jackson2 into any code that injects and expects a DgsJsonMapper but can cast back and access object mapper impl.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/CustomGraphQLClient.kt
iparadiso: Something I just learned.  ReplaceWith has an argument for imports.  For all of these @Deprecated statements you added, please add the import as the second argument.  This will properly work in IntelliJ to  and AI to assist in migration.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/Jackson3RestClientGraphQLClient.kt
```diff
@@ -0,0 +1,103 @@
...
+        mapper,
+    )
+
+    constructor(restClient: RestClient, headersConsumer: Consumer<HttpHeaders>) : this(
+        restClient,
+        headersConsumer,
+        Jackson3RequestOptions.createJsonMapper(),
+    )
+
+    constructor(restClient: RestClient, options: Jackson3RequestOptions? = null) : this(
```
iparadiso: Any reason we need this explicit null default constructor.  I find these API's a little messy and unnecessary.  For new clients we can perhaps start a bit cleaner.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsGraphQLResponse.kt
```diff
@@ -0,0 +1,81 @@
...
+        typeRef: TypeRef<T>,
+    ): T
+
+    /** Extracts RequestDetails from the response if available. Returns null otherwise. */
+    fun getRequestDetails(): RequestDetails?
+
+    /** Returns true if the response contains errors. */
+    fun hasErrors(): Boolean
+
+    companion object {
```
iparadiso, line 89: Minor: Duplication here with companion getDataPath with the derived class GraphQLResponse.  Perhaps deprecate the GraphQLResponse version and call the interface version.  Kotlin vs. Java can have different behaviors if you call GraphQLResponse.getDataPath() in Java.  It won't delegate to the interface .

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/SpringGraphQLDgsQueryExecutor.kt
```diff
@@ -41,9 +42,13 @@ class SpringGraphQLDgsQueryExecutor(
     private val executionService: ExecutionGraphQlService,
     private val dgsContextBuilder: DefaultDgsGraphQLContextBuilder,
     private val dgsDataLoaderProvider: DgsDataLoaderProvider,
+    private val dgsJsonMapper: DgsJsonMapper,
     private val requestCustomizer: DgsQueryExecutorRequestCustomizer = DgsQueryExecutorRequestCustomizer.DEFAULT_REQUEST_CUSTOMIZER,
     private val graphQLContextContributors: List<GraphQLContextContributor>,
 ) : DgsQueryExecutor {
+    private val parseContext: ParseContext
+        get() = JsonPath.using(dgsJsonMapper.jsonPathConfiguration())
```
iparadiso: A get() with no backing field runs `JsonPath.using(dgsJsonMapper.jsonPathConfiguration())` every time, which can be a performance hit.  consider backing this in a top level field so it's set once.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/SpringGraphQLDgsReactiveQueryExecutor.kt
```diff
@@ -100,7 +101,7 @@ class SpringGraphQLDgsReactiveQueryExecutor(
         clazz: Class<T>,
     ): Mono<T> =
         getJsonResult(query, variables, null)
-            .map(BaseDgsQueryExecutor.parseContext::parse)
+            .map(JsonPath.using(dgsJsonMapper.jsonPathConfiguration())::parse)
```
iparadiso: Another opportunity to compute this once as the json mapper is injected at constructor time and is immutable, so you can compute this once and store it as a private field.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
```diff
@@ -177,6 +182,19 @@ open class DgsSpringGraphQLAutoConfiguration(
...
     }
 
+    @Configuration(proxyBeanMethods = false)
+    @ConditionalOnClass(name = ["tools.jackson.databind.json.JsonMapper"])
+    @ConditionalOnProperty(
+        name = ["dgs.graphql.preferred-json-mapper"],
+        havingValue = "jackson3",
+        matchIfMissing = true,
+    )
+    open class Jackson3DgsJsonMapperConfiguration {
```
iparadiso: Should we consider a fallback `@ConditionalOnMissinBean(DgsJsonMapper::class)` so we always have a mapper interface?  Could we realistically get into this state?

Maybe in this case, throw an RuntimeException with a useful message to add a Jackson preference and configuration.

sjohnr (approved):

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
```diff
@@ -177,6 +182,27 @@ open class DgsSpringGraphQLAutoConfiguration(
...
+        @Bean
+        @ConditionalOnMissingBean(DgsJsonMapper::class)
+        open fun dgsJsonMapper(): DgsJsonMapper = Jackson3DgsJsonMapper()
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(DgsJsonMapper::class)
+    open fun dgsJsonMapperFallback(): DgsJsonMapper =
```
paulbakker: When would this kick in? The bean above would create this bean?

jjacobs44: I think the only way you could get here is if you had `dgs.graphql.preferred-json-mapper=jackson2` and the jackson2 module somehow got dropped from your classpath (by accident on your own or a lib dropping it and you no longer getting it transitively). 

Or any of the other conditionals for the bean above failing. Happy to remove if that seems overly cautious.

paulbakker:
I'm really on the fence about what to do with the client interfaces. 
It's not obvious for users why `DgsGraphQLClient` is jackson version agnostic, while `GraphQLClient` is Jackson 2. 
Having `Jackson3CustomGraphQLClient` explicitly in the name is also not great for the future. 

However, if we "fix" this now, we're introducing a breaking change, and there is a lot of use of these interfaces and classes.

One alternative idea, could we introduce the new stuff in a new separate module, so we can make this depend on the classspath?

graphql-dgs-example-jackson2-only/src/test/kotlin/com/netflix/graphql/dgs/example/jackson2/Jackson2OnlyTest.kt
```diff
@@ -0,0 +1,136 @@
...
+
+        val response = client.executeQuery("{ hello(name: \"Custom\") }")
+
+        assertThat(response).isInstanceOf(DgsGraphQLResponse::class.java)
+        assertThat(response.hasErrors()).isFalse()
+        assertThat(response.extractValue<String>("hello")).isEqualTo("hello, Custom!")
+    }
+
+    @Test
+    fun `Jackson 3 client classes fail with NoClassDefFoundError`() {
```
paulbakker: This is a weird test

graphql-dgs-example-jackson3-only/src/test/kotlin/com/netflix/graphql/dgs/example/jackson3/Jackson3OnlyTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        assertThat(response).isInstanceOf(DgsGraphQLResponse::class.java)
+        assertThat(response.hasErrors()).isFalse()
+        assertThat(response.extractValue<String>("hello")).isEqualTo("hello, Custom!")
+    }
+
+    @Test
+    fun `Jackson 2 client classes fail with NoClassDefFoundError`() {
+        val result =
+            runCatching {
+                Class
```
paulbakker: Same weird test

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
```diff
@@ -51,21 +45,56 @@ import java.util.concurrent.CompletionException
...
+                e,
+            )
+        }
+    }
+
+    @Deprecated(
+        message = "Use DgsJsonMapper instead. This field requires Jackson 2 on the classpath.",
+        level = DeprecationLevel.HIDDEN,
+    )
+    val objectMapper: com.fasterxml.jackson.databind.ObjectMapper by lazy { jackson2ObjectMapper }
```
paulbakker, line 70: It looks like this is unused?

paulbakker (changes requested):
Approach looks good, few minor things I noticed in the code

graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/BaseDgsQueryExecutor.kt
jjacobs44: Removed all usages in the framework but it's a public api so removing it is technically breaking (some people are calling `BaseDgsQueryExecutor.objectMapper`). Very few internal usages though so I could remove it and proactively move users off of it.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsGraphQLClient.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package com.netflix.graphql.dgs.client
+
+import org.intellij.lang.annotations.Language
+
+/**
+ * Jackson-agnostic blocking GraphQL client contract. Implemented by the new `Dgs*` client
+ * classes and (for back-compat) by the deprecated [GraphQLClient].
+ */
+interface DgsGraphQLClient {
```
iparadiso, line 25: Feel free to push back on this, but since we have an opportunity to start new, I'm wondering if these interfaces are doing too much.  Rather than all of these overloads that every client has to implement we have one method.  

```
interface DgsGraphQLClient {                                                                                                                                         
      fun executeQuery(request: DgsGraphQLRequest): DgsGraphQLResponse
  }         
```

Where `DgsGraphQLRequest` represents the request and is not a function of the client itself and uses a builder pattern for all the required and optional params which is a cleaner API contract for future changes.  

Likewise, we can trim the `DgsGraphQLResponse` to stable caller contracts 
```
  interface DgsGraphQLResponse {                                                                                                                                       
      val errors: List<GraphQLError>
      fun hasErrors(): Boolean                                                                                                                                         
      fun <T> dataAsObject(clazz: Class<T>): T                                                                                                                         
      fun <T> extractValue(path: String): T   
      fun <T> extractValueAsObject(path: String, clazz: Class<T>): T                                                                                                   
      fun <T> extractValueAsObject(path: String, typeRef: TypeRef<T>): T
  }      
```

The other response fields move to `DefaultDgsGraphQLResponse`

Where this might be more challenging is in migrations.  :(

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsWebClientGraphQLClient.kt
```diff
@@ -0,0 +1,114 @@
...
+ * must pass [Jackson2DgsJsonMapperAdapter] explicitly.
+ */
+class DgsWebClientGraphQLClient(
+    private val webclient: WebClient,
+    private val headersConsumer: Consumer<HttpHeaders>,
+    private val mapper: DgsJsonMapper,
+) : DgsMonoGraphQLClient {
+    constructor(webclient: WebClient) : this(webclient, Consumer {})
+
+    constructor(webclient: WebClient, headersConsumer: Consumer<HttpHeaders>) :
```
iparadiso, line 32: General nit:  This (and other client variants) carries forward the previous multi-constructor overloads of the old clients.  Consider if a builder pattern would be cleaner here to prevent a future 6th or 7th constructor.  

Similar to my previous comment though can make migrations a bit busier.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/DgsSpringGraphQLAutoConfiguration.kt
```diff
@@ -177,6 +182,27 @@ open class DgsSpringGraphQLAutoConfiguration(
...
+        throw IllegalStateException(
```
iparadiso: This will result in a BeanCreationException with this exception buried below in a "caused by" at the bottom.  For errant bean creation paths, we usually define our own exception type. 

```
class DgsJsonMapperMissingException : RuntimeException()
```

Throw 
```
@Bean
@ConditionalOnMissingBean(DgsJsonMapper::class)
fun dgsJsonMapperFallback(): DgsJsonMapper = throw DgsJsonMapperMissingException()
```

And then handle it with a Failure Analyzer for our exception

```
class DgsMapperFailureAnalyzer : AbstractFailureAnalyzer<DgsJsonMapperMissingException>() {
    override fun analyze(rootFailure: Throwable, cause: DgsJsonMapperMissingException): FailureAnalysis {
        return FailureAnalysis(
            "No DgsJsonMapper bean found.",
            "Add 'tools.jackson.core:jackson-databind' (Jackson 3) or 'graphql-dgs-jackson2' to your classpath.",
            cause
        )
    }
}
```

iparadiso: And register it in spring.factories.  :)

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsRestClientGraphQLClient.kt
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }.headers(this.headersConsumer)
+                .body(serializedRequest)
+                .retrieve()
+                .toEntity<String>()
+
+        if (!responseEntity.statusCode.is2xxSuccessful) {
+            throw GraphQLClientException(
+                statusCode = responseEntity.statusCode.value(),
+                url = "",
```
iparadiso, line 69: The url would be useful here...Anyway to extract it from the client a a top level field?

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsGraphQLRequestOptions.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Jackson-agnostic request options for DGS GraphQL clients. Consumed by adapters
+ * (e.g. [Jackson3DgsJsonMapperAdapter.fromOptions]) to build a [com.netflix.graphql.dgs.json.DgsJsonMapper].
+ */
+class DgsGraphQLRequestOptions(
+    val scalars: Map<Class<*>, Coercing<*, *>> = emptyMap(),
+    val graphQLContext: GraphQLContext = GraphQLContext.getDefault(),
+) {
+    constructor() : this(emptyMap())
+
+    constructor(scalars: Map<Class<*>, Coercing<*, *>>) : this(scalars, GraphQLContext.getDefault())
```
iparadiso: nit:  TIL, you can just use `@JvmOverloads ` and it'll produce all the constructor for kotin and java.   

```
class DgsGraphQLRequestOptions @JvmOverloads constructor(                                                                                                            
      val scalars: Map<Class<*>, Coercing<*, *>> = emptyMap(),                                                                                                         
      val graphQLContext: GraphQLContext = GraphQLContext.getDefault(),                                                                                                
  )
```

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsGraphQLClient.kt
jjacobs44: I'm on the fence with this so wanna hear others opinions. `client.executeQuery("{ foo }");` becoming `client.executeQuery(DgsGraphQLRequest.builder().query("{ foo }").build());` is quite a ergonomics downgrade for users imo, but I agree with the general point you're making about the complexity of the overloads not being ideal. Same thing for your comment below.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsRestClientGraphQLClient.kt
jjacobs44: I'm looking around and it seems like once you've built your Rest/WebClient there's no getter where we can grab the baseUrl from. Observability should show what service you're calling though and 99% of the time the actual endpoint is just `/graphql` so I think this is acceptable.

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/DgsGraphQLClient.kt
iparadiso: The academic angle here is if builders become too much boilerplate and telescoping constructors reek of code smell, we have an API that's simply too flexible.  

The decision then becomes refactoring.  Use more static factory methods, focus only on required constructor params and moving optionals to configuration setters.

iparadiso: We don't need to answer all of these questions in this major version, but I think we'll eventually have to break away from this pattern as client functionality grows.

graphql-dgs-spring-graphql/src/main/kotlin/com/netflix/graphql/dgs/diagnostics/DgsJsonMapperMissingException.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2026 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ */
+
+package com.netflix.graphql.dgs.diagnostics
+
+class DgsJsonMapperMissingException : RuntimeException()
```
sjohnr: This exception has no message. It might not produce very nice details for DX? Since you have a failure analyzer, might be a moot point but still probably would be nice to include a message.

sjohnr (approved):
This looks great @jjacobs44! Only thing I am thinking about is surface area of the API. For changes like this in OSS, I like to look for opportunities to reduce visibility, and open it up later if/when users see a need for accessing the classes and provide that feedback.

You might consider making some/all of the following classes `internal`:

  1. `DgsJsonMapperMissingException` → internal — zero risk, only thrown inside the auto-config
  2. `Jackson3DgsJsonMapperAdapter` → internal — no external instantiation found; client wiring detail
  3. `Jackson2DgsJsonMapper` → internal — only constructed within graphql-dgs-jackson2
  4. `Jackson3DgsJsonMapperConfiguration` inner class → internal open class — Spring auto-config inner class
  5. `DefaultDgsGraphQLResponse` → internal (judgment call) — its KDoc invites users to construct it for tests; either commit to that or hide it

There are a few others where it would be nice to hide (e.g. `Jackson2DgsJsonMapperAdapter`, `DgsJackson2AutoConfiguration`, `Jackson3DgsJsonMapper`) but I believe these are needed/referenced in other places and can't be reduced.

PR #2300, opened 2026-04-17
Backport new client changes to 10.x

jjacobs44:


graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClientResponse.kt
```diff
@@ -0,0 +1,108 @@
...
+
+package com.netflix.graphql.dgs.client
+
+import com.jayway.jsonpath.DocumentContext
+import com.jayway.jsonpath.TypeRef
+import org.slf4j.LoggerFactory
+
+/**
+ * Common interface for GraphQL response types, providing Jackson-version-agnostic access
+ * to response data. Both [GraphQLResponse] (Jackson 2) and [Jackson3GraphQLResponse] (Jackson 3)
```
sjohnr: Minor: This class doesn't exist yet. Does it chuck a warning in javadoc generation? (Might be fine to ignore, but could be slightly confusing.)

sjohnr (approved):

graphql-dgs-client/src/main/kotlin/com/netflix/graphql/dgs/client/GraphQLClientResponse.kt
jjacobs44: Oh cherrypick/copy pasta fail. Will clean that up thanks!

PR #2330, opened 2026-08-05
Fix: more vigorous handling of syntax for `spring.autoconfigure.exclude`

### LINKED ISSUE #2329: bug:
Please read our [contributor guide](https://github.com/Netflix/dgs-framework/blob/master/CONTRIBUTING.md) before
creating an issue.

## Expected behavior
* When a Spring Boot application which uses DGS framework specifies some autoconfiguration classes to exclude via `spring.autoconfigure.exclude` property in a `application-<profile>.yaml` file using array syntax, then this array gets augmented with:
  * `org.springframework.boot.graphql.autoconfigure.observation.GraphQlObservationAutoConfiguration`
  * `org.springframework.boot.graphql.autoconfigure.security.GraphQlWebMvcSecurityAutoConfiguration`

  For example, the `application-<profile>.yaml` could be:

  ```yaml
  spring.autoconfigure.exclude:
    - com.example.AutoconfigA
    - com.example.AutoconfigB
  ```
* Alternatively, when index-based syntax is used for a said property, the expectation is the same - the two above classes should be _added_ to the array of exclusions, but not replacing it completely.
  ```properties
  spring.autoconfigure.exclude[0]=com.example.AutoconfigA
  spring.autoconfigure.exclude[1]=com.example.AutoconfigB
  ```

## Actual behavior
The value is not augmented but rather gets _overridden_ with the two graphql configuration classes given above.

anenviousguest:
This fixes #2329

Pull request checklist
----

  first

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Changes in this PR
----

In a nutshell - this PR makes `ExcludeAutoConfigurationsEnvironmentPostProcessor` defer the retrieval of `spring.autoconfigure.exclude` to Spring's binder instead of manually looping through available properties source and doing the guess work wrt syntax which was used for a property.

As far as the Test Properties are concerned - there was a dedicated code in the `ExcludeAutoConfigurationsEnvironmentPostProcessor` which was trying to ensure that those takes precedence - but yet again, this is rather the job of SpringBoot itself and not of DGS to make this happen. I checked #2032 which was describing the original issue with test properties not taking precedence and added a dedicated `SpringBootTest` to illustrate that it should work without special handling on the side of DGS.


graphql-dgs-spring-graphql/src/test/kotlin/com/netflix/graphql/dgs/springgraphql/autoconfig/ExcludeAutoConfigurationsEnvironmentPostProcessorOverrideTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.test.context.TestPropertySource
+
+
+@SpringBootTest(
+    classes = [
+        ExcludeAutoConfigurationsEnvironmentPostProcessorOverrideTest.TestApp::class,
+    ],
+    webEnvironment = SpringBootTest.WebEnvironment.NONE
+)
+@TestPropertySource(properties = ["spring.autoconfigure.exclude=someotherexclude"])
```
anenviousguest, line 17: basically, that's a replacement of the test which i removed from `ExcludeAutoConfigurationsEnvironmentPostProcessorTest` (see below).

Here, we have test property source, with our expectation that this will take priority over the same property defined via `application-<profile>.properties` which I also added

paulbakker (approved):

anenviousguest:
thanks @paulbakker - I pushed one more commit to fix linting errors.

jjacobs44:
Hey @anenviousguest, first off thanks for the report and contribution!

It seems that while this does fix your initial bug report, it also changes the underlying behavior of the exclude mechanism more broadly. As we have it implemented now, we join the value of `spring.autoconfigure.exclude` across all property sources. With the implementation here, it now only resolves to the the value of `spring.autoconfigure.exclude` for the single highest priority property source. You might argue that's more "correct", but in any case that's a breaking change.

Any opposition to just using a comma separated list instead of yaml array syntax? IIRC there are many pitfalls around using yaml array syntax for a variety of spring properties and comma separated lists avoid the headache.

anenviousguest:
Thanks @jjacobs44 , this is a fair point.

Ultimately, it would be your call as code owners, but please consider this:

* `spring.autoconfigure.exclude` is a property maintained by SpringBoot, not by DGS
* ergo, it is liable to the rules outlined in [Externalized Configuration](https://docs.spring.io/spring-boot/reference/features/external-config.html) documentation:
  > Spring Boot uses a very particular PropertySource order that is designed to allow sensible overriding of values. Later property sources can override the values defined in earlier ones.
  > 
  > Sources are considered in the following order:
  > ...
  
  There follows a list of 14 bullet points whose order is important. 

  And further down on the same page - under Merging Complex Types section, where it says:
  > When lists are configured in more than one place, overriding works by replacing the entire list.

  This is what all SpringBoot developers are relying upon. 

* The documentation of DGS (https://netflix.github.io/dgs/) does _not_ seem to mention anywhere that DGS fiddles around with `spring.autoconfigure.exclude`, let alone any particular algorithm of _how exactly_ it does that. Neither did I manage to find an occurrence of `GraphQlObservationAutoConfiguration` or `GraphQlWebMvcSecurityAutoConfiguration` in the DGS documentation. Taking that into account, it's probably unfair to treat internals of the class we're modifying in this PR as the contract DGS has to stick to, especially given that the property itself is coined by SpringBoot which outlines the contract for merging properties in another way. 

Also I looked once again at #2032 - even there, the author complained that:
> Setting the spring.autoconfigure.exclude e.g. in unit test by the @TestProperties appends the value to the configuration property already defined e.g. in application.yaml file. Does not override it.

So, the expectation _was_ based on SpringBoot contract - that `@TestProperties` would override what was defined in `application-<...>.yaml` (or .properties, for that matter), and not that they would be concatenated from multiple sources. 

WDYT?

PS: to this one:
> Any opposition to just using a comma separated list instead of yaml array syntax? 

That's the work-around we are currently forced to use, but it's far from perfect because from ergonomics standpoint, arrays are easier to maintain. This particular property holds an array of fully qualified class names, so keeping it in a single string drastically reduces readability and makes it harder to grok the diff.

Thanks.

anenviousguest:
hi @paulbakker @jjacobs44 , kindly bumping this one - would appreciate your review - thanks!

jjacobs44:
Hey @anenviousguest! After discussing a bit internally I think we're aligned with supporting the yaml list syntax and making the breaking change to resolve to the single highest precedence property source. That being said we'd like to do some experiments with other implementations that will make the transition easier for us internally. 

We appreciate your patience, I'll be looking into this as soon as I can.

jjacobs44:
> I think we're aligned with supporting the yaml list syntax and making the breaking change to resolve to the single highest precedence property source

After even more discussion it was decided that making the breaking change to have it align with normal Boot rules isn't worth the risk despite being more "correct". https://github.com/Netflix/dgs-framework/pull/2339 was accepted as a way to fix the original bug (supporting yaml list syntax) while preserving the old behavior of merging the property sources.

